PR #1, opened 2017-01-23
Allow manual entry of channel information in grid files using command line flag

egthomas:
This is necessary because data from UAF radars are written to a separate rawacf file for each channel.  Once these data are processed into fitacf and then grid files, the original channel information (which was only identifiable from the rawacf filename) is no longer retained in the grid file.  Users are now able to manually set the channel information stored in grid files for these radars by passing values of a,b,c, or d with the 'cn_fix' flag which will store an integer value of 1, 2, 3, or 4 respectively in the output grid file.


egthomas:
There was a problem with my modification which prevented the 'cn' flag from working properly;  See #2 for a revised version.

Issue #6, opened 2017-02-07
vt idl library


egthomas:
So there currently is a folder of IDL routines in codebase/superdarn/src.idl/vt/ where it looks like AJ uploaded some of his old software.  As several of these routines are dependent on common blocks and functions native to Davit (Lasse Clausen's IDL package, not the newer python software), it doesn't seem appropriate to distribute all of this stuff with the RST.  I guess my question is whether anyone has an objection to removing this directory (and the related output_plots directory)?  These IDL routines would still be archived in VTRST3.5 on the vtsuperdarn github account if someone really wanted to access them.

sshepherd:
I would remove them.

egthomas:
Similarly, do we need to keep the compare_files.py script?  This looks like the only piece of python software in the RST and was again added by AJ.

ksterne:
Yea, I'm not what these files are for.  It seems as though they are IDL versions of some of the fitacf processing routines?  Like you say, they are being backed up in the older VTRST3.5 repo, so it should be safe to remove them.  I definitely agree with removing the `compare_files.py` code.  That is a neat short script on comparing the file lengths of two fitacf files.  But this could easily be reproduced if needed.

pasha-ponomarenko:
Kevin, I think compare_files.py has been developed by AJ when we satrted refactoring FITACF in August 2014. We wanted to be sure that consecutive modifications ot the package produce the same result.

egthomas:
Closing this issue as it was addressed by pull request #9 as well.

Issue #4, opened 2017-02-03
Removing object and other compiler generated files?


ksterne:
I think this was a to-do for me a while ago, but should we remove all of the object files, binaries and other files that are created during the compilation process?  I see the ~/bin/ directory is still removed here, but object files are still up.

egthomas:
I did this in a commit on the feature/grid2 branch (https://github.com/SuperDARN/rst/commit/9ed327b335e74c302bef3e0472b3cbc8f077b9ec) but it's kind of buried with a few other gridding-specific commits.  It's probably not something that should have been rolled out together.

egthomas:
Just cleaned up the *.a, *.so, and *.o files (139f267ea0e654364154d348b3244f2a821a7916).  Is there anything else that should be removed?

egthomas:
Ok, after today's round of commits the develop branch should no longer be tracking any files that are created during the compilation process.  After a fresh clone/download of the rst and compilation the only files that show up from a "git status" are the base.bash and idl.bash files (which must be modified to set some user-dependent environment variables).

ksterne:
Looks like we're a lot better here. So, closing this out.

PR #9, opened 2017-02-08
File cleanup

egthomas:
This pull request removes the VT-specific IDL library and compare_files.py script described in issue #6 .  Also strips out and ignores nearly all of the .h files in the include/ directory (except mpfit.h) which are populated there during compilation.  Finally, the install notes have been simplified as it is not necessary to copy the cdf.h file if the CDF_PATH environment variable is properly set.


ksterne:
I've added back the fitexfilter binary that was wiped out.  I had forgotten past discussions that we need to keep this around as people (at least here at VT) are using this binary, but we do not have the source code to recompile it.  So, while it may become a useless binary at some point, it's still useful today.  Hopefully one day someone will re-write the code or write a better filter and keep the source code around.

egthomas:
I have to disagree here - I don't think that a binary which was compiled years ago on a particular machine and doesn't have any supporting source code or documentation should be included in the main RST distribution.  Can you guys put it in a different directory on your local setup so that you can still run it as needed?

egthomas:
Although I still disagree I've left the fitexfilter in the commit history for now to avoid confusion.  This latest round of commits removes and stops tracking the dlm/ and doc/ directories which are populated during the compilation procedure by executing `make.code superdarn rst` and `make.doc superdarn rst` respectively.  

This is also where I'm going to draw the line on adding commits to this pull request.  If someone wants to try compiling this branch to verify there are no errors and everything is still populated it would be greatly appreciated.

ksterne:
@egthomas, I can certainly agree with the dangers of having a binary which we can't see the source code for in addition to why all the other pre-compiled files were removed.  It seems as though I took the fitexfilter binary out and placed it in our local davit binary directory and at least in shell its pointing to the davit directory.

The pro I see to having the fitexfilter up here is that then we are not the sole source of the binary.  That also means no one aside from VT has access to it.

I don't have strong feelings either way.  I worry about losing this if we're the sole source but I also understand the frustration and complications of having a binary we don't have source code for. Maybe we should get a ruling from the DAWG chair?  @pasha-ponomarenko?

sshepherd:
The issue here is that this is a binary that probably only works on
specific VT hardware and will likely stop working when any of the dependent
libraries change. Without source code the lifetime of this piece of code is
limited. It doesn't make any sense to keep binaries in the RST distribution.

egthomas:
From a file backup perspective the fitexfilter binary is still contained in at least one branch of the VTRST3.5 repository.

pasha-ponomarenko:
Kevin S, could you please clarify the exact function of this binary?

ksterne:
From the --help output:

This program takes a fitex (fitacf) format file and performs a boxcar filter
calling format:
fitexfilter [-options] fitex_file > filtered_fitex_file
options:
-help : display this menu
thresh f.f : filter threshold, a float in [0,1], default is 0.4
tdim x : time dimension of boxcar filter, 1 or 3, default is 3
bdim x : beam dimension of boxcar filter, 1 or 3, default is 3
rdim x : range dimension of boxcar filter, 1 or 3, default is 3
NB -- a 1x1x1 filter is equivalent to no filter at all
nocomb : do not combine multiple camping beams for a single scan
NB -- nocomb uses a 3x1x3 (TxBxR) filter on camping beams
vb : verbose

So, in short this is a boxcar filter for fit-level data.  

Again, it seems as though I've been able to pull this binary out of the rst and its working for our immediate uses here at VT.  I do take that, for now, the VTRST3.5 repo can serve as a backup...though I think it's now only in one branch.

Thoughts @pasha-ponomarenko?

egthomas:
Hang on, I guess I had assumed fitexfilter was AJ's ionospheric/ground scatter search algorithm.  Instead it sounds like it does the same thing (boxcar median filter) as the fitfilter function in codebase/superdarn/src.bin/tk/tool.

ksterne:
Yea, no, this isn't the blob analysis software that AJ also developed.  The result of the fitexfilter binary can be seen with:

http://vt.superdarn.org/tiki-index.php?page=Latest+Data

by toggling the filter on/off by clicking on the "Turn Filter On/Off" link.  Right now I think the filtered plot is set to not mark groundscatter.

pasha-ponomarenko:
Well, I don't see why would we need to distritbute it. It is very easy to replicate it in IDL or Python if necessary anyway... So, I think we can safely remove it from here.

egthomas:
OK, fitexfilter has been removed.  Getting back to the original intention of this pull request, it should now address issues #4 (removing compiler generated files) and #6 (removing VT-specific IDL software).  As a result a zipped download copy of the RST will be reduced in size by nearly 30% (88MB vs 124MB).

There may be more files to trim from the ROS side but I'm not as familiar with that portion of the RST and am hesitant to attempt a similar purge.  I would suggest merging this sooner rather than later so we all have cleaner file tracking in place as we continue to modify the software.

ksterne:
Sounds good here.  I don't know that much about the fitexfilter, but like we've said, I've got some backups if we need.

I have confirmed (not from scratch) that this branch does compile without any problems.  Should we wait on double confirmation or is the developer pool too small for that?  If it isn't, I can merge (pull requesters shouldn't merge their own branches).

egthomas:
The developer pool may be too small like you say but at this point I don't see the harm in waiting until Monday to give a little extra time.

PR #15, opened 2017-02-17
Fixing issues with normalising CRI

kkotyk:
In the original version, the echoes for all ranges were generated without scaling, i.e. effectively with the same lag 0 power. The dependence of the amplitude on range, amp0[r], has only been applied to the already simulated ACFs, which was also normalised on the average power across all ranges and divided by the number of avaraged pulses sequences. This post-scaling is incorrect with respect to the level of cross-range interference -- it was always the same regardless of amp0. I noticed the problem in 2013 but was able to locat its source only recently. The fix was rather simle: I (a) abandoned normalisation and (b) shifted scaling back into the simulation loop so that echoes from interfering ranges are accumulated with correct weights. Testing showed the expected changes in CRI levels at different ACF lags.

### DIFF
--- a/codebase/superdarn/src.lib/tk/fitacf.2.5/src/badlags_s.c
+++ b/codebase/superdarn/src.lib/tk/fitacf.2.5/src/badlags_s.c
@@ -74,11 +74,11 @@ void FitACFBadlagsStereo(struct FitPrm *ptr, struct FitACFBadSample *bptr) {
     	   }
     	   /* If txpl is a valid value, lets set it as smsep and throw off a warning */
     	   if (first == 0) {
-	   	fprintf( stderr, "FitACFBadlagsStereo: WARNING using txpl instead of smsep...\n")
+	   	fprintf( stderr, "FitACFBadlagsStereo: WARNING using txpl instead of smsep...\n");
 	        first=1;
            }
            ptr->smsep = ptr->txpl;
-   
+    }   
 
     while (i < (ptr->mppul - 1)) {
 	/* first, skip over any pulses that occur before the first sample */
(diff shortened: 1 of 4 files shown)


kkotyk:
I believe this is what you are looking for @pasha-ponomarenko ?

egthomas:
This should probably go into the develop branch rather than master, correct?  In a couple of months we're going to roll all of these changes into master at once.

kkotyk:
His phrasing was "the main one", so if you guys consider develop the main one, then I will do that.

pasha-ponomarenko:
Yes, and I just wanted it not to be lost in the transition. Develop branch is ok.

Sent from my Sony Xperia™ smartphone

---- kkotyk wrote ----


I believe this is what you are looking for @pasha-ponomarenko<https://github.com/pasha-ponomarenko> ?

kkotyk:
Okay, so apparently, you can't just cherry pick this one commit for the develop branch and these files need conflicts resolved. Everyone good with this?

sshepherd:
Has it been sufficiently tested like the other recent commits?

kkotyk:
I will wait for @pasha-ponomarenko to comment, but I found some more git advice that I might be able to use to cherry-pick only this commit.

pasha-ponomarenko:
Keith, if "cherry-picking" is such a problem (probably, it's a wrong season for the snow is still on the ground), I think a much safer way to do it is to create a new pull request and to copypaste the whole modified sim_data.c code in there.

pasha-ponomarenko:
Simon, if you mean testing the simulator, I have extensively done it a couple of years ago and made sure that it produces an expected level of CRI for both power and phase (see the very forst mesage in this pull request).

pasha-ponomarenko:
How about this? I just created a separate branch with sim_data.c changes ONLY:
https://github.com/SuperDARN/rst/commit/9125e4f2b06d05be1d1d80ec51101b993e81d795
I have checked carefully that the changes are the same as in the respective file from the refactor branch.
I hope it will work, just want someone else to have a look and then push.

egthomas:
I assume this pull request can be closed as it has been replaced by #19 ?

kkotyk:
@egthomas Yes I think thats completely fine.

PR #13, opened 2017-02-17
Removing ROS files from RST

egthomas:
As discussed in issue #12 , there is a lot of MSI-specific ROS software in the RST which is not necessary for general distribution to the SuperDARN community.  This commit removes all files from the os/ directories located within the codebase/ and usr/ paths.  These paths are also removed from the build.txt file which is used by make.code to determine which binaries and libraries to compile.  All of the files removed by this pull request will still be accessible under release v3.5 and from the VTRST3.5 repository.

PR #16, opened 2017-02-17
Integration of cmpfit library into standard compilation procedure

egthomas:
This pull request removes the band-aid method for compiling the external cmpfit library introduced by AJ for his lmfit software.  The cmpfit makefile has been re-written and the library now follows a directory structure consistent with the rest of the RST.  As a result, the include/ directory no longer needs to be tracked by git as the mpfit.h file is now populate during compilation, resulting in a cleaner home RST directory.


egthomas:
Just realized that there is a (non-fatal but confusing) naming conflict with an existing binary in the codebase/superdarn/src.bin/tk/testting directory called cmpfit.1.6 which is used to compare fit files, so the final commit here renamed this external library from 'cmpfit' to 'mpfit'.

ksterne:
Is there a particular test here?  Or see that the rst software package compiles?

egthomas:
Compilation is the biggest thing to test (I'd recommend using a fresh clone to avoid any untracked library/header files possibly interfering).  Otherwise, the binaries which call the mpfit library are fitcomp, make_fitex2, make_lmfit, test_fitex2, and test_lmfit.

egthomas:
make_lmfit and make_fitex2 both appear to work fine, but I can't even get test_fitex2 or test_lmfit to work on the develop branch (both fail after printing 2 lines and test_lmfit gives a segmentation fault).  fitcomp is another piece of AJ code that I'm not exactly sure what it's trying to do (none of these binaries have documentation).

ksterne:
Testing this out, it looks like on my spot check, the fitacf data product is the same as to what was produced a while ago.  However, the fitex (from make_fitex2) is slightly different.  I'm finding differences (although very small) in the w_l, and p_l at least at first look.  On the w_l, array it was only 1 of 15 values and even then it was only off by 0.000057 (4.038948 versus 4.039005).  On the p_l array, it was only 1 of 17 results and even then it was different by 0.000005 (13.604535 versus 13.604530).  It's odd as it doesn't happen on every beam record and it doesn't often happen that both p_l and w_l are different for the same record.

I guess a couple of questions, should we expect this small of a difference in these results?  Is the fact I'm running this in a virtual machine and desktop version whereas the "good" fit-level data I'm comparing to was run on a server (non-GUI) machine?  Or does my lack of IDL mess with the results here?  I did the same tricks as noted in #13 to get around not having IDL.

egthomas:
I don't think the workaround to having IDL should impact this.  If you supply / indicate the rawacf file you're using for testing, I can run make_fitex2 on it from this branch for comparison.  That might help indicate whether it's related to this pull request or something to do with the virtual machine.

ksterne:
I used the 20161012.16*bks.rawacf* file for testing.  But it might be good to try another one...maybe it's radar specific?

pasha-ponomarenko:
cmpfit is used in mpfit only so it should not affect other fitting routines like fitacf or fitex.

egthomas:
Apparently when AJ wrote make_fitex2 he included some functionality from the lmfit library (which depends on mpfit).  From what I can tell, these include the calc_err, setup_fblk, lm_noise_stat, and calc_phi0 functions.  So make_fitex2 won't compile if you remove the lmfit and mpfit libraries from the makefile.

pasha-ponomarenko:
Yep, now I dimly recall that he might have used lmfit for fitting the saw-tooth model to the wrapped measured ACF phase.

ksterne:
I guess my basic question was, I can find that there are differences, but do we care about such a small difference?  It is odd that a different file is being generated, but the values seem to be pretty close to the same.  This is only for a spot check on one file though.  One thought is that maybe if I can the fitex processing again on the same file, I might get another file that is different from either two or the others.  Then this might be some kind of randomness within the cmpfit function?

@egthomas, were you able to compare the results that I found on the bks file?

pasha-ponomarenko:
I guess one possibility could be a different MPFIT version:
https://www.physics.wisc.edu/~craigm/idl/cmpfit.html

egthomas:
I ran make_fitex2 on the 20161012.1601.00.bks.rawacf file from a fresh clone of the cmpfit branch on my Ubuntu desktop, then deleted that clone, and ran make_fitex2 again from a fresh cone of the develop branch on the same Ubuntu machine.  A diff of the dmapdump output of the two files shows the only differences being the string "origin.time" of each record (which makes sense, since I created the fitex files at different times).

My guess is that the differences found by @ksterne are not related to this pull request but an OS/memory-type issue with make_fitex2.

egthomas:
Furthermore, this branch also successfully compiles using Debian 8.7.1, openSUSE 42.2, and Mint 18.1 virtual machines.  I suggest this pull request be merged.

sshepherd:
Last call before I merge this request.

pasha-ponomarenko:
Simon, calm down. While the difference is minisque and does not seem to occur on your system, the problem has not been fully identified yet. I don't like to leave these apparently unimortant things behind and I believe it's Kevin's call right now. 
By the way, I cannot reacall you testing FITACF3? It's been up there for about a year...

sshepherd:
The errors are not related to this commit and has been tested on multiple
systems. Perhaps you should test it on your system.

Evan has tested fitacf3. Has there been a pull request? I don't see it on
the rst branch.

On Mon, Mar 6, 2017 at 10:18 PM, pasha-ponomarenko <notifications@github.com

pasha-ponomarenko:
I am talking about you personally, Simon. I would like to have your qualified opinion on the difference between FITACF2 and FITACF3 outputs. 
I am merging this request but leave a tickmark about the unresoved issue, which is not directly related to this thread.

ksterne:
Sorry I was not able to retest this again THE next business day.  Since the schedule had just come out I was busy updating our website and scheduling 6 radars.

I was going to note that I'm not sure @egthomas 's test of pre-merge develop branch versus this branch comparison isn't exactly what I had tested since the "old" file I was comparing to used a codebase that's more akin to the VTRST3.5 tag.  There's been several changes since then and I'm not sure which commit would have made this small differences in the fitex files.  My follow up test was going to be to reprocess fit-level data a second time and compare these 2nd round files to the first round files as well as to the VTRST3.5 files.  I can still do that if anyone's interested.

PR #19, opened 2017-02-22
Fixing issues with normalising CRI (second try)

kkotyk:
In the original version, the echoes for all ranges were generated without scaling, i.e. effectively with the same lag 0 power. The dependence of the amplitude on range, amp0[r], has only been applied to the already simulated ACFs, which was also normalised on the average power across all ranges and divided by the number of avaraged pulses sequences. This post-scaling is incorrect with respect to the level of cross-range interference -- it was always the same regardless of amp0. I noticed the problem in 2013 but was able to locat its source only recently. The fix was rather simle: I (a) abandoned normalisation and (b) shifted scaling back into the simulation loop so that echoes from interfering ranges are accumulated with correct weights. Testing showed the expected changes in CRI levels at different ACF lags.

### DIFF
--- a/codebase/superdarn/src.lib/tk/sim_data.1.0/src/sim_data.c
+++ b/codebase/superdarn/src.lib/tk/sim_data.1.0/src/sim_data.c
@@ -393,7 +393,8 @@ void sim_data(double *t_d, double *t_g, double *t_c, double *v_dop, int * qflg,
             /*calculate phase of scattered signal depending on its location*/
             phase = -4.*PI*(smptime*(v_dop[r]+range_gates[r][i].velo)+range_gates[r][i].space)/lambda;
             /*record scattered signal as a raw sample*/
-            raw_samples[smpnum] += amplitude*(cos(phase) + I*sin(phase));
+            /* raw_samples[smpnum] += amplitude*(cos(phase) + I*sin(phase)); */
+            raw_samples[smpnum] += amp0[r]*amplitude*(cos(phase) + I*sin(phase));
           }
         }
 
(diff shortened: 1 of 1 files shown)


kkotyk:
@pasha-ponomarenko Okay I think I got it. Look good to you?

ksterne:
Assuming this looks good (it seems as though you've got it right @kkotyk), @pasha-ponomarenko, do you have a test that we can run to see the problem and how this change in the code fixes the problem?  Maybe this was apart of an older Workshop presentation, but I'm not recalling this issue.

pasha-ponomarenko:
Kevin S, I will try to explain the way I tested it so that you decide if you wish to repeat it yourself. 
           As one would expect, power and phase flucutations at the CRI-affected lags should be proportional to the lag0 power(s) at the interfering range gate(s). These levels also depend on the number of averages and can be calculated theoretically.  Somewhere on my desktop in Saskatoon I have a piece of IDL code which runs the simulator as an external routine with given input parameters (power, width, velocity, number of averages) at given range gates. In 2013, in order to test if the simulato generates correct CRI levels, I simulated a 1000 ACFs at each of only two interfering range gates and analysed lag dependence of the fluctuation magnitudes for both power and phase.  Initial comparison showed that these values were, quite unexpectedly, independent of the interfering power and always stayed at a fixed level corresponding to equal lag 0 powers at both range gates. I was surprised and discussed at length this problem with AJ, but at that time we have not been able to find its cause. In 2015 I came back to this issue and after a detailed anaysis of the simulator code reliased that the actual simulation was run for the equal powers at both range gates, and only the resulting (averaged) ACF were scaled according to the input power values. When I changed the code in such a way that the scaling occured before the actual averaging, the simulated fluctuations started to show the expected levels for the given number of average and lag 0 power ratio. 
              In my view, because this is a non-critical part of RST which does not affect the actual data processing, I would just make sure that the changes made to it do not affect compilability of the RST.

PR #21, opened 2017-02-22
Bug fix to makeall. Major version increment was not tracking properly.

sshepherd:
This is a one-line change with an addition comment line. It fixes a problem with incrementing major version numbers in the makeall function.

### DIFF
--- a/codebase/base/src.bin/build/makeall.1.21/makeall.c
+++ b/codebase/base/src.bin/build/makeall.1.21/makeall.c
@@ -153,7 +153,8 @@ void add_make(char *name,char type) {
 
       /* now compare version numbers */
       if (major<makevmajor[k]) return;
-      if (minor<makevminor[k]) return;
+      if (minor<makevminor[k] && major == makevmajor[k]) return;
+			/* bugfix SGS: 17Feb2017 was returning when major> and minor< */
 
       makevmajor[k]=major;
       makevminor[k]=minor;


ksterne:
Seems simple enough, but example on how to test this?  Is it the version of the makeall binary or another binary?  I don't see where there's any version numbers to edit here since they were cleaned up in bafca29.

sshepherd:
The build procedure is somewhat complicated but if you look at it you will see that there is an option to build only the latest version and this is in makeall.c. It has been buggy and you will see several bug fixes that I made prior to RST 3.5.

The test is the create another directory say codebase/analysis/src.lib/aacgm/aacgm.2.0/, copy everything from codebase/analysis/src.lib/aacgm/aacgm.1.15/ into it (including the .rst hidden directory). Change the version file in .rst from 1.15 to 2.0 then build and look what library is created.

We are removing the need for the .rst hidden directory and simplifying the build process in a later commit, but this issue should be fixed first.

ksterne:
@sshepherd, I'm guessing that the expected result is that the compilation code should always pick the higher version to use during the compilation process?

To test this, I created `~/codebase/analysis/src.lib/aacgm/aacgm.2.0/` with the `.rst`, `include`, and `src` directories populated with code from `aacgm.1.15/`.  Is it OK that I skipped the `doc` directory?  Then when I run `make.code superdarn rst` on the `develop` branch, I get the following in the `~/lib/` directory:

```
-rw-rw-r-- 1 davit davit 47796 Feb 27 21:19 lib/libaacgm.1.15.a
-rwxrwxr-x 1 davit davit 41315 Feb 27 21:19 lib/libaacgm.1.15.so
lrwxrwxrwx 1 davit davit    15 Feb 27 21:19 lib/libaacgm.1.a -> libaacgm.1.15.a
lrwxrwxrwx 1 davit davit    16 Feb 27 21:19 lib/libaacgm.1.so -> libaacgm.1.15.so
-rw-rw-r-- 1 davit davit 47796 Feb 27 21:12 lib/libaacgm.2.0.a
-rwxrwxr-x 1 davit davit 41315 Feb 27 21:12 lib/libaacgm.2.0.so
```

Whereas when I run this with the `makeall_fix` branch, I get:

```
-rw-rw-r-- 1 davit davit 47796 Feb 27 21:27 lib/libaacgm.1.15.a
-rwxrwxr-x 1 davit davit 41315 Feb 27 21:27 lib/libaacgm.1.15.so
lrwxrwxrwx 1 davit davit    14 Feb 27 21:33 lib/libaacgm.1.a -> libaacgm.2.0.a
lrwxrwxrwx 1 davit davit    15 Feb 27 21:33 lib/libaacgm.1.so -> libaacgm.2.0.so
-rw-rw-r-- 1 davit davit 47796 Feb 27 21:33 lib/libaacgm.2.0.a
-rwxrwxr-x 1 davit davit 41315 Feb 27 21:33 lib/libaacgm.2.0.so
```

This is a bit tricky to test with flipping back and forth between branches as it seems like I have to run the compilation process twice before I get these results.  Otherwise, it seems like it sticks with the old code and doesn't update the results.  Is there a better way to test this?

I agree here that there is a larger discussion about the RST version control being clunky or obsolete when working within git.  A long-term goal may be to strip this out.  Though I'm guessing major versions may need to stay around during a transition period?  I thinking about aacgmv1 versus aacgm2 or maybe make_fitex versus make_fitex2 (I think that's a thing).

sshepherd:
Yes. This is the purpose of makeall. However, you can explicitly tell it to compile all versions. Notice the timestamps on your output. In the first case it only built v.1.15 in the second it only built v.2.0.

This is a one-line bug fix.

The version numbering is fine. The only problem is the rpkg dependency and we are working on a fix but it depends on this bug-fix.

ksterne:
<grumble> Merging your own pull request should be discouraged here.  Again, with patience I would have merged this in the next couple of days. </grumble>

That sounds good on the explanation of it only building one or the other versions.  This looks good.

My note about the version numbers on these functions is that they are made obsolete within the git development model.  Git itself is a version control system, so having version numbers is not meaningful unless as I mentioned we're trying to run two version of functions in parallel.

ksterne:
Also, I think I figured out the answer to my own question.  Before when I was flipping between branches, the behavior wouldn't update on the first compilation and the resulting library would act as if I had compiled on the previous branch.  I think this happens because the makeall program is recompiled during the first compilation process and the changes by switching between the `develop` and `makeall_fix` then take effect when `make.code superdarn rst` is run for a second time.

So a better way to test this might have been to compile the makeall.x.xx program, and then run `make.code superdarn rst` when changing between the old `develop` branch and the `makeall_fix` branch.

sshepherd:
make.build compiles the necessary programs, which is stated in the README.

PR #23, opened 2017-02-28
Removing java functionality from RST

egthomas:
This pull request removes all src.java files and folders from the codebase directory.  Basically, this amounts to removing a java applet for performing AACGM conversions (which uses outdated/incorrect coefficients and is being replaced by AACGM_v2) and some real-time capabilities (which we're abandoning for Keith's new real-time software).

The deleted files do not tie into any of the other RST routines.  To test, create a fresh clone of the repository (to avoid leftover untracked files from previous compilations), switch to the java_cleanup branch, and execute `make.build` and `make.code superdarn rst`.  Everything should compile successfully.


egthomas:
I should also note that these files were removed from AJ's RSTLite as well, so there is a >4-year history of users safely working with the RST with java functionality removed.

ksterne:
Same basic test of general compiling ability?  Also, is gutting out the real-time code good here?  Does any of it relate to the real-time grid/map generation that was available on the old APL website?  Otherwise, it does seems as though @kkotyk has a real time radar data working.

egthomas:
Yes, the only test is to make sure you don't get any compilation errors (again, I highly suggest testing this using a fresh clone of the repository to avoid conflicts with untracked libraries/directories).  The java real-time software has been useless for at least 3-4 years due to all of the security vulnerabilities associated with it so I don't see a problem with removing it.  As with the other deletions, we can always resurrect this later from the v3.5 tag if necessary.

The real-time gridding was done via rtgrid which is in C; all it needs is incoming radar data via TCP/IP.  Even so, Keith has found major flaws with the existing real-time software and has entirely re-written it for displaying fit-level data.  We plan on doing the same thing for gridding eventually (map potential doesn't care if the data came from rtgrid or make_grid so that's done in a post-processing manner).

kkotyk:
I can confirm that this builds correctly(without IDL) within the Ubuntu 14.04 and OpenSUSE 42.2 Docker containers.

egthomas:
I can also confirm that this builds correctly (without IDL) using Debian 8.7.1 and Mint 18.1 virtual machines.  As no remaining RST functionality was modified and compilation has been confirmed on 4 different Linux distributions by 2 different users, I suggest we merge this pull request.

PR #24, opened 2017-03-01
get.version fix

sshepherd:
This change is the the script get.version which is used in the RST build process to determine the version number of the different components in the RST structure. Currently the version number is determined by reading a file in a hidden directory in each of the component directories, i.e., .rst/version. These hidden files were created as part of the original RST packaging mechanism (rpkg). By removing the dependence on rpkg we simplify the build process and allow for additional components to be added to RST more easily.

In addition, the .rst/ directories have been removed from the RST directory structure. In addition to the version file there was some log information from the original developer, Rob Barnes. If it is thought that this information is necessary in the current version of RST it can be added, however, it will not be updated as we move forward.

The test is to do a fresh clone, delete all of the files in lib/ (if they still exist), run make.build and make.code, then (1) check the log file to see that everything built and that the log file has a version number, (2) look at the lib/ directory to see that version numbers were created, check as many of the src/ directories as you like to see that the version.h files was modified and has the correct version number.


egthomas:
I can confirm that this builds correctly (without IDL) using Debian 8.7.1, OpenSUSE 42.2, Mint 18.1, and Ubuntu 16.04.2 virtual machines and all of the version.h files I've checked have been accurate.  The old log files don't seem to be a big loss as they are incompatible/unnecessary given our current collaborative git development process.

ksterne:
I'm guessing this is just a follow up from #25 where all of the log, id, version and any other version controlling notes are removed?  In some spot checking, it doesn't seem like there's much documentation that's lost from the commit messages.  I agree that this type of version controlling is now obsolete now that this has moved to github.  Merging this in now.

PR #27, opened 2017-03-03
Compile without IDL access by removing DLMPATH

egthomas:
This pull request modifies make.code to allow automatic compilation/recompilation without access to IDL by checking the length of the DLMPATH environment variable.  This way, users can compile the RST with or without IDL by modifying a single line in idl.bash (or idl.tcsh).

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ your computer.
    In order to load the environment variables you just setup, you'll need to close your current terminal and
    open a new terminal.
 
+   Note: to compile without any IDL dependencies, comment out the DLMPATH entry in ~/rst/.profile/idl.bash
+
 2. Run `make.build` from the command line.  This runs a helper script that sets up other 
 compiling code.  The source code for make.build can be found in ~/rst/build/script/
 
--- a/build/script/make.code
+++ b/build/script/make.code
@@ -168,6 +168,30 @@ cp -vp *.a ${LIBPATH}
 # Get back to the old directory that we were in before
 cd ${currentdir}
 
+# This modification allows for automatic compilation for systems without IDL
+if [ ! -n "${DLMPATH}" ]; then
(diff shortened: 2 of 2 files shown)


kkotyk:
Alternatively you can leave the scripts as they are and run "unset DLMPATH" on the cmd line or in your .*rc file.

egthomas:
I don't think that will work because make.code will still try to compile all of the libraries in codebase/superdarn/src.lib/tk/idl and crash because it doesn't find the idl_export.h file (like you noticed while testing compilation in the ros_cleanup comment chain).  Users need to unset the DLMPATH environment variable and somehow remove those libraries for compilation to succeed.

kkotyk:
Yes. What I was trying to say is that if you use unset DLMPATH, you dont have to modify idl.bash or idl.tcsh

egthomas:
Ah yes, very true.

egthomas:
So I haven't tried using the docker files yet, but I am able to compile this branch on an openSUSE (64-bit,, v42.2) virtual machine without having IDL installed by unsetting the DLMPATH environment variable.

egthomas:
also succeeds on Debian 8.7.1 virtual machine

egthomas:
also on a Mint 18.1 Cinnamon (64-bit) virtual machine.  That makes 4 different Linux distros I've used to successfully test compilation of this branch (Ubuntu, openSUSE, Debian, Mint).  It would be good if someone else could try this out so we're all able to more easily test the various pull requests.

kkotyk:
I confirmed this works in the Ubuntu 14.04, OpenSUSE 42.2 and Debian 8.7 Docker containers by just commenting out the line that removes the idl directory. I would say this branch is ready to merge.

sshepherd:
Has been tested by multiple people on several different distros.

PR #28, opened 2017-03-04
Feature/map potential plotting mods

sshepherd:
The color tables used in the RST plotting software (grid_plot, map_plot, etc.) require the full path for a color table key file, which are located in the tables/base/key/ directory. I have added an environment variable (COLOR_TABLE_PATH) to ./profile/superdarn.bash and ./profile/superdarn.tcsh which has this path. Changes have been made to the .c files in order to use this environment variable. A test directory was add which includes a script (test.sh) and a small grid file. This will be the basis for testing features that are added to map_potential.


sshepherd:
Too many commits to ever get accepted at the current rate. Canceling this request.

ksterne:
@sshepherd, I don't understand here.  How is it possible to have too many commits?  I'm guessing #31 replaces this pull request?

sshepherd:
Dependencies with other branches and pull requests is becoming complicated.

ksterne:
Ah, I did notice a good bit of inter branch merging here.  Was this so that you could continue to work with code from this branch?  If so, I might suggest creating another branch to copy one in a pull request that you don't push but you can keep cherry-picking commits if you want.  Otherwise, I don't think you would need to keep this branch up to date with the current develop unless you had merge conflicts going on.

PR #31, opened 2017-03-07
Feature/colortable fix

sshepherd:
The color tables used in the RST plotting software (grid_plot, map_plot, etc.) require the full path for a color table key file, which are located in the tables/base/key/ directory. I have added an environment variable (COLOR_TABLE_PATH) to ./profile/superdarn.bash and ./profile/superdarn.tcsh which has this path. Changes have been made to the .c files in order to use this environment variable. A test directory was add which includes a script (test.sh) and a small grid file. This will be the basis for testing features that are added to map_potential.


egthomas:
I've tested and confirmed that compilation of this branch works on Debian 8.7.1, Mint 18.1, OpenSUSE 42.2, and Ubuntu 16.04.2 virtual machines (and my Ubuntu 16.04 desktop).

General question - am I prohibited from merging Simon's pull requests even though we develop independently on different machines with different clones of the repository?

sshepherd:
I would say no, but I suspect this comment will draw the compulsory reaction.

pasha-ponomarenko:
Methinks you meant "compulsive", Simon...
Since we have got these docker/container things involved in checking different flavours of Linux (thanks, Keith!), from my unqualified viewpoint they look sufficient for making sure that there will be no immediate craches (please correct me if I am wrong here!). However, the self-approval by people from the same institution still does not sound right, and I feel that at least a cursory sanity check on the essence of the proposed changes by someone from outside is still necessary.

ksterne:
@egthomas merging @sshepherd's pull request or vice versa does pass the one other tester we put forward, but like @pasha-ponomarenko, it doesn't sound quite right.  The one thing reason I could see getting another institution to test is that you may be using the same source data or general setup.  Otherwise, it could also make the codebase institution-centric.

egthomas:
I can appreciate the desire to avoid an incestuous review process, but I don't see how my testing someone else's pull request on 4 unique Linux distributions via virtual machine (on a personal laptop) or using docker files falls under the category of being institution-specific.

ksterne:
Do we have expected results from running `~/codebase/superdarn/src.bin/tk/testing/map_potential.1.0/test.sh` ?  I see a bunch of numbers printing out and there's not really an expected print out versus what I got.  Also, I see I've got 4 graphics that were generated, is there an expected result for these graphics to compare to?

sshepherd:
It's pretty simple. As stated in the first message the only change here is to add color to figures. Do you see colors in the figures? If so, great, it worked. Compare it to something you get from your current distribution and you will see that they are black and white. I added the script and test data only to make things easier. You can test it on whatever data you want.

ksterne:
Sounds good Simon, I'm definitely seeing colors.  I'm not as familiar with generating plots from RST, so this just looked like a regular plot to me.  It doesn't sound as though you're concerned about the numbers here, so I'll merge it in.

PR #33, opened 2017-03-10
Improved automatic detection of whether to compile with/without IDL

egthomas:
This pull request modifies the make.code script to test for the presence of the idl_export.h within the $IDL_IPATH directory rather than testing if the $DLMPATH environment variable is set (as was implemented in a previous pull request).  This has the advantage of not requiring any additional steps for users to compile the RST if they don't have access to IDL.

A change has also been made to the way the symbolic links are created within the idl/ directory.  There was a small bug in the previous version where a dead link would be created within the codebase/superdarn/src.idl/app directory if you tried compiling multiple times.  This is now corrected by first testing whether the symbolic links have already been created during a previous compilation of the RST.

I've tested and confirmed that compilation still works without IDL on Debian 8.7.1, OpenSUSE 42.2, Mint 18.1, and Ubuntu 16.04.2 virtual machines on my laptop.  I've also verified that the symbolic links are now properly handled after multiple compilations on my desktop (Ubuntu 16.04) which has access to IDL.

### DIFF
--- a/build/script/make.code
+++ b/build/script/make.code
@@ -132,7 +132,8 @@ fi
 
 
 # This modification allows for automatic compilation for systems without IDL
-if [ ! -n "${DLMPATH}" ]; then
+if [ ! -r ${IDL_IPATH}/idl_export.h ]; then
+  echo "Failed to locate idl_export.h - will not compile IDL libraries or DLMs"
   if [ -d ${RSTPATH}/codebase/superdarn/src.lib/tk/idl ]; then
     currentdir=(`pwd`)
     cd ${RSTPATH}/codebase/superdarn/src.lib/tk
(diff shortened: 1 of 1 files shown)


sshepherd:
Seems like an obvious improvement for those without IDL.

kkotyk:
This worked for me using the Docker images. I think we can merge it.

ksterne:
Did this pull request result in not needing to comment out the DLMPATH environment variable?  I think that's what you meant between the lines, but just wanted to be sure.  If so, I can update the `README.md` file as the note to do this is no longer necessary.  Just tested this with this code in develop.

kkotyk:
I tested without unsetting the DLMPATH and it worked.

egthomas:
Yes, this pull request removes the need to unset the DLMPATH.  The readme can be updated accordingly, however I think we're going to need to update a majority of the readme before we publish the 3.6 release.

ksterne:
OK, I've hit develop with the hotfix commit to the `README.md` file.  Agreed, updating this file can be a good push for the 3.6 release.

PR #35, opened 2017-03-10
Moving dmap format conversion routines from tool/ to reformat/

egthomas:
This pull request simply moves a few command line functions (cnvmaptomap, grdmaptogrid, gridtogrdmap, maptocnvmap) from the `codebase/superdarn/src.bin/tk/tool` directory to the `codebase/superdarn/src.bin/tk/reformat` directory where the dat/raw and fit/fitacf conversion routines are already located.  There is also a change to one line of the gitignore file so that modifications to files within doc/ subdirectories are still tracked by git.


sshepherd:
Any idea why these functions were put in this directory? I would agree that a more natural location is in the reformat directory.

egthomas:
No idea.  I tried looking at the since-deleted log files but they didn't contain any useful information.

ksterne:
I see that everything here compiles.  However, I haven't test any one of the three binaries that had their source code moved.  Should we be worried about whether the binaries changed?  I would think not if the content of the c code said the same, but just directory paths changed.  Just wanted to make sure before merging.

egthomas:
Eh I can't speak to whether these binaries worked before (I have used grdmaptogrid and gridtogrdmap pretty extensively on different branches/repositories) but as you noted the content of the C code hasn't changed.

ksterne:
Sounds good. I'll merge this in then.

Issue #29, opened 2017-03-06
oldfitdlm not working (opening/reading fit files in IDL)


egthomas:
Copying this over from the original issue raised on the VTRST3.5 repository (https://github.com/vtsuperdarn/VTRST3.5/issues/15).  Basically, there is a problem with the use of 'strncpy' in the oldfitdlm routines for opening/reading fit (not fitacf) files in IDL.

> I first ran into this issue several years ago at VT (~2012) and now again at Dartmouth after making a fresh installation of VTRST3.5 from github. Calling the IDL routine OldFitOpen on a .fit file results in a buffer overflow immediately after it loads oldfitdlm, however if you explicitly use the IDL function and not the DLM then everything works normally. Similarly, using the IDL routine to open a .fit file and then trying to use the DLM to read the file (via OldFitRead) fails with a returned value of -1.

below is a follow-up attempt to track down the issue, identifying it as being related to strncpy.

> So after browsing through the log files due to the sim_real compilation error in the other thread, it seems like this particular issue is due to the use of 'strncpy' in OldFitIDLtoFitFp and OldFitFitFpToIDL (Lines 53 and 73 of oldfitdlm.c):

> In file included from /usr/include/string.h:640:0,
from oldfitdlm.c:14:
In function ‘strncpy’,
inlined from ‘OldFitIDLToFitFp’ at oldfitdlm.c:53:10:
/usr/include/x86_64-linux-gnu/bits/string3.h:120:3: warning: call to __builtin___strncpy_chk will always overflow destination buffer [enabled by default]
return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
^
In function ‘strncpy’,
inlined from ‘OldFitFitFpToIDL’ at oldfitdlm.c:73:10:
/usr/include/x86_64-linux-gnu/bits/string3.h:120:3: warning: call to __builtin___strncpy_chk will always overflow destination buffer [enabled by default]
return __builtin___strncpy_chk (__dest, __src, __len, __bos (__dest));
^

and here is the most recent attempt I made to fix the error (17 Nov 2016):

> Changing lines 53 and 73 of rst/codebase/superdarn/src.dlm/oldfitdlm.1.7/oldfitdlm.c from

>strncpy(fitfp->date,ifitfp->date,80);

>to

>strncpy(fitfp->date,ifitfp->date,32); (to match the size of the date value in the oldfitfp structure)

[...]

PR #37, opened 2017-03-17
Buffer overflow fix for oldfitdlm

egthomas:
This pull request addresses the issue identified in #29 by correcting the size of the date string provided to strncpy (which previously would always result in a buffer overflow).  I had already identified this fix in #29, but I wasn't convinced that it was sufficient since a few other values in the prm structure such as bmazm were odd-looking values like -999.0.  After digging deeper into the code, I realized that bmazm is hardcoded to -999 for .fit files and is only calculated for fitacf files from the radar site information.

The other commit on this pull request addresses several warnings upon IDL startup about undefined calls to 'pickfile' in the legacy fitlib, istplib, and rawlib .pro libraries.  This function was "obsoleted" in IDL 5.0 and replaced by dialog_pickfile (http://northstar-www.dartmouth.edu/doc/idl/html_6.2/Obs_in_50.html).

To test, load the oldfitdlm in IDL and execute something like `fitfp = oldfitopen('/path/somefile.fit')` and then `ret = oldfitread(fitfp,prm,fit)`.  If you don't get kicked out of IDL by a buffer overflow / core dump then you're halfway there.  Next you can compare the contents of the prm and fit structures to the output of `fittofitacf /path/somefile.fit > /path/somefile.fitacf` and then `dmapdump -d /path/somefile.fitacf > /path/somefile.txt`

### DIFF
--- a/codebase/superdarn/src.dlm/oldfitdlm.1.7/oldfitdlm.c
+++ b/codebase/superdarn/src.dlm/oldfitdlm.1.7/oldfitdlm.c
@@ -50,7 +50,7 @@ void OldFitIDLToFitFp(struct OldFitIDLFp *ifitfp,struct OldFitFp *fitfp) {
   fitfp->etime=ifitfp->etime;
   fitfp->time=ifitfp->time;
   strncpy(fitfp->header,ifitfp->header,80);
-  strncpy(fitfp->date,ifitfp->date,80);
+  strncpy(fitfp->date,ifitfp->date,32);
   strncpy(fitfp->extra,ifitfp->extra,256);
   fitfp->major_rev=ifitfp->major_rev;
   fitfp->minor_rev=ifitfp->minor_rev;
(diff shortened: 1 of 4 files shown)


egthomas:
I've since realized that pickfile.pro is included in the idl/lib/obsolete directory, but still think it's ok to replace the call to this obsolete function with the updated version (even if no one uses these legacy libraries anymore).

ksterne:
Now this is what I'm talking about with a test plan!  I see the core dump on the current develop branch when running the `oldfitopen()`.  This branch fixes that though.

I don't know I have a good way to do the comparison between the IDL loaded structures and the dmapdump of a converted file here.  I'd think small differences could have also been introduced in the fittofitacf code.  However, the fact that the oldfitopen() call doesn't cause IDL to core dump, I'll call that a win!

Merging in now.

egthomas:
Note that you can probably remove the patch I added to rad_fit_read.pro in davit for handling .fit files.

PR #38, opened 2017-03-20
Modifying igrf_date variable in igrfcall.c to resolve naming conflict

egthomas:
This pull request addresses a naming conflict between the igrf_date variables in igrfcall.c and igrflib.c (of the IGRF_v2 library).  This commit is necessary for any future attempts to link both the old and new IGRF libraries for the same binary.


ksterne:
Simple "does it compile?" a good enough test here?  Seems as though the change is small enough.

egthomas:
I believe so.  The library with the conflicting variable name (IGRF_v2) isn't yet being accessed by any functions or routines on the develop branch so there should be no differences in functionality.

PR #42, opened 2017-03-28
Adding new function IGRF_Tilt

sshepherd:
Adding a new function IGRF_Tilt() to return dipole tilt angle and eliminate dependency on geopack or other packages.

### DIFF
--- a/codebase/analysis/src.idl/lib/astalg.2.0/astalg.pro
+++ b/codebase/analysis/src.idl/lib/astalg.2.0/astalg.pro
@@ -2,7 +2,7 @@
 ; ==========
 ; Author: R.J.Barnes (Based on C routines by Kile Baker)
 
-;                                                                    
+;                                                                     
 ; $License$
 ;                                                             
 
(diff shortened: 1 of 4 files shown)


sshepherd:
Hey, did I just merge my own request?! I just hit the "confirm" button, as in "yes, I really want to make this request". If someone wants to back it out I am fine with that... Sorry.

ksterne:
Doh!  Careful on that next time!  Would there have been any good test here?  Hard to do any testing without testing this in the develop branch since @egthomas deleted the branch already.

egthomas:
The branch can always be restored if the pull request was approved (accidentally or otherwise).  I'm not sure there's anything to test here anyways because Simon hasn't actually integrated his AACGM/IGRF/MLT _v2 libraries into any of the existing software yet on this branch.  But maybe that's just me wanting to see some of the other 2+ week-old pull requests get looked at...

sshepherd:
Test would be to compile, then call function and compare to know values. I
compared with geopack and had agreement to within 1e-3 degrees. We don't
use this function except for selecting the CS10 model for convection
mapping, so this is a new feature.

ksterne:
Getting to this today.  Of note, I see this undid the readability commit I made in the astalg.pro file with 8ca319b.  So, neat.  I'll try to pork barrel that one into another pull request here.

Since the AACGM_v2_errmsg function no longer exists, should it have been removed from the top documentation?

I don't have test setups that include IDL, nor would I know what an expected value would be here, so I can't really give this a good test.  Once I pull it to our main analysis computer that does have IDL, I can test it there.  Can you give a quick example of expected output for an input?  Ballpark is fine if you're getting 1e-3 error in results.

ksterne:
I feel like that should be possible, but sure enough it doesn't seem to exist.  Wiki says it's around 11 degrees these days.  That sound right?

I'll leave the other small bits for cleanup during the 3.6 release branch since none of the current pull requests include the overwrite of the 8ca319b commit for me to re-correct.  I'm sure there will be some other small things we'll find we want to tidy up for the 3.6 release.

PR #45, opened 2017-04-07
Reorganizing routines in codebase/superdarn/src.bin/tk/ directory

egthomas:
Similar to #35, this pull request simply shifts several reformatting and testing routines from the `codebase/superdarn/src.bin/tk/tool/` directory to the `codebase/superdarn/src.bin/tk/reformat/` and `codebase/superdarn/src.bin/tk/testing/` directories respectively.  Files have only been "renamed" without any additions or deletions to the actual code.  This will help make the RST directory structure more intuitive and less confusing for unfamiliar users.


ksterne:
Tried test_rawacf, test_fitacf, and test_fitex2 here and got mixed results.  But these are the same results as from develop, so the move isn't a problem.

PR #47, opened 2017-04-11
Adding old RFC documentation

egthomas:
This pull request adds the old RFC documentation found in the copy of RST 2.11 at Dartmouth College.  While these files may be out of date in some regards, they contain a lot of useful information which was previously available on the JHU/APL website explaining various file formats and data structures.  The html documentation for these pages can be produced by executing the following commands (skip to step 3 if you've already compiled the RST):

1. `make.build`
2. `make.code superdarn rst`
3. `make.doc.rfc codebase superdarn`
4. `make.doc superdarn rst`


egthomas:
Sorry, forgot to mention you can view these pages at the temporary address below:

https://engineering.dartmouth.edu/superdarn/documentation_rfc_test/superdarn/src.doc/rfc/index.html

ksterne:
Any thoughts on updating some of the outdated material in the RFC docs?  Would be nice to have accurate documentation if we're going to add it in.

Or should the `make.doc.rfc` and `make.doc` commands be added to the install notes?  These could be added as optional steps for the installation since it's not necessary for executing codes.

I can confirm that this branch does compile without a problem though.

egthomas:
I've added instructions to the readme file about how to compile the html documentation.  I can see two ways forward with the RFC docs - either leave them as they are as a historical reference, or make an effort to update them so they accurately reflect the current RST.  I don't know what our long-term plans are for documenting the RST (I think @kkotyk may have some suggestions) but for now I'd just like to make sure we don't lose any of this old 'institutional knowledge'.

kkotyk:
We can start discussing this somewhere else, but my personal favorite is readthedocs.io. You can document the code such that a tool like doxygen will automatically generate documentation, and readthedocs has a hook to github so that when you make new commits, it automatically compiles your documentation on your page on their site.

PR #49, opened 2017-04-12
Removing example .svg, .ps, and .rpz images and download links from documentation

egthomas:
As the title states, this pull request deletes a couple of large (~33 MB) sample .svg images which are used strictly for documentation purposes and can be viewed at the following two pages:

https://engineering.dartmouth.edu/superdarn/documentation/general/src.lib/map/MapBndFread.html
https://engineering.dartmouth.edu/superdarn/documentation/general/src.lib/map/MapPlotOpenPolygon.html

I personally wouldn't be opposed to removing all of the sample .svg, .ps, and .rxml (which seem to have been renamed with an extension of .rpz?) images from the RST.  This would further reduce the download size of this repository and we would still have .png images for visualization.


egthomas:
After further examination I decided to go ahead and delete the remaining .svg, .ps, and .rpz example images.  The download links have also been removed from the relevant build files in `build/doc/cfg/`.  You can compare the difference in html documentation to the above pages with the output of this branch here:

https://engineering.dartmouth.edu/superdarn/documentation_img_test/general/src.lib/map/MapBndFread.html
https://engineering.dartmouth.edu/superdarn/documentation_img_test/general/src.lib/map/MapPlotOpenPolygon.html

The zip file size of this branch is now ~11 MB compared to ~43 MB for the current develop branch.  I think a 4x reduction in download size should be a good thing.

kkotyk:
This is awesome. Is there a needed test here, or can I merge this?

egthomas:
If you want you can run `make.build`, `make.code superdarn rst`, and `make.doc superdarn rst` and check to make sure the figures on the html documentation pages match my output.  In my opinion that's unnecessary since no one has looked at the html documentation since the APL links went dead ~5 years ago, but there may be others here who disagree.

kkotyk:
Okay I can see the files but im having trouble seeing the images using Docker. I'll wait for any other comments but I agree with you.

egthomas:
Yeah I can see how that would be difficult with Docker.  Is there any way you can export the contents of rst/doc/html to something like an external web server?  With that approach you would also have to modify the URLBASE environment variable to get the addresses correct.

kkotyk:
Perhaps. Im going to play around with this for a bit because this should be possible.

kkotyk:
I can confirm that those htmls are generated and look the same.

egthomas:
Thanks!  This should also secure my # 1 place atop the "Deletions" leader board for all eternity.

PR #50, opened 2017-04-20
Removing unused files from tables/ directory as well as outdated, ROS-related frequency and sounder site information

egthomas:
This pull request makes several changes to the tables/ directory of the RST.  I've removed an empty "example" overlay.dat file and the corresponding MAPOVERLAY environment variable which I think was intended to allow users to overlay custom boundaries on plots but was never actually implemented.  I've also removed the convert/ subdirectory from tables/superdarn/ which contained several .cdl and .skt files that appear to be examples for the dmap -> cdf/netcdf conversion procedure.  Finally, this pull request removes the restricted frequency and sounder files from the radar site subdirectory of tables/superdarn/.  These files have not been maintained or updated since the release of RST 3.3 nearly 6 years ago and are much more closely related to the ROS side of things than the RST.  Furthermore, none of the analysis or processing routines in the RST use any of these files.  Ideally they should be stored in whatever ROS repository eventually comes along.

Compilation of this branch has been tested and succeeds on Ubuntu 16.04.  There aren't really any other tests as these files aren't used by any of the other RST code.


egthomas:
I've heard from the OS-WG chair (Mike Ruohoniemi) that he's fine with removing these site files from the RST.

ksterne:
Can confirm that this compiles on my virtual machine setup with a fresh install, so I'll merge it in now.  Hopefully this cleans most everything up here?

egthomas:
I think so, yes. I may have run out of things to delete!

PR #51, opened 2017-04-21
Enhancement/map ascii doc update

ksterne:
After looking at the map_ascii tool, found that the output format wasn't at all documented, so I've added it in with this pull request.

To test, after compiling with this branch, do `map_ascii --help` and see the output here.  The html documentation is a little messy since I had to cheat and manually insert newline characters to get the command line to look nice.  But at least the info is there and looks relatively good.

Also just removed extraneous spaces in the map_ascii.c file.


kkotyk:
I tested this using both the Ubuntu and Opensuse Docker images. Seems to work for me just fine.

PR #52, opened 2017-04-21
feature/grid_v2

egthomas:
This pull request introduces numerous upgrades to the standard gridding procedure (make_grid) used to create grd/grdmap files from fit/fitacf files for input to the Map Potential routine.  The list of major changes is as follows:

- Added Chisham virtual height model
- Added AACGM_v2 functionality as new default coordinate transformation
- Added functionality to load correct AACGM_v2 or AACGM coefficient file according to year
- Removed 50 degree magnetic latitude lower limit for binning vectors
- Added capability to filter input data by slant range
- Added native IDL gridding
- Major revisions to documentation

Several other minor bug fixes and modifications are listed in the commit history.  These new features have been fundamental to my current research project (development of a new statistical convection model) and have been used to generate more than 10,000 grid files over the last several months.

To test, first make sure that the feature/grid_v2 branch compiles.  Then try out the new options (eg `-chisham`, `-minsrng 800`) to make sure you get sensible results.


ksterne:
Can you post some example figures (with example make_grid options) that try out these new features and how they compare to the current develop branch?  Maybe comparison to current develop isn't so good, but some kind of comparison in results would be good to make sure someone else is getting the same results.

egthomas:
I'll try to work up some example figures over the weekend using the C plotting code (since not everyone will have access to the same IDL/python routines).

egthomas:
The plotting side of things has become more complicated than I expected - mainly because Simon and I have each been introducing new features on different branches and I want to keep the conflicts to a minimum.  I may not be able to wrap this up until next week after returning from Canada.

ksterne:
Hey @egthomas, sorry, I didn't necessarily mean something native to the RST codebase as I'm not looking to compare pixel by pixel to make sure it's right.  That's not what you're trying to do with this pull request.  I meant something that I can eyeball and maybe get a few parameters (N_points, CPCP, etc.) off of to see that they match.  Is that something you might be able to get to?

pasha-ponomarenko:
I think here I have to intervene. The stuff Evan is talking about is going well beyond the basic refactoring/optinisation process. It introduces some new procedures which most probably alter functionality of the gridding/mapping code. With all due respect, in this case we need to be more cautious with going ahead with in cluding these changes to the master branch and should try to follow the steps prescribed in the WG charter. They are there for a good reason: we have to see what is done and what it changes with respect to the previous code. I might be mistaken, but I ccannot recall seeing comparisons of the  results from the "new" (2008?) virtual height model or removing the 50 deg limit.

pasha-ponomarenko:
Again, please correct me if I am mistaken!

pasha-ponomarenko:
I should avoid typing directly in the comment window -- I tend to submit the text accidentally before it is ready... :-(

egthomas:
Hi Pasha - as you'll see in my talk on Wednesday (tomorrow? what time is it?) I've done a fair amount of statistical analysis of grid vector distributions using the different new gridding options with >6 years of data (~308 million vectors). We can talk about the appropriate way to document all of this for the purposes of the DAWG once I arrive.

Don't worry - I'll show more than just a few geographic plots of grid vectors!

sshepherd:
Pasha,

Evan is on his way to visit you, so this is good timing. He will explain in
detail during his visit and at the SD Workshop the new features he has
implemented for gridding.

I just reread the WG charter and am wondering what specifically you are
referring to when you say we need to be more cautious? Evan has made some
bug fixes and added a lot of functionality. One can continue to use many of
the old features by setting command line options, or even still use the
older software. As you will see, Evan has used the new features extensively
in deriving new statistical models and the features he has added make a big
difference.

In terms of the 50 degree limit, I first showed this issue at least six
years ago at the Dartmouth SD workshop and pointed out the exact lines of
code in the RST that needed to be changed. I think we can all agree that
moving scatter which occurs below 50 degrees to 50 degrees is worse than
just excluding it all together, but the best thing to do is to put it at
the location at which it occurs.

I am not objecting to more testing, but I think that others (not in this
WG) will want to use the new features and will significantly expand the
testing base by just using the new software.

I suspect you will be busy over the next few days with Evan. Enjoy!

Best,
Simon


On Tue, Apr 25, 2017 at 1:11 AM, pasha-ponomarenko <notifications@github.com

pasha-ponomarenko:
I just wanted to make two points, which is apparently not too easy to do near midnight:
1)	I would like to draw a demarcation line between technical changes like refactoring/bug elimination and algorithmic changes. In that respect, it makes sense to separate these two things in releasing first the refactored and easy-to-install RST4.0 replicating the functionality of RST3.5, which will provide a basis for the significantly modified/extended RST4.1.
2)	The authors have to explain the objectives of the proposed algorithmic changes and present proofs that they are achieved based on extensive testing. This one follows from the implementation procedures, not from the Charter (my apologies, Simon!).
 
While the second point, I am sure, will be exhaustedly covered by Evan’s presentation (by the way, it’s at 2:30pm, Evan!), the first one seems to be missing. I am concerned that without setting these milestones, we might gradually sink into a perpetual piling-up of changes. In my opinion, each version should have a clear goal or two, but there should not be too many of them, otherwise we will end up DavitPying! 

Right now I have a feeling that right now we should be ready for releasing 4.0, but producing 4.1 as an official version before the workshop seems to be a bit of a stretch. Making it available for trials to other people is a different thing, and I am all for it!

pasha-ponomarenko:
One more thing: how would one test reliablilty of velocity/E-field mapping? 
Ideally, one would start with a model circualtion map, convert it into velocities observed by the radars and then process these velocities in order to see if the output pattern matches the initial one. I have a felling that someone has done it before, but I don't really remember...

egthomas:
Hi Pasha - I was more asking what the current time was when I wrote that reply, as I was typing on my phone while riding an airport shuttle sometime before 4am and had just woken up a few hours earlier. Now that I've had my Tim Horton's I am much more alert.

Anyways, I don't want to spoil too much of my talk for tomorrow so I'll leave you in suspense. I would be hesitant to label the reorganization/simplification we've done so far as "refactoring" though. That would be a disservice to the actual refactoring you guys have done with fitacf.

pasha-ponomarenko:
Well, I guess that you are this side of the border by now. It's exactly 1pm here in Saskatoon. 
;-)

pasha-ponomarenko:
"the actual refactoring you guys have done with fitacf"
It might a treminology problem here. The actual refactoring happened in what we called FITACF2.6 which exactly replicated the last official release, FITACF2.5. It has not been released for a very obvious reason that, in contrast to RST, there were no issues with its installation/compilation as long as one has a working RST. ;-)

FITACF3 was written from scratch, and it is still being tested.

ksterne:
Hey @egthomas,

Are you still good with this pull request being open?  Are there still more commits to make?  Each commit you're pushing here is spamming some inboxes.

Also, do you have some figures and/or example code to run for comparison to do a test on this?

egthomas:
I'm fine with spamming people's inboxes - maybe it will get some more activity going :)

This branch should be fairly complete at this point.  Today's round of updates was motivated by a request from Nishitani-san for a copy of RST with the Chisham virtual height model - I realized I hadn't included the 'chisham' keyword option for rbpos in IDL, which also required me to go back and actually implement that keyword in the rposdlm.  I was also able to track down a solution to the myriad of compilation warnings about missing braces which show up for all of the DLMs; most of those were fixed on the `warning_cleanup` branch but a few were more appropriate for their respective outstanding pull requests.

Example code for testing will take some time to produce.  The simplest test would be to produce a grid file using some of the new options (`-chisham`, `-minsrng`, etc) and see if it works.  Another simple test to see if the Chisham virtual height model is correct would be to call one of the plotting functions with the `-chisham` flag and see if you get the expected discontinuities at the 0.5E / 0.5F / 1.5F region boundaries.

sshepherd:
I agree, if pull requests are sitting around, updates should be made if
they are identified. It's hard to tell if anyone is even looking at a
request.

ksterne:
Finally carving off some time to test this while putting off things that are months behind.  First test passes that this compiles.  Next I tried out an example test period and maybe got some expected results but I'm not sure since I haven't seen anything (at least not in a while). Maybe someone can help confirm the results.  The first is the `develop` branch (or at maybe sometime pre 4.0):

20171013 1531-pre

With this pull, the grid seems to shift "up and to the right" a little or maybe counter-clockwise centered somewhere around Greenland as seen below.  I'm guessing this is a change in the AACGM codebase?  Does this look right?

20171013 1531-pro

* I made these with VT's 'Map Velocity Tool' with grid velocities on and AJ's filter off.

I tested the `-chisham` option and from what I remember of Ray's talks, this should bring the scatter closer to the radars which I can see.  So that seems to work.  Confirmed the `-minsrng` functionality by eye and as well that it will override `-minrng` if it is included.  Also see this with the `-maxsrng` / `-maxrng` pairing.

Minor afterthought here, seems as though the `develop` branch grid software is putting out a 0.91 version number even though it contains `make_grid.1.19`.  I guess at some point there was a 1.x version, but it never got updated on the dmap record; we're good with bypassing this in the dmap record?  I think it would be confusing to have the directory version number (again something that doesn't completely make sense in the git world, but another story) different than the file output.

One last comment, would it be possible to put units on the `-minsrng`, `-maxsng` options?  I'm guessing these are meant to be kilometers, but it would be good to have this documented so to not lead to confusion later.

I missed it from the meeting, but when this goes through, when should I be reprocessing grid files back to?  As far back as possible?

egthomas:
The primary reason the figures look so drastically different is because the grid data in the first plot is in "old" AACGM coordinates using year 2000 coefficients and the coastlines are drawn using "old" AACGM coefficients.  In the second plot, the grid data is in AACGM-v2 coordinates with year 2016 coefficients, but the coastlines are still drawn using "old" AACGM (year 2000) coefficients.  While the grid data vectors may still move from one plot to the next, they shouldn't move nearly as far relative to a fixed geographic location / coastline.  So ideally you should set the "v2" keyword in the call to map_plot_panel and that should be passed all the way down to the underlying cnvcoord function.  It might be slow but you can speed that up by using the feature/mlt_v2_dlm and feature/aacgm_v2_dlm branches.

Regarding the version numbers, the major and minor version numbers written to grid files are actually set in a gtable.h file within the gtable library (https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.lib/tk/gtable.1.14/include/gtable.h).  So the version number is related to the underlying library (gtable) and not the binary (make_grid), similar to how the version of FITACF we refer to (currently 2.5) is the version number of the library (fitacf) and not the binary (make_fit).  I do agree there is an inconsistency with how it is implemented on the develop branch so that's why I've updated the version number on the feature/grid_v2 branch from 0.91 to 2.0 in the gtable.h file, and I've also updated both the gtable library and make_grid binary to 2.0.

Sure, I can add units to the minsrng and maxsrng options.  It looks like units are also missing from several of the other filtering options so I can add those to the documentation as well.

Ideally everyone will reprocess all of their grid files using this new functionality.  To maintain backwards compatibility, the -old_aacgm option can be used.  The only differences then will be bug fixes like removing the 50 degree lower limit or using an updated geodetic spheroid model.

egthomas:
Although now that I think about it, the tool on the website may be using year 2010 AACGM coefficients rather than year 2000 - however those will still be quite different from year 2016 AACGM-v2 results.

ksterne:
Thanks for the notes here Evan, though I'm running short on time if I'm going to get to #65 so I don't think I'll be able to modified the underlying code for that plotter.  I'll go with your reasoning here.

It seems as though you were still finding some bugs and typos here.  Are you sure this is ready to go?

egthomas:
The issue addressed by commit 1920acf also afflicts the current version of the rpos library and make_grid in RST 4.0 - ie, not a fix for something my code introduced.  The other changes were a simple misspelling ("Comemnts" vs "Comments") in a comment header block and then a compiler warning I noticed (of which there are literally 300+ others right now in the RST).

ksterne:
So, is this pull request ready to go or are there more things you would like to check?

egthomas:
Not quite yet, I'm tracking down some fringe cases where NAN's are written to grid files by make_grid in both the `master` (ie our existing codebase) and `feature/grid_v2` branches.

ksterne:
Standing by once again.  I'll retest things once you're for sure done.

egthomas:
The issue I've identified (#85) is part of the default behavior of make_grid and was not introduced by anything in this pull request.  Similar to how I've left "fixing" the fmax units which would wreck twofsound data for a later date, I think it's also safe to leave this `-is` flag behavior for a separate pull request.

ksterne:
So, you're good with merging this in now?

egthomas:
There are some very fringe cases where the azimuth to a beam/gate position is calculated as `nan` rather than zero, causing `nan`'s to be written to the kvect and vlos values; I've got two incoming commits to address this.  The first introduces a check for `nan` azimuth values in fldpnt_azm of invmag.c, and the second commit "normalizes" the AACGM_v2 date/time being set in make_grid by dropping the hour/minute/second values for more reproducible results (ie without this, you may see slightly different results depending on the start time you pass to make_grid even for the same date).

egthomas:
I've slightly changed the behavior of make_grid to no longer throw away data if the scan flag is < 0 by default.  Previously, if a user wanted to turn off this feature they had to set the `-is` option (which was entirely undocumented).  Now, users must set the `-ns` option if they want to throw out fit records with scan flags < 0.

Otherwise, after processing 7 years of NH data the only two cases of `nan` values observed in grid records can be traced back to a corrupted rawacf file (20160202.0201.00.sto.rawacf) and a questionable rawacf file (20120124.0200.03.pyk.rawacf) which has multiple `nan` and `inf` values in the affiliated fitacf file.

ksterne:
So is this good for retesting and merging in or do you have more tweaks here?  I'll try to find time today or more probably tomorrow to re-test and merge in.

egthomas:
Yes I think this is ready for re-testing.

ksterne:
Hey @egthomas, sorry for being slow to test this again, I was out on vacation and am under a crunch for getting things ready to make more repairs to fhw prior to the eclipse.

With the changes since I posted a couple of results, are you expecting some of the scatter to move slightly?  As well, it seems as though I'm getting fewer points now.  For comparison, here is what I'm getting now:

20161013 1531-pro2

It's probably hard to eyeball, but namely I see the three lowest latitude scatter points over Labrador shrinking to two.  Another notable difference that was introduced in the last three weeks is the scatter for ADW appears to be moving towards the radar in some cases.  Just wanted to make sure this is expected behavior as it didn't seem like anything should have changed from the discussion here.

egthomas:
Yes, the scatter will shift in range because I changed the code in be3d2f8 to use the receiver rise time from the hardware file rather than the default value of 100 us hardcoded within the ROS and written to the rawacf files.

egthomas:
However this shouldn't be merged yet - there is some behavior only becoming apparent when processing 7+ years of data that I want to investigate.

ksterne:
Sounds good Evan aside from now needing to wait and possibly retest again.  Please let me know when you're done completely with this.

egthomas:
The problem I've found is also present in the current develop branch.  Setting different start and end times for gridding the same fitacf file with the same options will result in different numbers of vectors being included in the output grdmap file:

```
ferguson:~$ make_grid -i 120 -tl 120 -st 19:52 -et 19:60 -vb -new -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:52:0 19:54:0 pnts=1 (209)
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=15 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=21 (209)
ferguson:~$ 
ferguson:~$ make_grid -i 120 -tl 120 -st 19:54 -et 19:60 -vb -new -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=10 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=22 (209)
ferguson:~$ 
ferguson:~$ make_grid -i 120 -tl 120 -st 19:50 -et 19:58 -vb -new -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:50:0 19:52:0 pnts=1 (209)
Storing:2016-1-19 19:52:0 19:54:0 pnts=1 (209)
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=10 (209)
ferguson:~$ 
ferguson:~$ make_grid -i 120 -tl 120 -st 19:50 -et 20:02 -vb -new -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:50:0 19:52:0 pnts=1 (209)
Storing:2016-1-19 19:52:0 19:54:0 pnts=1 (209)
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=10 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=16 (209)
Storing:2016-1-19 20:0:0 20:2:0 pnts=24 (209)
```

egthomas:
Even more troubling, when trying to disable the median filter while debugging the above issue I realized the `-nav` option doesn't actually turn off median filtering but simply passes a single scan into the boxcar filter, so instead of operating on 3 consecutive time intervals it only operates on one with the same cell weight thresholds being applied.  Apparently the user has to set the `-fwgt -1` option to disable the median filter.  But setting or not setting the `-is` and `-fwgt` options appear to all give slightly different number of grid vectors (this is all on the develop branch btw):

```
ferguson:~$ make_grid -i 120 -tl 120 -st 19:52 -et 20:04 -vb -new -fwgt -1 -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:54:0 19:56:0 pnts=3 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=1 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=8 (209)
Storing:2016-1-19 20:0:0 20:2:0 pnts=0 (209)
Storing:2016-1-19 20:2:0 20:4:0 pnts=0 (209)
Storing:2016-1-19 20:4:0 20:6:0 pnts=2 (209)
ferguson:~$ make_grid -i 120 -tl 120 -st 19:52 -et 20:04 -vb -new -fwgt 0 -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:52:0 19:54:0 pnts=1 (209)
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=15 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=22 (209)
Storing:2016-1-19 20:0:0 20:2:0 pnts=26 (209)
Storing:2016-1-19 20:2:0 20:4:0 pnts=0 (209)
ferguson:~$ make_grid -i 120 -tl 120 -st 19:52 -et 20:04 -vb -new -is 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:52:0 19:54:0 pnts=1 (209)
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=15 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=21 (209)
Storing:2016-1-19 20:0:0 20:2:0 pnts=25 (209)
Storing:2016-1-19 20:2:0 20:4:0 pnts=0 (209)
ferguson:~$ make_grid -i 120 -tl 120 -st 19:52 -et 20:04 -vb -new 20160119.ade.a.fitacf > 20160119.ade.develop.grdmap
Storing:2016-1-19 19:52:0 19:54:0 pnts=1 (209)
Storing:2016-1-19 19:54:0 19:56:0 pnts=1 (209)
Storing:2016-1-19 19:56:0 19:58:0 pnts=13 (209)
Storing:2016-1-19 19:58:0 20:0:0 pnts=23 (209)
Storing:2016-1-19 20:0:0 20:2:0 pnts=26 (209)
Storing:2016-1-19 20:2:0 20:4:0 pnts=0 (209)
```

I find it troubling that many fewer grid points are recorded for the `-fwgt -1` vs `-fwgt 0` case (since the value passed through fwgt is initialized to 0 in make_grid).

egthomas:
In case anyone is curious, I've tracked the spurious grid vector numbers down to a bug in the median filter code in `filter.c`.  Basically, when there is a change from a larger (eg 100) to smaller (eg 75) number of range gates along a beam associated with a change in the control program, portions of memory are being improperly accessed for the median filter weight assignment.  Think of it as the center scan having 16 beams and 100 range gates and trying to access the parameter identifying whether scatter is present for gate 80 on an adjacent beam with only 75 range gates.  The unpredictable value being multiplied by that cell's filter weight oftentimes causes it to be "turned on" by the median filter.

The fix is to set the maximum number of range gates for all 3 scans being passed through the filter to the smallest nrang of the beams on the 3 scans.  I went with this approach rather than tossing out the data entirely with an nrang comparison via the FilterCheckOps function in order to retain usable data from the 0-maxrange region.

egthomas:
I guess to summarize, the big takeaway is that everyone's grid files containing data from control program transitions with different numbers of range gates are wrong and contain absolute nonsense vectors.

egthomas:
Here's a fun demonstration (and not meant to pick on the VT site - that's just where I could easily grab the figures from).  Below are some spurious ADE grid vectors from the 19:58-20:00 UT record on 20160119 on the VT server (notice the very large velocities in the 7th column):

ex_grid_file

That's odd.  Let's take a look at where one of those grid cells is located relative to ADE:

ex_grid_cell

OK it's within the FOV (albeit at very far ranges), fair enough.  Now let's check out the FITACF-level velocity data from this interval:

ex_fit_scan

Hm, that's strange, nothing there.  How big of a deal could it really be in terms of all of the grid vectors at this time?

ex_grid_vels

Those 6 big red vectors near the pole at 7-8 MLT are the junk grid vectors coming from ADE.  And finally, the Map Potential output:

ex_grid_map

Given the close spacing of the contours in that region, I'd say those vectors are more than likely contributing to the final solution.  I'd like to reiterate that this issue is affecting everyone's grid and map files generated using either VTRST3.5 or RST4.0-era software.

sshepherd:
Nice sleuthing Evan and very clear description of the problem and
consequences.

pasha-ponomarenko:
Good job, Evan! 

I guess the main purpose of "median" filtering in cleaning out spurious data, but I am not convinced if the 3x3x3 smoothing effect it has on the data (plus all the already detected and still undetected bugs) is an acceptable trade-off...

sshepherd:
Let's not minimize the potential impact of the bug that Evan identified and
fixed. You may be correct that median filtering is not the "correct" thing
to do when gridding our data, but it is the "standard" thing that we do
when using the RST. Turning off median filtering is possible, but not at
all obvious. Until a suitable study is performed that demonstrates your
suspicion, median filtering SuperDARN data when gridding will likely be the
norm. The fact that spurious data vectors with large velocities appear in
an inconsistent manner is quite troubling. This is an important bug-fix.

Simon

pasha-ponomarenko:
>Let's not minimize the potential impact of the bug that Evan identified and fixed. 
...The fact that spurious data vectors with large velocities appear in an inconsistent manner is quite troubling. This is an important bug-fix.

So, let's seat back and contempalte its potential impact for a while...
;-)
I guesse this happens within two scans from the switch to a lower number of gates, but does it happen if the number of gates goes up? I think it should.

egthomas:
This may have also been an issue going from smaller to larger number of range gates, but the fix introduced in bdece6e appears to have pre-empted any of those errors as well.  I've now processed 7 years of grid files two separate times, and both sets of files have the exact same total number of grid vectors.  Before the bug fix, I think I was seeing variations on the order of ~10,000 grid vectors from one processing run to the next of those same 7 years.

ksterne:
So, is this finally ready for a final test before merging?  No more changes you want to do? At this point it could be good for now and any additional changes could be smaller bug fix pull requests.

--------------------
....thumbed from my phone; auto-correct may have improved this message...

----- Reply message -----

egthomas:
Yes, it's ready for testing but I'd like to hold off on merging until more of the Map Potential fixes on `feature/map_potential_v2` are also ready.

ksterne:
You don't have a pull request open for merging the `feature/map_potential_v2` branch right?  I'm not missing something with that?

I guess holding off for now....

ksterne:
Hey @egthomas,

Just wanted to follow up here.  I'm not as familiar with the work going on here, but what in this pull request is dependent on the 'feature/map_potential_v2' branch?

egthomas:
This pull request introduces the AACGM_v2 coordinates into the RST data processing stream for the first time.  Because the higher level `map_*` routines still use the older AACGM coefficients and MLT calculations, I think we should avoid mixing the two.  The AACGM_v2 coefficients are in the process of being implemented in those `map_*` routines on the `feature/map_potential_v2` branch but it is still a work in progress.

ksterne:
Is there a good way to decouple this branch and the eventual pull request for `feature/map_potential_v2`?  If the big thing is the AACGM v2 coordinates, then could this branch revert back to the older ones?  Then, as apart of the `feature/map_potential_v2` branch pull request, the AACGM v2 coordinate functionally back into the gridding process?

egthomas:
That sounds like creating additional work for little to no payoff.  This pull request has already been open for 4+ months - what's the harm in leaving it open a while longer?  There are plenty of other outstanding issues and pull requests for us to address in the meantime.

ksterne:
OK, that sounds good. I was operating in the interest of getting this pull request merged in without it sitting around.  But if you're OK on it sitting around, I'll standby for your instruction.  By the way, I haven't tested the latest bit as I'm waiting for this to be completely ready.

I'll go ahead and run some grid and map reprocessing that I've been holding off on for a month or more for this pull request.

ksterne:
Got around to test this today and I'm seeing something strange here that I don't think is good.  Maybe you can test it out with a fresh install.

So, I'll put in another grumble here about version numbers in the directory structure.  Git doesn't seem to be the best at keeping up with these as I changed over to this branch but still saw the 0.91 version numbers in the grid dmap headers.  I think this is more from how our `.gitignore` file is setup and it's keeping the old object file.  Anyways, so one additional thing here will be to make sure the old gtable.1.14 is cleared off of your system before compiling.

I thought I saw an odd patch of velocities that I couldn't find in the fit-level data, but giving it a few days and then recompiling it seems as though this weird patch is now gone.  So, maybe there is some more funny things going on with my setup.  I'll try to give this some more testing next week.

egthomas:
As I spend more time working on the RST using git I tend to agree that changing the software version numbers via the directory structure is not the ideal way to go about things.  That being said, when testing any new branch I strongly recommend either blowing away your local copy of the RST entirely or cloning the new version to a separate location and modifying the environment variables accordingly.

ksterne:
Blowing away our labs main processing codebase and cloning a new version sounds like looking for trouble (even though it should be OK with git).  To test something like this I can't as easily test it on a virtual machine, especially without testing procedures or examples.

egthomas:
You don't need to touch the lab's central processing codebase.  As I previously mentioned, you can follow the steps below to clone a local copy for testing (pick any of the lab computers except davit):

```
mkdir test_rst
cd test_rst
git clone https://github.com/superdarn/rst
git checkout feature/grid_v2
```

Next modify the `CDF_PATH` environment variable in `.profile/base.bash` to match your  local CDF installation directory (probably `/davit/lib/cdf33_0/cdf33_0`).  Then comment out the following line from your .bashrc file:

```
source /davit/env/env_davit
```

and add the following lines to the bottom of the .bashrc file:
```
export RSTPATH="test_rst/rst"
. $RSTPATH/.profile.bash
```

You can source your .bashrc file or I'd recommend closing your terminal and opening a new one.  Then all you have to do is
```
cd $RSTPATH
make.build
make.code
```

and you should be in business.  I just tested this process on sd-work7 without any problems.

ksterne:
Yes, I know how to clone a repo and build this repo's code.  But then modify existing gridding shell scripts to work with the local environment and collecting or mounting a network for all of the fit-level data in order to run and blanket test the gridding code is the extra layer here.  It's much simpler for me to use the existing setup.

egthomas:
Why are the davit environment variables necessary?  They shouldn't affect the mounting of /sd-data or any other shared drives.  If the RST environment variables are set correctly then `make_grid`, `combine_grid`, etc will already be in your path.

ksterne:
Davit environment variables aren't necessary. But changing output locations.  Also then being able to plot any results.  Again, this is probably all possible, but that requires getting some davit or davitpy routines setup and all of this is available with the current workings.  Not to mention, if this pull request is merged in, it will need to go on our lab's central processing anyways.  So, might as well test it now.

ksterne:
So, @egthomas, as I mentioned in #100, if you use the `-new` flag on this code do things crash/stop for you?  I am using commands that are using `-new` but things appear to run as normal.  Is this a factor of directory changes and I'm not deleting off enough of the old directories when changing between branches?  So far I am removing the old `codebase/superdarn/src.lib/tk/gtable.1.14` and `codebase/superdarn/src.bin/tk/tool/make_grid.1.15` directories before recompiling the code.  Is there another directory that's changed here that I should remove?

ksterne:
Another data point of testing here and what I could get to.  I see where the documentation has been updated here with the --help, but whenever I run:

`make_grid -new -xtd -tl 120 -i 120 -sd 20161030 -st 00:00 -cn A -c 20161030.1201.00.zho.fitacf > out.grd`

with or without the `-new` option, I don't see a difference in the output file.  So, it's almost as though what I was noting recently in #7 has taken place here, but without the deprecation warnings.  But this might be from not being able to or wanting to blow way my codebase and start fresh.  For a single file, I could try this out on a virtual machine to see the difference.  I'll try to do that later or tomorrow.

ksterne:
Yea, the results in my last post seem to be from legacy code from the `develop` branch.  Did a fresh pull of just this branch on a VM and with make_grid -new ...(more options)... 20161030*fitacf`` I get `File not found.`  And then not using the `-new` option results in a seemingly good file.

So if this is merged into develop, the solution for any current users is to remove all rst code and start over?  This will extend to the next release as well.  Can the key code removal pieces be identified so that  notes/instructions can be included with the release on how to upgrade?  Otherwise this is offloading work from us and onto our users on any local configurations they might have.

egthomas:
The vast majority of users download and install from the zip/tar archive rather than github.  I also specifically remember discussions at a past working group meeting leading to an agreement that we don't want our users going through the github interface to obtain the RST package.

I can only reiterate that testing of any pull request should be performed on a fresh install of the RST.  To do otherwise only invites the issues encountered during past testing of pull requests.

ksterne:
Fair enough on not having to rely on github interfacing, I know not everyone is comfortable with git.  Is the phrase more like "the vast majority of *new* users..." or recurring ones though?  Is this your feeling or are there good numbers how users upgrade?  I'm probably the worst to judge since I'm not at a lot of meetings.

I'm not apposed to having to blow away everything just so that I can keep up with the latest `develop`.  My point is this is another release note that we need to add then that users should not upgrade the code, but instead replace working directories with a fresh codebase.  I'm all for making things clear and documented here.

If there's not a good solution aside from fresh install, I think I can't do more testing for this and #100 as I'm at risk of local reconfigs on our main server.  It could be nothing at all, or it could be a bunch of my day debugging and I'd eventually want to go back to `develop` for standard operations.  I'll put effort into thinking of how I can easily/safely test these pull requests on our main server (directory symlinks comes to mind) and reduce the risks involved.  I certainly feel timid on picking up testing until things are resolved in #7 and the result is implemented here.

egthomas:
It seems non-optimal to perform testing of feature branches on the main analysis server at VT which is responsible for running the website relied upon by the broader SuperDARN and space science community.  The `master` branch is the only thing we can reasonably guarantee to be stable; even the `develop` branch can be subject to pull requests and reverts on an unpredictable basis.  Requiring that testing not disrupt the main VT server seems like an unnecessarily high bar for any pull request to meet.

@kkotyk has provided docker files, the Readme lists dependencies for other Linux distributions capable of running on a virtual machine, and earlier in this comment chain I provided advice for installing the RST on a non-davit lab computer.  Beyond that I'm not sure what else to suggest.

ksterne:
I agree here Evan, but the docker files don't provide a test bed for access to all of the SuperDARN files.  Again, doing the testing on the main server provides another data point as merged code will eventually hit this server.  Thanks for the suggestion though, I think there may be a way I can do testing without disrupting too much and reducing the risk.

My thinking wasn't to require testing to not disrupt our server, but more so that it's git-friendly.  Is it a high bar that testing of a git repo is git-friendly?  I understand that this is a bigger problem with the directory structure that's for larger discussions.

PR #57, opened 2017-05-05
Fix fitacf.2.5

pasha-ponomarenko:
These changes fix problems with processing two-channel non-stereo data like the two-frequency mode from Canadian radars, in which different channels are assigned to different frequencies.

### DIFF
--- a/codebase/superdarn/src.lib/tk/fitacf.2.5/src/badlags_s.c
+++ b/codebase/superdarn/src.lib/tk/fitacf.2.5/src/badlags_s.c
@@ -134,7 +134,7 @@ void FitACFBadlagsStereo(struct FitPrm *ptr, struct FitACFBadSample *bptr) {
     offset=ptr->offset;
     if (ptr->channel==1) offset=-offset;
 
-    if (offset==0) return;
+   /* if (offset==0) return; */
 
 
     while ( offset != 0 && i < (ptr->mppul - 1) && k < maxbad ) {
--- a/codebase/superdarn/src.lib/tk/fitacf.2.5/src/do_fit.c
+++ b/codebase/superdarn/src.lib/tk/fitacf.2.5/src/do_fit.c
@@ -203,7 +203,7 @@ int do_fit(struct FitBlock *iptr,int lag_lim,int goose,
         return -1;
     }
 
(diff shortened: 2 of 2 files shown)


egthomas:
These changes look straightforward enough. Unfortunately I'm away from my computer until Wednesday so I won't be able to test this until the end of next week. Are the DA-WG fixes going to be a separate commit or are those a matter of resolving the existing merge conflicts with the master branch?

pasha-ponomarenko:
These changes address issues related to twof/twochannel data not being processed by FITACF2.5. The conflicts you are talking about arose form the fact that the initial WG recommendations were applied directly to the master branch, i.e. the develop branch was bypassed.

ksterne:
I'll try to test them out over the weekend, or if not on Monday.  These could have been directly committed to the release branch though.

pasha-ponomarenko:
> Are the DA-WG fixes going to be a separate commit or are those a matter of resolving the existing merge conflicts with the master branch?

Well, as far as I am concerned, these should be just ignored, i.e. the respective master branch content should be preserved during the merge. I am not sure how to do it technically though...

pasha-ponomarenko:
>These could have been directly committed to the release branch though.

Well, I was not sure if this the right way to do things, but if you wish I can cancel this pull request and do a direct commit.

ksterne:
I'm getting results from the "multi-channel" radar data now:

image

Here I've only processed the 14, 16, and 18 utc files with this pull request.  If you're happy with this @pasha-ponomarenko, then I'm merging it in.

ksterne:
For reference, I also reprocessed the same time with pyk data and got:

image

So there doesn't seem to be any issue with prior code here.  I wouldn't expect it, but just so we know it was checked subjectively.

ksterne:
A random side note to the fix here, I wonder if this will allow UAF data to be relabeled using the channel parameter and then we can distribute one file instead of multiple files (.a, .b, .c, .d)?

pasha-ponomarenko:
@ksterne, this sounds logical, but I am not sure what was the original motivaiton behind their decision to go for separate files. Furhtermore, I am not sure if it is that simlpe to interleave prm and fit arrays while combining two or more files into one.

PR #58, opened 2017-05-10
Removing fitacfT.2.4 library

egthomas:
As described in the commit message, this pull request removes the deprecated fitacfT.2.4 library which has been replaced by the refactored fitacf.2.5 library.  Compilation has been tested and succeeds without this library present.


ksterne:
Is this a bug fix?  If not, I believe this should be saved for post-4.0 release develop branch.

ksterne:
Again, for reference, I'd like to follow this development model:

https://datasift.github.io/gitflow/GitFlowReleaseBranch.png

So we're on the part of the timeline that says "only bugfixes".

egthomas:
We're trying to clean up the current working version of FITACF (2.5) on the release branch.  This pull request simply deletes some extraneous code to reduce confusion for users trying to access that FITACF library.  By leaving this deprecated version in the RST it is also not clear which files are being used to build the html documentation for the FITACF library.  No functionality is affected.

I can see the benefit to following some guidelines for releasing new versions of the software but I don't think we should keep things so arbitrarily rigid that we unnecessarily handicap ourselves.  However if @pasha-ponomarenko disagrees then I'll close this pull request and save it for later.

pasha-ponomarenko:
I think we should be reasonably flexible unless the extra flexibility inflicts some damage. 
@ksterne, what negative effects would be caused by this particular change? 
I am asking also because we were contemplating inclusion of FITACF3 in RST4.0.

sshepherd:
Perhaps just get 4.0 out the door as something that works and is relatively
bug-free. Then you can include fitacf3.0, grinding, mapping, etc. in the
next version.

On Wed, May 10, 2017 at 3:55 PM, pasha-ponomarenko <notifications@github.com

ksterne:
@pasha-ponomarenko, will we know the damage until it's been done here?  The answer to your question is I don't know what negative effects there are.  But the question back to you and others is how confident are we that nothing is using this library and won't break something in the code?

I could see being flexible on removing this as it's extra code, however, I think I agree with @sshepherd that **adding** in code/a feature is crossing the line.  I think the 4.0 release should still be a cleaned up, bug free version, and then we can add in features in later releases.

pasha-ponomarenko:
Ok, I believe that removing the ftacfT2.4 folder would not affect other functions because this has been done with all its predecessors without any noticeable ill effects. 

And I would probably agree on your second point. FITACF3 is a sort of stand-alone package anyways...

egthomas:
Upon further inspection, it looks like the fitacfT.2.4 library is used solely for the `test_fitacf` tool written by AJ.  Presumably the 2.4 library should not be used for any FITACF procedures anymore but for the sake of not breaking anything I'll go ahead and close this pull request.

This is tangentially related to why the ACF Plot and Fit tool on the VT website is now broken.  The IDL code working behind the scenes for that web tool was located in the `codebase/superdarn/src.idl/vt` directory of the VTRST3.5 and has since been removed from this repository.  @ksterne - it may be worth copying those routines into a Davit subdirectory so they can still be accessed by the website.

pasha-ponomarenko:
So, that is where T comes from...

egthomas:
yes, unfortunately AJ never bothered to make that distinction clear

ksterne:
@egthomas, good to see a check here found something random.

Also, care to share what files from the old repo I need to copy over?  I tried a few ones that seemed to make sense, but didn't get a good result.  I'm not sure if blind copying over them all is a good idea.

egthomas:
I was actually mistaken about that too - it only appears broken for fitacf files because the input hour/minute criteria was stricter for test_fitacf.c than test_fitex or test_lmfit.  I've got an incoming pull request to make the logic consistent.

PR #59, opened 2017-05-12
Fixing hour/minute logic discrepancy in test_fitacf.c

egthomas:
This commit changes the logic used to test the input or 'target' hour and minute against the time.hr and time.mt values in the prm block of the rawacf file(s) to match the logic used by test_fitex2 and test_lmfit. The current method in test_fitacf which searches for the exact hour and minute often causes the routine to miss a record which may be only a few seconds after the desired time. For reference, see:

https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/testing/test_lmfit.1.0/test_lmfit.c#L221
and
https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/testing/test_fitex2.1.0/test_fitex2.c#L223

Practically speaking, this should fix the issue on the VT website where the default options for the ACF Plotting Tools page (http://vt.superdarn.org/tiki-index.php?page=ACF+Plotting+Tools) gives you a broken link.  My only doubt is whether the fprintf to stdout statements that I've removed are actually necessary.  I don't think so after looking at the other testing tools and the make_fit routine but someone else should verify that.

### DIFF
--- a/codebase/superdarn/src.bin/tk/testing/test_fitacf.1.0/test_fitacf.c
+++ b/codebase/superdarn/src.bin/tk/testing/test_fitacf.1.0/test_fitacf.c
@@ -208,19 +208,21 @@ int main(int argc,char *argv[]) {
 
   if (vb)
       fprintf(stderr,"%d-%d-%d %d:%d:%d beam=%d\n",prm->time.yr,prm->time.mo,
-	     prm->time.dy,prm->time.hr,prm->time.mt,prm->time.sc,prm->bmnum);
+        prm->time.dy,prm->time.hr,prm->time.mt,prm->time.sc,prm->bmnum);
 
   fblk=FitACFMake(site,prm->time.yr);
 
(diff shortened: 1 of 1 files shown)


ksterne:
Seems as though this pull request doesn't exactly fix everything here.  I'm getting an error on line 113 of the acf_plot_fitacf.pro code, but this is outside of rst.  But, I'm unable to test it for now.

For those that care some on davit, it seems as though the nrang value is being set to zero here with the loaded example, so the intarr() call fails.  I'm guessing there is an error within the readf call earlier in the code.

egthomas:
That must be because I deleted the lines printing the stid, year, month, day, hour, minute, second, and beam which acf_plot_fitacf.pro tries to read at line 74.  After adding those back it should work.

ksterne:
Looks like that got it.  Looks good here so merging it now.

PR #63, opened 2017-05-22
Abandoning negative velocity error values

pasha-ponomarenko:
The negative values of the velocity erros were designed as an indicator that the piece-wise esitmate of the phase slope by `omega_guess` failed due to lack of the consecutive "good" lags (in this case the phase fit is still performed). The above sign change might be useful for testing purposes but it might cause lots of confusion while using fitting errors in generating electric field distribution, and I decided to abandon the negative error values.


egthomas:
Tested and approved.

PR #64, opened 2017-05-25
Fixing some final documentation bugs

egthomas:
I've recompiled the documentation hosted at http://superdarn.thayer.dartmouth.edu/software.html using the release-4.0 branch and came across a few final errors.  The first commit here fixes a broken link, and the second re-introduces the dmap library documentation from an older copy of RST (3.2) here at Dartmouth.  Not sure why the dmap documentation was removed (apparently between releases 3.2 and 3.3), but it caused a lot of links to break.  There may be a few more minor documentation fixes incoming over the next few days; these should in no way affect compilation or functionality of any of the software.

### DIFF
--- a/codebase/base/src.lib/graphic/fbuffer.1.18/doc/fbuffer.doc.xml
+++ b/codebase/base/src.lib/graphic/fbuffer.1.18/doc/fbuffer.doc.xml
@@ -7,7 +7,7 @@
 <name>FrameBufferBezier</name>
 <location>src.lib/graphic/fbuffer</location>
 <header>base/rfbuffer.h</header>
-<syntax>int FrameBufferBezier(<sn href="structFrameBuffer.html">struct FrameBuffer</sn> *ptr, int x1,int y1,int x2,int y2, int x3,int y3,int x4,int y4,float step, unsigned int color,unsigned char m,int width, <sn href="structFrameBufferDash.html">struct FrameBufferDash</sn> *dash, <sn href="strucFrameBufferClip.html">struct FrameBufferClip</sn> *clip);</syntax>
+<syntax>int FrameBufferBezier(<sn href="structFrameBuffer.html">struct FrameBuffer</sn> *ptr, int x1,int y1,int x2,int y2, int x3,int y3,int x4,int y4,float step, unsigned int color,unsigned char m,int width, <sn href="structFrameBufferDash.html">struct FrameBufferDash</sn> *dash, <sn href="structFrameBufferClip.html">struct FrameBufferClip</sn> *clip);</syntax>
 
 
 <description><p>The <fn href="FrameBufferBezier.html">FrameBufferBezier</fn> function plots a bezier curve.</p>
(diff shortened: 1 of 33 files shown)


egthomas:
I think this is where I'll stop with the documentation updates.  A few of the libraries related to gridding and map potential have been revised on the feature/gridding_v2 branch - I've left those alone here for the 4.0 release to try to avoid future conflicts when that branch is eventually merged.

pasha-ponomarenko:
@egthomas, I have a question about one IDL routine which you might be able to answer. Inside https://github.com/SuperDARN/rst/blob/release-4.0/codebase/superdarn/src.idl/lib/main.1.24/radar.pro there is a function `loadradar` which refuses to cooperate. However, when I took it out of `radar.pro` and saved in a separate file `loadradar.pro`, it worked properly. Other fuctions from that source file like `RadarYMDHMSGetSite` or `RadarGetRadar`seem to work OK.

egthomas:
@pasha-ponomarenko - there actually is no `LoadRadar` function in `radar.pro` (even though it is referenced in the header of `RadarLoadHardware`: https://github.com/SuperDARN/rst/blob/release-4.0/codebase/superdarn/src.idl/lib/main.1.24/radar.pro#L240).

The function is actually called `RadarLoad` (https://github.com/SuperDARN/rst/blob/release-4.0/codebase/superdarn/src.idl/lib/main.1.24/radar.pro#L153)

I'm able to get it to work correctly using the following commands:

```
IDL> envstr = getenv('SD_RADAR')
IDL> openr, fp, envstr, /get_lun, /stdio
IDL> network = radarload(fp)
IDL> free_lun, fp
IDL> help, network
NETWORK         STRUCT    = -> RADAR Array[40]
IDL> help, /str, network[1]
** Structure RADAR, 11 tags, length=4312, data length=4176:
   ID              LONG                 1
   STATUS          LONG                 1
   CNUM            LONG                 2
   CODE            STRING    Array[8]
   NAME            STRING    'Goose Bay'
   OPERATOR        STRING    'Virginia Tech'
   HDWFNAME        STRING    'hdw.dat.gbr'
   ST_TIME         DOUBLE       4.3381440e+08
   ED_TIME         DOUBLE       1.6725226e+10
   SNUM            LONG                 0
   SITE            STRUCT    -> SITE Array[32]
```

There is a comment in `RadarLoad` that it functions differently than in the main.1.21 version of `radar.pro` from a past release of the RST (https://github.com/SuperDARN/rst/blob/release-4.0/codebase/superdarn/src.idl/lib/main.1.24/radar.pro#L172) - maybe that's why you're seeing inconsistencies?

pasha-ponomarenko:
Of course, it is `radarload`, not `loadradar` ;-) 
In any case it refuses to read infromation from `radar.dat` unless I put it in a separate file. 
Now I can see that it works with adding `/stdio` to `open` procedure -- I previously used `/get_lun` only based on the example from somewhere. I am in total darkness what difference it makes and why a separate file works without `/stdio`, but thanks anyways!

egthomas:
It could be because there is a `RadarLoad` DLM in the rposdlm.  Maybe by putting it in a separate file you're calling the actual IDL code rather than the C code via the DLM?

pasha-ponomarenko:
I see! I actually renamed the procedure to `radarload_` after putting it in a separate file, so it would indeed call the actual IDL code rather than its DLM counterpart. 

Thanks again!

ksterne:
See that this compiles still.  Not running documentation, so I'll have to take your word for it.

PR #66, opened 2017-05-31
Fixing 'update' keyword bugs in several DLMs

egthomas:
This pull request fixes two bugs identified in the file open functions in several of the superdarn DLMs.  First, the "Keyword UPDATE not allowed in call to: ***OPEN" error message returned when trying to set the
'update' keyword with DLMs enabled is fixed by re-organizing the elements of the kw_par structure into alphabetical order such that the "UPDATE" keyword is listed before the "WRITE" keyword rather than after.
Second, this commit fixes a bug where the iwrite variable was being used for both the 'update' and 'write' keywords in the same file open functions within the kw_par structure of the DLMs.

### DIFF
--- a/codebase/analysis/src.dlm/igrfdlm.1.3/igrfdlm.c
+++ b/codebase/analysis/src.dlm/igrfdlm.1.3/igrfdlm.c
@@ -1,5 +1,5 @@
-/* aacgmdlm.c
-   =========== 
+/* igrfdlm.c
+   =========
    Author R.J.Barnes
 */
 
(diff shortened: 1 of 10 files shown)


ksterne:
@egthomas,  I'm not as familiar with the DLMs here.  Care to give a little bit of an example on how to test this?

egthomas:
If you use the `develop` branch, you'll get the following result when trying to open a fitacf file with the update keyword set:

```
IDL> inp = fitopen('~/20160101.bks.fitacf',/update)
% Loaded DLM: FITDLM.
% Keyword UPDATE not allowed in call to: FITOPEN
% Execution halted at: $MAIN$          
```

Ideally, on this branch you won't get that error.  If any other errors arise, those I can't speak to, since the update keyword has never actually worked before.  I found the explanation about keywords needing to be in alphabetical order here:

http://www.devsuperpage.com/search/Articles.aspx?G=5&ArtID=506243

ksterne:
Hey @egthomas, sorry I wasn't able to get around to testing this.  I don't think there are, but any concerns with needing to reprocess data with this merged in code?

PR #71, opened 2017-06-15
Removing sctnorm binary from codebase

egthomas:
This commit removes the sctnorm.1.2 binary from the codebase/superdarn/src.bin/tk/web directory. Previously, this binary resided in the usr/codebase/.../web directory of the VTRST3.5 but was
moved when the usr/ directory was deleted and its contents were moved into codebase. The sctnorm routine has no documentation and does not operate on one of the standard SuperDARN file types. Documentation for the extract_grid and extract_map routines suggest sctnorm may operate on
the "scatter statistics from grid/map files", but it is not apparent anyone uses these .sct file types anymore and may have only been used for the old JHU/APL website.


kkotyk:
Is there a test for anything here Evan, or is it safe to merge this for you?

egthomas:
I don't think there's anything to test besides compilation; this binary isn't called anywhere else in the RST.

kkotyk:
Compiles just fine for me!

Issue #55, opened 2017-05-01
How are the src and include directories supposed to be laid out in src.lib?


kkotyk:
When I look at the src.lib directories, there are headers in both include and src. What headers are supposed to go in each if I am to add a new library?

egthomas:
The header files in the include subdirectory of the src.lib directories are the ones copied into the main include directory of the RST.  So I think the difference may be whether another function outside the library will need that header information.

kkotyk:
Thought so. Ill structure it that way.

egthomas:
Closing this since it looks like you've figured it out.

Issue #7, opened 2017-02-07
File format transition


egthomas:
Throughout the RST there is the following reference to supporting the dmap file format transition (circa July 2006):

`/* File format transistion`
`   * ------------------------`
`   * `
`   * When we switch to the new file format remove any reference`
`   * to "new". Change the command line option "new" to "old" and`
`   * remove "old=!new".`
`   */`

This was never actually implemented so users still have to set the `-new` flag when dealing with rawacf/fitacf/grdmap/cnvmap format files; by default RST software will assume files are in the older dat/fit/grd/map format.  Since it's been over a decade since the last .dat file was written by a radar (and we presumably won't return to that file format), should we go ahead and make the changes to the software described in the comment above?

I can foresee one argument against this being that it will break too many existing users' scripts, however I think it should be a straightforward modification to make and will be much more intuitive to new students or users of SuperDARN data.

ksterne:
Certainly can see the issue with removing the '-new' flag that will break existing external codes.  Also, what do with do with the "old" option?  Would we remove the '-new' flag and create a '-old' flag?  Does this break even more codes that might not use the '-new' flag?

egthomas:
Right now if someone wants to automatically process SD data they need to determine whether it is a dat/fit/grd/map file or a rawacf/fitacf/grdmap/cnvmap file to know whether to throw the `-new` flag.  If a script can handle that it should be easy to apply the `-old` flag for the opposite case.

pasha-ponomarenko:
OK, I think this is an occasion when I have to wear my WG chair's hat. 
:=O

I am on a conservative side here. There is no urgent need for this change, neither will it provide any extra convenience to the users, much more like the other way around. Lots (most?) of people use RST as a "black box" (myself included), and making any extra modification to their trusted old scripts written years ago by some smart students might be an uphill task. We do not just want to let the others to know that we are working on RST by disrupting their operations, do we? (almost a joke!)

Futrthermore, at this stage we are still looking at refactoring the existing RST3.5 so it would be premature to change the format of the main routines without major changes to the routines themselves. The appropriate place for this would be our future PST (Perfect Software Toolkit), a refcatored, optimised, cleaned, user friendly software utilising comprehensive data analysis algorithms (I must be dreaming!). 

Also, it might be better to implement an automatic recognition of the file type by its extension rather than to rely on poorly informed users in setting the (in)appropriate keywords, and this sort of change would make sense right now.

egthomas:
[...]

PR #74, opened 2017-06-16
Add fitacf.3.0 into RST

kkotyk:
This pull request adds fitacf.3.0 into RST. make_fit was changed to allow for both fitacf.2.5 and 3.0 to be used via a flag(-fitacf-version). Default is 3.0. 

To test, compile and try no flag or -fitacf-version 3.0, and -fitacf-version 2.5 with make_fit. Using -vb will print the version number. These flags should produce different data files.


egthomas:
I'm not sure that the command line documentation was properly implemented here.  When I enter `make_fit --help` I get

```
make_fit - Creates a fit or fitacf format file from a raw (dat) or radwacf format file.
make_fit --help
make_fit rawname fitname [inxname]
make_fit -new [rawacfname]
--help     print the help message and exit.
rawname    filename of the raw (dat) format file.
fitname    filename of the fit format file to create.
[inxname]  filename of the associated index file to create.
-new       input file is in  rawacf file format and the output should be in fitacf file format.
rawacfname filename of the rawacf format file. If this is omitted the file is read from 
           standard input.
```
ie the fitacf-version flag doesn't appear as an option.  Also, the fitacf.revision major and minor values are listed as 3.0 in the dmapdump output of fitacf files produced using versions 2.5 and 3.0.  Haven't tried plotting the data yet to see whether the fitacf-version flag is selecting the correct fitacf library.

egthomas:
Guess I should have fully read your original comment that make_fit defaults to 3.0 on this branch.  The `-fitacf-version` flag does appear to be working correctly (despite not showing up in the command line help).

One minor suggestion - go ahead and drop the `-new` flag and swap it for `-old`.  No radar has produced a dat file in over a decade and rawacf files now vastly outnumber the old dat files.  Let's finally move past solar cycle 23!

kkotyk:
I'll look into that documentation stuff cause I sear I did add that. Maybe I didnt commit it. 

I've discussed this -new flag stuff before with Pasha and it seemed like he was hesitant given how people will have to change their scripts and they are used it now. I personally agree though.

pasha-ponomarenko:
Well, my suggestion was that we just drop the new/old keyword and determine the file type by its extension. The new/old keyword still can be present as optional.

egthomas:
The trouble with that solution is you assume users only apply certain extensions to their filenames (ie dat, rawacf, fit, fitacf, grd, grdmap, map, cnvmap).  It can be useful to put a custom extension on the end of a data file if you're comparing the output from two different processing techniques.

kkotyk:
So Evan, I've improved the XML docs a bit, but the documentation was working in the previous commit for me. Did you run a fresh install before checking? Maybe you needed to run make.doc again? Regardless, I relooked at it and cleaned it up a bit more. This is what I get running a fresh install:
root@07772e41cd83:~/rst# make_fit --help
make_fit - Creates a fit or fitacf format file from a raw (dat) or radwacf format file.
make_fit --help
make_fit -fitacf-version [version number]
make_fit rawname fitname [inxname]
make_fit -fitacf-version [version number]
make_fit -new [rawacfname]
--help           print the help message and exit.
-fitacf-version  Select fitacf version.
--h              print the help message and exit.
rawname          filename of the raw (dat) format file.
fitname          filename of the fit format file to create.
[inxname]        filename of the associated index file to create.
[version number] Valid choices are 3.0 or 2.5.
-new             input file is in  rawacf file format and the output should be in fitacf file 
format.
rawacfname       filename of the rawacf format file. If this is omitted the file is read from 
                 standard input.

kkotyk:
> It can be useful to put a custom extension on the end of a data file if you're comparing the output from two different processing techniques.
Can you maybe give an example of this @egthomas? Thanks.

pasha-ponomarenko:
This is more like an exception, and for that exact situation we can leave optional -new and -old flags.

egthomas:
I know the VT crew adds "ex" to the end of their grd and map file names to indicate they were generated from FITEX (2) format files rather than FITACF.  And that would also be doubly inconsistent, since grd and map are supposed to refer to old file formats (ie pre-dmap) while grdex and mapex files are actually in the newer grdmap and cnvmap formats respectively.

ksterne:
@egthomas, that's "fake news" you're spreading there.  We're generating grid data from fitacf files these days.  At one time we may have used fitex files, but I found that fitex broke on some files and/or fitacf is the standard everyone else is using making for easier comparing.

+1 here for *not* using filename extensions to signify data processing.  Simple "mv 20170403.2200.02.fitacf.bks.bz2 20170403.2200.02.fitacf3.bks.bz2" and now I've got a fitacf3 file?  So easy to confuse.  There's more I've got here, but don't have the time for now.

sshepherd:
I agree with the consensus, file extensions are not entirely consistent and
the command-line flags should be modified to reflect the current files that
we are all using and distributing. The default operation should be on the
"new"-style files (.rawacf, .fitacf, etc.) and one should have to set a
command-line flag to indicate the "old'-style files (.dat, .fit, etc.). I
have made these changes in most of the programs that I have touched in
recent commits. The RST is undergoing a fairly major overhaul and I can't
think of a more appropriate time to make the change.

Simon

egthomas:
> @egthomas, that's "fake news" you're spreading there. We're generating grid data from fitacf files these days. At one time we may have used fitex files, but I found that fitex broke on some files and/or fitacf is the standard everyone else is using making for easier comparing.

My mistake, didn't realize the processing scripts had been changed.  My points still stands though that the files are still referred to as "grdex" and "mapex" files.

pasha-ponomarenko:
@sshepherd, what consensus? 
With you?
;-)

It looks like we are miximg several thing here:
(1) abandoning/modifying -new switch
(2) using raw file extension in order to identify the old and new raw data formats
(3) reflecting in the filename/extension the type of post-processing package/algorithm 

1. My only concern is that the removal of -new will inevitably lead to crashing all data processing codes within the community. If we want to make this modification, a clear warning should be sent well in advance and, probably, repeated two or three times.

2. I still think that this is a valid solution because the deviations from .dat and .rawacf extensions are not conventional and usually they are dealt with individually/manually. Optional -old/-new switches should help to deal with these individual cases. I have a feeling that this is also applicable to map/grid formats. 

3. I agree with @ksterne on this one: it might be useful at the testing stage but should not go beyond it. Also, we are not distributing fit/fitacf data any more (although we might start distributing them again...).

sshepherd:
Consensus: Keith, Kevin, Simon, Evan.

I think you are proposing something more drastic by having all the
processing software correctly identify the file type by it's extension. We
have been using .rawacf files since 2006 and require the use of a -new
command line option. Even the comments in the source code indicate that
this should change when these files are adopted as the standard. If another
format comes along we will need to add -newer or -newest command line
options. I think your scenario of all the data processing codes in the
community crashing are a bit dire. We have made these changes since we are
running what is effectively going to be RST 4.1. "Crashing" is actually a
good thing since you know that things are not correct and forces you to
make sure you are doing the correct thing.

On Fri, Jun 23, 2017 at 1:51 PM, pasha-ponomarenko <notifications@github.com

kkotyk:
@pasha-ponomarenko That's not exactly true on point number 3. Can't remember which funding agency specified this, but we are supposed to be freely distributing fitacf data for the Canadian radars. I had this set up on our old website, but the developer is taking a long time to get it working on the new site.

kkotyk:
What is everyone's thoughts if we don't use flags like -old and -new, but instead something like "--input-type rawacf" or "--input-type dat", with rawacf being default if this is not supplied. This way we could add further input types if need be in the future.

pasha-ponomarenko:
@kkotyk, the extension-based keys sound like a good solution!
With respect to not distributing fit data, I meant data distribution within the SuperDARN community, although it might cnage, according to the discussion at SD'17.

@sshepherd, I am not against the change, and I wish you read carefully what the others write before typing your next message, 
:-)
otherwise it might look like you are somehow obsessed with removing the flag right now and at any cost!
:-)
> "Crashing"  is actually a good thing
Not to those who try to process the data.  I imagine your reaction if this happens to you!
;-)
So, let's find a proper (human!) solution to this issue.
I hope you notice the smileys above...

pasha-ponomarenko:
We have just had a brief discussion with @kkotyk, and from our point of view an optimum solution  **with respect to make_fit** seems to contain following points:
(1) automatic identification of the new or old raw data format from the file's extension (Note: this is for the INPUT DATA ONLY!)
(2) the output format by definition is .fitacf for both .rawacf and .dat files with a possible switch for those interested in antique things
(3) flexible piping of the fitted output including that to null directory.

And I reiterate again a necessity of being gentle with respect to the users while introducing these (or any other) changes that affect the format of the commonly used routines.

sshepherd:
I will continue to disagree on this point. I believe implementing auto
identification (and making sure if works 100% of the time) is a bigger task
than simply switching a new flag for an old one. In terms of the code it is
a _very_ simple change. Scripts can (and should) be modified.

On Fri, Jun 23, 2017 at 3:59 PM, pasha-ponomarenko <notifications@github.com

pasha-ponomarenko:
Well, your solution looks like just another patch-up fix identical to that of Rob's while ours is a bit more robust. Furthermore, it is Keith who would be modifying the make_fit thing, and he does not seem to be as concerned as you about implementing these changes. 
@kkotyk, please correct me if I am wrong!

egthomas:
It's more than just make_fit that would require this implementation.  We would need to make sure every single binary in the RST has consistent input handling, otherwise make_fit has some automatic detection but make_grid and map_fit and all the plotting code still require `-new` flags.

kkotyk:
Well I am not particular to either suggestion, however based off my experience with make_fit, I think it will be about the same amount of work either way, provided that input files are always of ".rawacf" or ".dat" extension. I don't believe there is much variation from that extension for our mirrored files.

I do think that the current system of piping the output to any sort of filename is useful. That should go unchanged.

The bigger question I have is what type of output make_fit should produce. Currently make_fit will produce a fitacf file from a rawacf file and a fit file from a dat file. I'm wondering if its a good idea to investigate just always producing the fitacf format as default no matter the input type.

sshepherd:
It is more than make_fit. Gridding, mapping, trimming, plotting must all be
changed. Make sure you deconflict with all the changes that have already
been made to these functions. They have been committed to github and are
still sitting in the queue.

On Fri, Jun 23, 2017 at 4:20 PM, pasha-ponomarenko <notifications@github.com

kkotyk:
@egthomas Okay well thats a different question then of whether or not every binary needs the same method of input arguments. If thats the case, then we need to use something like I suggested with --input-type. However, either way we are going to have to investigate every binary if they need to be consistent.

kkotyk:
I also feel like this discussion has ballooned out of the scope of this pull request and should be moved....

pasha-ponomarenko:
> It is more than make_fit.

Then open a new discussion topic please or refresh the respective old one.

sshepherd:
https://github.com/SuperDARN/rst/issues/7

On Fri, Jun 23, 2017 at 4:41 PM, pasha-ponomarenko <notifications@github.com

sshepherd:
This has already been done on the gridding, plotting and mapping software.
Evan and I have been using it since before the workshop.

sshepherd:
Keith, a couple of questions about your implementation of make_fit:

First, I see that you have introduced a branching statement inside the
do-while loop. Will that not inherently slow down processing? We are
currently processing (in parallel) several years of data to be sure that we
are using fitACF2.5. I would hate to see anything that slows down the
process. Normally I would suggest setting function pointers, but the
refactored version now calls two functions instead of one, so that may not
be possible.

Second, you are using == with floating point numbers. Everything I have
read suggests that this is just asking for trouble. Can you either #define
constants for the different versions and set them to integers?

Simon

kkotyk:
I'm not sure precision in this case is an issue, but I can look at it.

add_fitacf.3.0 branch: 
time make_fit -fitacf-version 2.5 -new 20070101.2201.00.sas.rawacf > /dev/null 

real	0m0.769s
user	0m0.760s
sys	0m0.008s

develop branch: time make_fit -new 20070101.2201.00.sas.rawacf > /dev/null 
real	0m0.774s
user	0m0.746s
sys	0m0.028s

Seems like timing shouldn't be an issue. The application spends very little time in make_fit comparative to the call to fitacf. That loop you are talking about is also already loaded with branching statements, so I'm not sure it is adding much there.

sshepherd:
Not worried about precision, but rounding errors and weird complier issues
can cause == to be close but not actually equal.

Your test implies it is faster? Perhaps we should run it on a few years of
data. Is fitACF3.0 ready for that? I can run it for the six-year period
Evan is using for his statistical study.

Simon

egthomas:
After we finish sorting through the rawacf files with data quality issues we should be able to try mass-processing 2010-2016 data from the Northern Hemisphere using the fitacf3 library.  That would probably not be complete for another 2-3 weeks.

Has anyone else tried using fitacf3 yet?

pasha-ponomarenko:
> Has anyone else tried using fitacf3 yet?

Yep, I processed several days of normal mode from different years over the whole network.

Visually, a very significant gain in coverage for both ionospheric and ground scatter is generally observed in the polar cap and mid-latitude radars. The auroral radars  does not show much of it but the patches of ionospheric scatter look more cosistent and continuos. 

FITACF2.5
image

FITACF3.0
image

There are some problems that might require a closer look.

(1) At some locations there is an issue with letting too much noise in. It seems that this happens when the background noise/interference distribution significantly deviates from Gaussian. This started to happen, for example, in HOK data after the second radar has been put into operation, so the mutual interference might be the cause. 

HOK in 2010
image

HOK in 2016
image

(2) The "streaks" @egthomas discovered some time ago are also more evident in fitacf3-processed data from some sites. They might be raleted to strong lightning strikes. There is a way to detect them but it needs further testing. 

(3) Quite a few radas show horizontal gaps in data usually related to an inaccurate estimate of a distance to the first range gate. We might need to introduce the range offset correction paramenter (rdiff?) to the hardware informastion.

pasha-ponomarenko:
Besides the Hokkaido radars, the other stations affected by the excessive noise, although not as severly, seem to be GBR and WAL.

egthomas:
Are the horizontal streaks in your last figure (HOK 2016) due to how the plotting software handles gaps or do the same data values actually stretch over that ~2 hr interval from 10-12 UT?

pasha-ponomarenko:
This is just a data gap.

kkotyk:
I am working on investigating whether or not some small pieces of code are working correctly. I am closing this for now.

sshepherd:
I've lost track of the status of fitacf 3 but it seems that it is not yet
ready to use for mass processing, i.e., in a large statistical study of
many years of data. Pasha showed a few RTI plots of various parameters for
a few radars and pointed to some concerns (I think.) Presumably these are
being worked on. It would also be helpful to also look at histograms of the
various parameters (for a day) using versions 2.5 and 3 so they can be
directly and quantitatively compared.

Simon

pasha-ponomarenko:
> It would also be helpful to also look at histograms of the various parameters (for a day) using versions 2.5 and 3 so they can be directly and quantitatively compared.

By all means, go ahead and do it!
;-)

sshepherd:
Well, I think the onerous is on the developer. We have tested v3 several
times and each time found 'errors'. Based on the last response I am not
sure whether they have been addressed or not.

On Fri, Jul 7, 2017 at 12:42 PM, pasha-ponomarenko <notifications@github.com

pasha-ponomarenko:
>Well, I think the onerous is on the developer.  

Well, you've got the idea. Or was that a polite request?

> We have tested v3 several times and each time found 'errors'. Based on the last response I am not sure whether they have been addressed or not.

Your use of plural form suggests that there are some unaddressed 'errors' besides the "streaks". Please remind me what they are.

sshepherd:
It was some time ago, but there was quite a bit of what you called
'salt&pepper'. I suspect that all gets filtered out at some point, but
there was a noticeable increase.

On Fri, Jul 7, 2017 at 12:58 PM, pasha-ponomarenko <notifications@github.com

pasha-ponomarenko:
You must have missed my last presentation in SanQuirico...
;-)

sshepherd:
You were there?!

pasha-ponomarenko:
In body but not in spirit...

pasha-ponomarenko:
https://github.com/SuperDARNCanada/fitacf.3.0/pull/11

kkotyk:
I've cleaned up some bugs. I don't think they affected velocity much, but I think elevation errors may have been off in places. Should be fine to test now.

aburrell:
So I've started testing and some things look odd.  I've downloaded and installed this pull request and there are two things.

First, the option order I had to use differed from what others appear to have used.  And is the only way to get the data read and written correctly:

make_fit -new -fitacf-version 2.5 Raw_Data/20170217.0001.00.han.rawacf

Second, when I try to run version 3.0, I get Segmentation fault 11.  Has anyone else had this problem after the last round of commits?

kkotyk:
@aburrell I tried that same file with either order of command line options. I also tried running optimized code and debug mode. I'm not able to reproduce a segfault. Do you know how to use gdb Angeline? It will tell you the line that segfaults.

aburrell:
I used it 10+ years ago...

Downloaded and tried it now, and the debugger isn't working.

I'll spend some time tomorrow seeing where the code is breaking for me.

egthomas:
I just tried 

```
babylon3:~$ make_fit -fitacf-version 2.5 -new 20170217.0001.00.han.rawacf > 20170217.0001.00.han.fitacf
babylon3:~$ make_fit -new -fitacf-version 2.5 20170217.0001.00.han.rawacf > 20170217.0001.00.han.fitacf
```

and both orders worked.  Next tried version 3.0 and didn't seg fault either:

```
babylon3:~$ make_fit -new -fitacf-version 3.0 20170217.0001.00.han.rawacf > 20170217.0001.00.han.fitacf
babylon3:~$ make_fit -new 20170217.0001.00.han.rawacf > 20170217.0001.00.han.fitacf
babylon3:~$ make_fit -fitacf-version 3.0 -new 20170217.0001.00.han.rawacf > 20170217.0001.00.han.fitacf
```

I am using the `add_fitacf.3.0` branch directly and have not merged it into the `develop` branch as I assume @aburrell and other mac users are needing to do.

aburrell:
I've not merged it into the develop branch, just needed to change one file to make things compile.  I'll try merging it into the develop branch and see if that helps.

aburrell:
The file I'm working with is available at: https://drive.google.com/open?id=0B_eODOjBDFmwM2ZURnl4U29SNkE

egthomas:
Thanks for uploading the file @aburrell.  I am still getting the same results as before (both arrangements of the new/fitacf-version options work and no seg faults with 3.0).  Tested on a Debian 8.7.1 virtual machine and no problems there either.  I'll let you and @kkotyk figure it out from here.

aburrell:
Alright, I found where it's dying.  It's in:

fitacftoplevel.c:

`
    if (radar_prms->time.yr < 1993){
        fit_prms->old=1;
    }
    else
    {
      fit_prms->old=0; /* Dies here */
    }
`

So it appears that there's been a problem allocating memory for fit_prms.  Not sure why this would be :-/  It's processing line: 2017-2-17 0:1:0 beam=15

kkotyk:
Yes again, no problems for me. In your case Angeline, I'm actually inclined to think there is an error allocating the radar_prms based off your message. Every single call to malloc within the allocation of fit_prms is error checked and should produce a message to stderr. Are you seeing any messages in your terminal when you run make_fit? If not I will have to get a mac to see if I can reproduce this issue.

aburrell:
I'm not seeing any error messages in the allocation, but I agree that it's a malloc issue.  Has this been tested for leaks with valgrind?

kkotyk:
Not recent commits, but fitacf.3.0 was tested and had no problems with allocations of fit_prms. I'll double check I have'nt missed any frees in recent commits, but those commits should not be affecting the line you show. 

The other libraries called by make_fit may not produce clean valgrind reports, but I have not modified any of their features.

aburrell:
I think the problem could be because the structure is malloc'ed but not initialized.  This is something I've found works ok on some systems but not others, and often responsible for memory leaks.  If you let me know where FITPRMS is defined (grepped w/o success in likely locations) I can test it out.

kkotyk:
/rst/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/fitacftoplevel.c

There are two functions. Allocate_Fit_Prm and Copy_Fitting_Prms.

aburrell:
Ok.  Still working on this.  For some reason, the initialisation performed in Alloc isn't being passed properly via the return function.  It's not returning a null (added an error check for that, seems like a good idea to do anyhow), but it segfaults if you try to access anything in the structure, even with all of the values being set and having no problem being accessed within Alloc.

I remember running into this problem before, and got around it somehow.  We'll see what tomorrow brings.

kkotyk:
Do the sizes of each call to malloc make sense? Malloc will basically let you allocate the max that size_t will hold, but you won't get an error until it actually tries to access the memory. The operating system does not actually create pages for your memory until its directly requested.

aburrell:
Yes, they make sense and it is capable of accessing the memory within the Alloc routine.  Just not outside of it.

aburrell:
Ok, so I placed the Alloc subroutine in make_fit and it runs without a problem.  It's definitely a problem returning the alloc'ed fit_prms structure.

aburrell:
Alright.  This works if fit_prms is passed as an arguement, and the memory allocated at the make_fit level.

kkotyk:
That seems very strange. I can't see why there is a problem returning a pointer from a function....

aburrell:
It's a functionality that is available on some operating systems and not others.  I gave up on doing that when I was writing code that I wanted to run on both SUN and Red Hat.

kkotyk:
I will need to modify some of this commit you made Angeline. I see what you are trying for, but some of this code is not working the way you think it is.

aburrell:
Glad you spotted it then, alter away.

kkotyk:
I've added some changes. Angeline can you see if they work on mac?

egthomas:
Maybe this is related to #55, but do `fitacftoplevel.h` and `leastsquares.h` need to be in both the `fitacf_v3.0/src` and `fitacf_v3.0/include` directories?  If they're used elsewhere outside the fitacf_v3.0 library shouldn't a single copy in the include directory be sufficient?

kkotyk:
Maybe? Can you remove them from your src dir and see if your thing compiles correctly? Mine was complaining at me, but maybe I just need to start with a clean install.

egthomas:
Just tried removing them from the src directory of a clean install and there were no compilation errors.  make_fit also executed properly on the random rawacf file I selected.

kkotyk:
:+1: I guess that was just my system complaining then!

egthomas:
Well someone else should probably verify that.  Maybe a good use for the docker files?

kkotyk:
Yes, I will be testing a clean install using Docker at some point here

egthomas:
It looks like some of the merge conflict text has been committed in makeall.c, eg

https://github.com/SuperDARN/rst/blob/add_fitacf.3.0/codebase/base/src.bin/build/makeall.1.22/makeall.c#L65-L71
https://github.com/SuperDARN/rst/blob/add_fitacf.3.0/codebase/base/src.bin/build/makeall.1.22/makeall.c#L264-L268

kkotyk:
I did fix this, but I guess the commit needs to be pushed. Pasha is looking at some final testing here, and I'll add that commit soon.

aburrell:
So I've been looking at how fitACF3 performs and it seems that there are a not insignificant number of instances where the velocity is NaN (this isn't the case for spectral width or power).  Is this a feature or a bug?

kkotyk:
I am not sure. I would need to see what file you are using.

aburrell:
If you'd like to look at a small sample, there are several NaN for han on 5 June 2001, cp=151, frequency ~ 9.942 MHz.  They don't seem to favour any particular beam or range gate.  The spectral width and elevation are zero for these points, power looks normal.

ecbland:
I have a question about the new method for calculating the lag zero noise level, which is done by the routine `ACF_cutoff_pwr` in `preprocessing.c`:

My understanding is that the time lag between the first two pulses in the multi-pulse sequence sets the maximum range for which the lag zero power can be determined. So, shouldn't we exclude any range gates beyond this maximum before sorting the lag0 power list with `qsort`? If the lag zero powers in these range gates are systematically underestimated, then this will affect the calculated noise level. 

Side note: the new noise calculation has significantly improved the quality of the fitacf-level data for the Svalbard radar. All of that low-power noise has vanished! Your efforts are highly appreciated :)

pasha-ponomarenko:
@ecbland, thanks a lot for looking closely at this routine! With respect to the noise level evaluation at these ranges, it is actually the other way around: lag 0 power here becomes a subject to cross-range interference so it is more likely to be overestimated. 
The side note is appreciated too!
;-)

pasha-ponomarenko:
@ecbland, thanks a lot for looking closely at this routine! With respect to the noise level evaluation at these ranges, it is actually the other way around: lag 0 power here becomes a subject to cross-range interference so it is more likely to be overestimated. 
Your compliments are also appreciated!

ecbland:
@pasha-ponomarenko - My comment was based on a casual observation that some radars (see below) often have very low lag 0 power at the furthest range gates. I had assumed that this was related to the maximum range at which we can determine the lag0 power (let's call it `Rmax`), but I could be wrong. 

I've taken a closer look at this effect by modifying my copy of `preprocessing.c` to write the actual range gates used in the noise calculation to a text file. I did this for an arbitrary day (1 April 2017, 24hrs) for 20 radars. The attached histogram plots show how often each range gate is used in the noise calculation. The red vertical lines indicate the range gate corresponding to `Rmax`.

These plots indicate that, for most radars, it is the furthest range gates which are most likely to be used for the sky noise calculation. This is no surprise. However, there are some strange occurrence spikes at far ranges for the DCE, HAN, HKW, HOK, INV, LYR and SAS radars. It's not clear whether these spikes are related to `Rmax`. There are also spikes at gate 0 for DCE, KSR and WAL. If these the low lag 0 powers are an artifact, then the noise level will be underestimated. 
[skynoise_gates.pdf](https://github.com/SuperDARN/rst/files/1263974/skynoise_gates.pdf)

pasha-ponomarenko:
Thanks, @ecbland! Good job!

From what I can see right now, it looks like the low power at the gates preseeding Rmax might be related to  some unaccounted time offset in data sampling. This is definitely the case with Hokkaido East: the sampling starts at about 1-2 gates later than required so that the blanked lag0 occurs at gate 64 instead of gate 66. It is also a possible cause for Hokkaido West and Hankasalmi, but I cannot say it for sure because the VT ACF plotting tool does not work with stereo data.

The very low power at the first (DCE) or last (SAS, INV, and HOK) gate might be related to the blanking pulse being a bit wider than anticipated, but these cases need a bit more investigation.

In general, although we need to keep all this in mind, I don't see a major problem here because (a) it is only one or two values out of ten used for the noise estimate and (b) this is happening only at selected radars.

pasha-ponomarenko:
Back to Angeline's inquiry on NaN velocity values. 
A closer look at the offending records (HAN from 2001) revealed that at that time the raw data were pre-filtered accrording to some unspecified criteria so that lots of ACFs have been replaced by zeroes leading to some interesting effects while being processed by FITACF3. Apparently, this was done in order to save some disk space. Furthermore, there are some ACFs with constant low-level Re and zero Im parts also causing some problems. All these "features" seem to be absent from the data after ~2006-2007 i.e. since when all ACFs are being recorded. 
Keith is working on the optimal ways to filter these cases out.

ecbland:
@pasha-ponomarenko:
> In general, although we need to keep all this in mind, I don't see a major problem here because (a) it is only one or two values out of ten used for the noise estimate and (b) this is happening only at selected radars.

I agree in principle. However, I think it's worth checking (a) more closely. I have noticed that, in some cases, the lag-0 power is significantly lower at the furthest range gates, which could indeed have an effect when one takes the average of **only** 10 numbers. 

The fact that this behaviour occurs only at selected radar sites indicates that it may be a symptom of a different problem (such as a time offset in data sampling, as you suggested). I'll have a look into it and report back next week.

pasha-ponomarenko:
@ecbland, when you say

>  in some cases, the lag-0 power is significantly lower at the furthest range gates

is it appening on scan-to-scan basis or statistically?

pasha-ponomarenko:
> appening

Just came from Montreal!
;-)

pasha-ponomarenko:
One thing we discovered while looking into the NaN issue raised by @aburrell is that the built-in RST routines for reading dmap convert NaN's and, possibly, inf's into zeroes somewhow. I used the IDL library but I believe that it uses C-based DLM-s. It transpared from comparing the "coventional" IDL output with that produced by Keith's Python code. This finding probably deserves to be treated as a separate issue.

egthomas:
Should `make_fit` even be writing NaN's to files?  I've encountered several cases where NaN values for power or velocity in a fitacf file can propagate upward into higher level data products and cause further issues.

pasha-ponomarenko:
> Should make_fit even be writing NaN's to files? 

That's what we are looking at with Keith. NaN velocities are produced by FITACF3 under rather specific but not uncommon (at least in the *.dat file referred to above by @aburrell) circumstances when an ACF possesses a constant-level non-zero Re while all Im values are zero. Keith

FITACF3 also generates inf's in spectral width.

FITACF2.5 generates some inf values in errors and error-related quantities and NaN's in elv_high and elv_low. 

While Keith is working on eliminating NaN's in FITACF3 velocities, the other three instances (inf's in FITACF2 and 3 and NaN's in FITACF2) are stilll to be investigated.

ecbland:
@pasha-ponomarenko:
> is it appening on scan-to-scan basis or statistically?

It is 'appening on a scan-to-scan basis at HAN, HKW, HOK & LYR. I haven't checked every single radar, so other radars may also be affected. In the figure below, I've plotted the lag-0 power in each range gate for the first 1000 scans on 1 April 2017. I've defined 'scan' to mean a single beam, not the full field of view, and all beams are included in the plot. These are the actual lag-0 powers used to calculate the noise level, not the values eventually written into the fitacf files as `pwr0`.

[skynoise_scan.pdf](https://github.com/SuperDARN/rst/files/1274384/skynoise_scan.pdf)

For the above radars, notice how the lag-0 power in a few of the furthest range gates is about 1 order of magnitude lower than the surrounding range gates. For HOK it is several orders of magnitude. If we include these outlying lag-0 powers in the sky noise calculation, the sky noise would be underestimated by a factor comparable to the sky noise correction factor you introduced.

pasha-ponomarenko:
@ecbland, I have had a closer look at the data and came to the conclusion that the low-power lag0 problems near Rmax seems to be confined to the stereo radars. The stereo control program seems to fail replacing the Tx-overlapped values with those calcualted from the alternate pulse. There are either two (HAN) or three (LYR) affected gates which are off-set by one gate between the channels, as expected from the stereo blanking procedure. The only one affected gate at HOK is also expected under this scenario because it is located next to the HKW stereo radar and blanks pulses from the stereo mode but applies the mono algorithms in delaing with Tx-overlap.

As you correctly pointed out, the overall effect of the near-zero values on the noise level estimates, +(1-1.5) dB,  can negate the correction implemented in FITACF3, -(1.5-2) dB. Even one "missing" range gate leads to ~0.5 dB underestimate of the noise level, so it seems to be reasonable to restric the background noise calcuaiton to data between gates 1 (e.g. low power at DCE) and 64 (one below Rmax).

kkotyk:
I added some changes to handle INF and NAN values. Seems to work fine for me now. If this doesn't compile for you, there might be a header conflict between branches or something. I had this issue and it worked after I deleted the "fit_structures.h" header that was leftover in the src folder from a previous branch/commit.

pasha-ponomarenko:
@aburrell, you were the first one who spotted the NaN instances in data fitted by FITACF3:

> So I've been looking at how fitACF3 performs and it seems that there are a not insignificant number of instances where the velocity is NaN (this isn't the case for spectral width or power). Is this a feature or a bug? 

Could you please test the last changes made by @kkotyk which are supposed to resolve this issue?

kkotyk:
It seemed to work for me in my testing, but I didn't plot the data.

pasha-ponomarenko:
@aburrell, no pressure!

aburrell:
Alright, I gave this a go and got segfault 11 again.  I imagine it's the same as last time and there are new instances of uninitialized but allocated memory.  @kkotyk , is this something you can check?

kkotyk:
I will investigate.

kkotyk:
Hey @aburrell when you get a chance can you run "valgrind --leak-check=full --logfile="valgrind.out" -v make_fit _args_" and send me the valgrind.out file? That might be able to give me a clue to when you are segfaulting since I am again unable to reproduce this. You will have to compile RST with -g instead of -O3.

aburrell:
Sure thing: 


[valgrind.log](https://github.com/SuperDARN/rst/files/1353999/valgrind.log)

kkotyk:
@aburrell I think I fixed it. Should be good to go whenever you want to try.

sshepherd:
I acknowledge that I might be mistaken, but I believe the error here was due to an assumed NULL-terminated link-list. I will repeat my earlier concern about the dramatic change in the underlying structure used for this new version (from arrays to a linked-list structure). A linked-list data structure is definitely an advantage in some situations, however, I am not sure that this is one of them. In addition, linked-lists are more susceptible to memory issues and most importantly, a linked-list is not something that most programmers will easily understand and be able to contribute to in the future. I have informally asked several programmers and most have never heard of a linked-list, but are comfortable with arrays.

I would suggest either going back to an array structure or abstracting the structure away from the user by providing set and get functions for the various elements. The alternative is that we have software that will be maintainable by fewer and fewer people. I don't think that should be the goal of this working group.

aburrell:
Ok, so now there's a new problem!

There are no longer NaNs.  However, there is a LOT of missing data.  This data is seen at all frequencies, program modes, et cetera and vastly outnumbers the previous NaN occurrences.

Old V3:
han_bm7_cp151

New V3:
han_bm7_cp151

kkotyk:
This is interesting. We can investigate.

kkotyk:
Another thing to note is not only NaNs or INFs, but values of exactly 0.0. Its a symptom of the same behaviour.

aburrell:
Yeah, I hate values of exactly 0.0 in variables like velocity the most.  I’d much rather have NaNs!

I’ll check out the 0.0 occurrences the next time I test.

kkotyk:
Well basically my last change involved adding 2 sets of filtering based on some of the data you showed had problems before. The filtering I added was based on fits of exactly 0.0 or something else nonsensical. I had to filter ranges in which all power values were constant. When I saw this, you either saw a fit of 0.0, or something else stupid since those constant power values still had different weightings for fit. The other thing I saw was if you had almost all constant power values, with one or two outliers that had much larger weightings than the other lags. This produced something similar with the fitted results. Did you happen to plot the raw data as well? Does anything look strange with the regions that are removed?

aburrell:
I’ve not plotted the raw data, and don’t know how I would do that.  Removing anything with exact fits of 0.0 could be problematic, since it could very well be a real value in some cases.  However, removing fits with a constant power across multiple ranges seems like a good idea (I don’t know enough about ACF to speak much on this area).  Clearly a lot of the data that was removed was just fine.  If you grab one of the times where data is missing in the plot I showed, can you go in and see why it was rejected?

kkotyk:
When I was talking to Pasha about the data, neither of us could really come up with a legit scenario in which a fit of 0.0 made physical sense, but maybe you have another idea of where you might see that? I'm just working on a report right now, but I can look into those values are rejected using GDB when Im done.

aburrell:
My immediate thought is groundscatter, which should look like identically zero at least a fraction of the time (to the sig fig we save).  On the other values, I agree with you.  But I think a closer look will yield more than my speculation, so let me know when you have time to take a peak.

pasha-ponomarenko:
Wrong button!
SORRY!

kkotyk:
hey @aburrell neither Pasha nor I get those gaps when we process the data. On your end, is there a noticeable different in output file size between the two versions? Maybe there is something else going wrong in the processing chain. I'm also going to dig out my mac computer to see if its only on mac.  If it still looks a problem for you, can you again send me the valgrind report?

aburrell:
Hi @kkotyk, there are differences in output file size between the two versions.  The current version does not always end up creating a smaller file, but that is usually the case.

I ran a few examples.  In every one there was a segfault that occurred at some point.  I suppose I didn't notice that before because I ran them from a script whilst I was at home in the evening and forgot to create a logfile :(  

Here are valgrind logs for three of the files.  One segfaults immediately; the others run for short and long periods of time before a problem is encountered.

Files: 20010609.0000.00.han.rawacf, 20010609.1600.00.han.rawacf, 20010609.1400.00.han.rawacf

Valgrind logs: 

[valgrind_00ut.log](https://github.com/SuperDARN/rst/files/1408703/valgrind_00ut.log)
[valgrind_14ut.log](https://github.com/SuperDARN/rst/files/1408704/valgrind_14ut.log)
[valgrind_16ut.log](https://github.com/SuperDARN/rst/files/1408705/valgrind_16ut.log)

kkotyk:
Alright thanks. This will be helpful.

kkotyk:
@aburrell Are you sure your code is up to date with commit 93e31a8aa309ce8f7f5a90dbfbbf40591ebb0cac? The lines from the valgrind report match up to the problem lines in 0b92a6705e28a4663a81ce3352b86aa947103d92 and one of those is commented out now.

aburrell:
Yes and no.  I updated RST before running this, but that doesn't mean something didn't go wrong.  It does give completely different results from what I had before updating it, so something changed for the better. I'll give it another go and see what happens.

kkotyk:
Okay, because I see reference to a problem at fitacftoplevel:313 in the valgrind report, but that line is commented out now. I've definitely had my share of weirdness with this version control stuff.

aburrell:
I'm running it now and it looks different, so that's good(ish).  There are a load of stderr warnings, so I'm capturing them for you as well.  Should be done in 10-15 min.

kkotyk:
Perfect! Thanks!

aburrell:
So the stderr capture didn't work.  Here's what came out when I ran it without trying to capture it, though (15->14):

[stderr_15ut.log](https://github.com/SuperDARN/rst/files/1408899/stderr_15ut.log)

Here are the new valgrind logs:

[valgrind_00ut_new.log](https://github.com/SuperDARN/rst/files/1408904/valgrind_00ut_new.log)
[valgrind_14ut_new.log](https://github.com/SuperDARN/rst/files/1408905/valgrind_14ut_new.log)
[valgrind_16ut_new.log](https://github.com/SuperDARN/rst/files/1408906/valgrind_16ut_new.log)

Only 14UT is a different size than the old ACF 3.0 file; the other two are the same size.  The binary files differ, but that's to be expected.

kkotyk:
Do you still have big gaps in the data? Or can that file size be attributed to some nans, infs, exact 0 being removed? So I'm not seeing anything in the logs now to indicate any memory issues with fitacf.3.0, but something in your system is freaking out in a way I've never seen before. I'm not even sure where to start looking for the issue with those stderr warnings.

aburrell:
It was "freaking out" so much I decided to let you have a peak before running everything.  But if you're not sure where to start I'll run the month and see how it looks.  Should be able to get you plots tomorrow.

kkotyk:
Okay, from quick searches other people have something similar show up on their projects while working on mac. Seems harmless, but annoying. I'll try see if there is a fix.

kkotyk:
It might also go away when compiled without -g flag

aburrell:
I'll recompile w/o the -g flag and let you know.

pasha-ponomarenko:
@ecbland, can you elaborate how you calculated Rmax? I believe it is defined by the time between the first and second pulses in the sequence. The gate number is then defined as a delay between the first and second pulses expressed in mpinc units (9 for the 7-pulse sequence and 14 for katscan) multiplied by the ratio of mpinc/txpl (8 and 5, respecitvely) minus 4 range gates. It gives gates 68 and 66 for the 7- and 8-pulse sequences, respectively.

pasha-ponomarenko:
Is there any difference for the tauscan?

aburrell:
Ok @kkotyk , the warnings go away when you use the -O3 flag.

Superficially, the data looks similar to what it was before:  

han_bm7_cp151

Removing the identically zero data has had a notable difference.  The NaNs made up less than 1% of the results, now the total number of points is about 2%-40% less than what it was (less in the common programming modes, more in some of the special modes).  This is most notable in the groundscatter, and worsened the performance of the backscatter return direction algorithm.

This may not necessarily be a bad thing.  However, I'd be interested to know more about the two filtering processes you implemented.  I'm still not sold on removing potentially viable groundscatter unless there's another reason other than the velocity being zero.

As another note, the new filters still allow a small number of negative power values through.

kkotyk:
Okay thanks for this. I think I will let Pasha investigate this a bit further. I should specify that I'm not filtering velocity. I'm filtering for when the slope of the fitted lines is 0.0. Even if this was real data, its still going to cause the velocities to blow up. INF and NAN and exact 0 shows up when the value of the fitted slope is either 0.0, or some floating point extreme like 10e313 or something crazy like that.

kkotyk:
Also, this is what I diagnosed for the cause of unrealistic velocities and I just tackled it right at the root cause. If anyone else has another idea of how that data should be handled, I'm open to suggestions.

pasha-ponomarenko:
@aburrell, I wonder what you mean when you say "negative power"? Is this the actual lag0 power in A/D units or fitted "power" which is actually SNR?

aburrell:
@pasha-ponomarenko: By power, I am indeed referring to SNR; the "p_l" power, to be specific.  I don't think this is an issue, but I didn't think the NaNs were a problem so I figured I'd mention it.

@kkotyk: It's entirely appropriate to tackle the root cause, but because quite a lot of data is being removed I think it would be enlightening if someone ran some stats on this.  If it's bad data, better to throw it out.  However, if it is so bad, how come it didn't appear to be previously (recalling that more than just NaN/unrealistic values are removed)?  I would like to be convinced, but looking at the month of data I ran it looks like the filtering may be overzealous.

kkotyk:
The other part of that data filtering that I needed to add was if the power at every lag was exactly equal, and I recall lots of that HAN data having that. I'm not sure if there was a problem during that period, buts its almost like the receiver was sampling a constant voltage or something. This behaviour would sometimes cause a fitted slope of 0.0, but not always, based off the weightings at each lag. I think this would actually create some false readings for ground scatter because it seems unlikely to me that every single lag would have the same power value.

kkotyk:
There might actually be even more suspect data in there that should be filtered, but I don't know. I believe there might be cases where like three quarters or more of the power values are equal, or where the power at the lags alternates between two values. More analysis would have to be done on the raw data to really understand that characteristic.

pasha-ponomarenko:
>@pasha-ponomarenko: By power, I am indeed referring to SNR; the "p_l" power, to be specific. I don't think this is an issue, but I didn't think the NaNs were a problem so I figured I'd mention it.

@aburrell, the negative p_l values are fully expected: while the actual filtering for SNR>0 is performed with the measured lag0 power (pwr0), p_l is estimated from fitting a linear function to log power dependence on time lag. For the latter, statistical variablilty of the power at diffrent lags brings some of p_l estimates for low-power ACFs to below 0 dB. As a reslut, pwr0 is always positive while small but noticeable portion of p_l values goes below 0 dB. 

>@kkotyk: It's entirely appropriate to tackle the root cause, but because quite a lot of data is being removed I think it would be enlightening if someone ran some stats on this. If it's bad data, better to throw it out. However, if it is so bad, how come it didn't appear to be previously (recalling that more than just NaN/unrealistic values are removed)? I would like to be convinced, but looking at the month of data I ran it looks like the filtering may be overzealous.

I would like to look at this more closely. It might help to narrow down the problem if you point to a particular patch in your data that, in your opinion, seems to be overfiltered.

aburrell:
HAN June 2001 with cp = -5253 has a noticeably large difference.

pasha-ponomarenko:
@aburrell, I have had a look at the data you mentioned above, specifically at 18 June 2001 records. It seems that the overhelming majority of the rejected data are restricted to the higher frequency ~13 MHz (08-18UT), while at ~10 MHz (00-08 and 18-24UT) there are virtually no data rejected based on Keith's criteria.

All data:
image

Rejected data
image

Most of the rejected data correspond to very low power levels. Here I plot a set of histograms for p_l over 08-18UT with black for unfiltered data, blue for filtered data and red for those rejected.
image

Furthermore, I looked at the raw data and found that the lag0 power of in these ACF is mostly equal to only three A/D discretisation units (for some reason the discretisation step in these records is 32).
image
On top of it, ACF phase at non-zero lags in the rejected data data consists mainly of zeroes (~93%).
 
The interferometer phase values in these records are concentrated at the integer multiples of 45 deg (plus and minus 0, 45, 90, 135, 180) which also signifies very rough discretisation (i.e. very low accuracy) of the data. 

So, I think there is no danger that the meaningful quality data are rejected. In fact, it seems that there is a need for more filtering with this particular dataset. Also, this dataset looks rather specific with some on-site filtering so that at some scans very few non-zero ACF were recorded. If there is less then 10 of them present, which is frequently the case with the 13-MHz data, then the noise claculation might be strongly affected as well... 

Anyways, I was unable to find any evidence of good data being rejected.

pasha-ponomarenko:
I want to clarify that, while I showed beam 7 data in the range-time plots, the following statistics includes all beams.

aburrell:
Thanks for the analysis Pasha, it’s pretty convincing.

I’m afraid I don’t understand all of your discussion about the lag0 power.  What are A/D discretisation units?

pasha-ponomarenko:
>  What are A/D discretisation units?

Oh, sorry, I should have explained it before! In analog-to-digital (A/D) convertors signal digitisation is performed in a discrete way, and the binning step size is what I call a discretisation unit. In your case this number seems to be 32.

pasha-ponomarenko:
This is resolution of the analog-to-digital conversion.

aburrell:
Thanks!

pasha-ponomarenko:
Some comparative stats for INV 2016. Black -- fitacf3.0, red -- fitacf2.5, left -- ionospheric scatter, right -- all scatter. Porcessed twofrequency and normalscan dta only. 

Increase in amount of data relative to fitacf2.5: 
all scatter -- 26%
ionospheric scatter -- 31%
ground scatter --12%

Increase within v=+-100 m/s
all scatter -- 28 %
ionospheric scatter -- 42%
ground scatter -- 12% 
 

image


The velocity distributions (top row) look very similar with the main increase in data amount within +-200 m/s.

The sharper peak in p_l (second row) reflects a more accurate power fitting as well as more accurate noise estimate by FITACF3. Ideally, the p_l distribution should have a vertical cutoff at 0 dB which is used as a threshold for ACF lag0 power.

Spectral width distributions (third row) are pretty much the same

Velocity errors (bottom row) from FITACF2.5 are still too low while FITACF3.0 give more realistic estimates close to what exepcted for the oberved values of spectral width, v_e=w_l/sqrt(nave)/sqrt(mppul-1) (see velocity error tutorial at SD'13).  For the width distribution peak at 75 m/s it gives  75/sqrt(30*7)=5.2 m/s, which is close to the observed peack in the v_e distribution from FITACF3.

I will try to run the same analysis for a couple more radars, one auroral and one at mid-latitudes, early next week.

pasha-ponomarenko:
I managed to get some monthly statistics on most of the radars, and here are three examples for different latitudes (March 2016). Please disregard the reference to twofreq mode for the non-USask radars, it's just a typo...
;-)

INV

image

HAN

image



CVE

image
 
The title on the left top plot shows pescentage of increase/decrease for ionospheric, ground and total scatter with respect to FITACF2.5. As one can see, the monthly statistics for INV looks very similar to that for the whole year, so these examples should be pretty representative. 

In general, almost all stations show a noticeable increase in the number of data points, but at some radars there some oddities that require a closer look...

kkotyk:
I changed the default version of fitacf back to 2.5. I think this PR is good to go for the freeze. I think a separate pull request can be done for changing old/new flag stuff in make_fit since its not related to fitacf.

pasha-ponomarenko:
A bit more insight into the HOK noise problem (see my post from 29 June). As I said at the teleconference, in contrast to most of the other sites, the background noise distribution has a very long tail stretching well beyond the ) dB level. The noise turned on after the HKW radar was put into operation, so I think there is some leakage between the two radars.  

On a positive side, it was found empirically that the wast majority of the "offending" echoes can be filtered out by setting a 50-60 m/s threshold for the Lorentzian spectral width error (for most of the data this value is <20-30 m/s). Below I just copy the original summary plot followed by that with a 50 m/s threshold for w_l_e.

Original 
image

With 50 m/s threshold for w_l_e:
image

Note that filtering also made periodic data gaps at fixed ranges more visible, but I believe that this is related to another problem with HOK, i.e. the  fact that their first range gate has a significant offset ~1 gate size. In this case, the Tx overlap analysis is applied to wrong ranges leaving the affected lags with low power unmarked which in turn leads to incorrect rejection of the respective ACFs. 

In any case, the noise problem seems to be related to hardware issues at a particular site, so at our end we can only mitigate it partially with some extra filtering of HOK data.

ksterne:
So, I guess something that I missed in the telecon a few weeks back, but this introduces new code that won't be the default operation?  (You need to use the -fitacf-version 3.0 option in order to use this code).  That sounds right, but my memory is fuzzy.  I did check it without using this code and it appears as though that part is working correctly.

In flipping from version 2.5 (or 5.0 in the file) to 3.0 (thanks for updating the version number!), I get the following for sto.  Before:

image

after (3.0):

image

I think the "vertical" stripes are something that's been discussed here and seems to be nominal for fitacf3.0?

Otherwise, testing tauscan data, it's looking like the following for gbr.  With fitacf 2.5:

image


with fitacf 3.0:

image

I'm guessing/hoping all that low SNR returns will get filtered out in the griding/mapping process...

Also, looking at some TIGER data from the same day I see with using the 2.5 library:

image

where as, the 3.0 library gives me:

image

which has some fairly nasty looking returns in the 16 to 18 utc timeframe.  Does that look nomial for 3.0?   Is anyone else able to reproduce this or is this just something my setup/codebase?

pasha-ponomarenko:
It loks like you re "flipping" the versions directlyu on the VT server...

pasha-ponomarenko:
I attempted to look at the TIG data dfrom 31 Oct 2016 and realised that there are 664 RAWACF files for that day. All data after 12:30 are stred in 1-min files. I am not sure if this is related to that "nasty" patch, but I was unable to read the concatenated file using RST procedures. I will continue tomorrow -- it's 1am here again...

ksterne:
Sounds good on the explanation. Again, I can't say anything to the science and algorithm here, just showing my results.  Yes, as with #52, #100, I am testing this with our main processing software. So, I *think* most of the radar data from 20161031 was processed with version 3.0 and should be available on our website.

Yea, I did notice the bunch of TIG files that happen, but the short files start a few hours before the messy looking returns happen during 16 to 18 utc.  

Otherwise, are we good to merge this pull request?  I'd think so if this is still something to be tested and refined in the future.  That way we can finally get to a 4.1 release branch creation.

ksterne:
I'll also reiterate my opinion of how changing version numbers on directory structures is not git friendly in this repo and a pain to work with.  I can understand the need to make a separate directory for the fitacf 3.0, but I'd hope there would be a better way we can do this in the future.

pasha-ponomarenko:
Despite the long weekend ;-), I managed to combine the TIG files into a single structure inside IDL, and indeed most of the data between 16 and 18 UT are contaminated by the "streaks", although I have never seen them being THAT bad. Now I am starting to have some doubts about the lightning hypotesis and intend to run some simulations in order to see what sort of external signal (e.g. CW?) would cause this sort of interference.
I must admit that resolving each of these issues (non-Gaussian noise, artificial peaks in velocity distribution due to noisy phase unwrapping, etc) almost  equal to writing a research paper, although the results are not that easy to publish...
Tomorrow I will have a last look at the 31 Oct data at your webpage just to make sure that there are no other problems, and then we should be good to go ahead with merging. 
And I am not qualified to say anything about the version control in git.

mtwalach:
Quick questions with regards to all these artifacts: Are these things that didn't show up with previous versions of the processing software or is it just that no one worried about them in the past?

pasha-ponomarenko:
@mtwalach, very good question!
These things are in the data, they do not result from change in the algorithms. 

The problem is that the original FITACF is overfiltering the data through applying pretty much empiric selection criteria. Furthermore, a custom-made approach to the least-square fitting produced unrealistic errors, sometimes too small, sometimes too large. Finally, a simplified approach to background noise calcualtion produces values which are 1-2 dB lower than the real noise. 
 
In FITACF3 we use the textbook least suqare fitting approach so that all available data exceeding the noise level (>0 dB) were utilised through weighted fitting (the weight is inversely proportional to the error/variance level). We also removed all purely empiric filtering procedures and left only those physically justified like Tx-overlap for lags or 0dB SNR for . While the overall data amount and quality were significantly improved, some extra things inevitably surfaced as the data selection criteria were relaxed. 

The "streaks" seem to be observed by a number of radars so they might have a common origin.  In FITACF2, they were removed by applying some empiric criteria, but there were no reports or notes on this sort of interference by the previous developers, so further investigation into the physical nature of these signals is required.

mtwalach:
Thanks for the explanation @pasha-ponomarenko! Looks like a lot of work to be done.

ksterne:
@pasha-ponomarenko, have you had a last look at data?  Nothing else to add here?

pasha-ponomarenko:
@mtwalach, @ksterne, somehow the last two comments did not make it to my mailbox. 

@ksterne, I am still looking at the rest of the data. It seems that SAN on this particular day lets lots of "streaky" noise in too:
image
There is virtually no data form FITACF2.5 at this time. 
Give me a bit more time to think on where the leak might occur. It might be something really basic...

@mtwalach, yep, lot of work has already been done, and lots and lots have to be done. It is essentially a two-man operation, and both me and Keith have heaps other things to do. So, I do appreciate help from people like you! Also I wish I had a couple of clever grad students to dig through the data.

pasha-ponomarenko:
I am trying to find out what is in common between the "streaks" at different sites (they look rather different at TIG and SAN) and which of the FITACF2 "filters" removes them.

pasha-ponomarenko:
It seems like at SAN the problem occured after a gap in the data from 04UT on 24 Oct 2016 until ~9UT on 27 Oct 2016 so that it might be related to some hardware issues. The noise seems to be still there in Oct 2017...

For TIG, I checked several days before and after 31 Oct 2016 and did not see anything like the band between 16 and 18 UT on the 31st Oct. I will look a bit more...

egthomas:
I don't think we should get too hung up on individual details of the data from a particular day, especially from southern hemisphere sites which may have questionable files or data to begin with.

pasha-ponomarenko:
Well, after looking at a couple of months of data it looks like both these radars have problems. For TIG, it is the lower working frerquency 10.20 MHz which seems to be probe to strong but intermittent interference. For SAN, it is also intermittent but much more frequent and it seems to occur when the backgorund noise level goes down. It seems to limit strongly the amount of data coming from SAN while using FITACF2.5
In any case, these problems seems to be related to the local hardware or interference, so we cannot do much about it anyways.  On a side note, Keith made a good point from the engineering prospective: it seems that FITACF3 might be useful in detecting these issues at early stages as compared to FITACF2 which pushes them under the carpet trough overfiltering.
I think we should go ahead with merging this branch. 

I conclude with a nice example of how much more useful data can be extracted with the new FITACF

FITACF2.5
image

FITACF3.0
image

ksterne:
Merging this in now. I think it's good to reiterate that since the fitacf 3.0 library isn't the default for the make_fit binary, this is a slightly lower risk of merging it in.

PR #62, opened 2017-05-22
installation update

aburrell:
Updated makebin.* files to use XPATH environment variable instead of
hardcoding X11 library directory

Updated README.md to include macOS dependencies, list of all
environment variables that need to be updated, and solutions to
problems encountered during macOS sierra installation.

PR #75, opened 2017-06-17
Fixing compilation warning for fprintf/printf/sprintf calls

egthomas:
This pull request addresses the "...: warning: format not a string literal and no format arguments [-Wformat-security]" error listed for multiple libraries and binaries during compilation via execution of 'make.code superdarn rst'.  As pointed out by Bill Bristow, the fix was simply to insert "%s", e.g. fprintf(fp,"%s",buf). Not only will this reduce the number of compilation errors listed in the logfile but it should also help with installation on MacOS systems.  Note that when I tried compiling on a Mac, make.code failed when I hit a compiler warning about sprintf in the fontdb.1.9 library (see my comments in #62).

### DIFF
--- a/codebase/base/src.bin/build/addtext.1.7/addtext.c
+++ b/codebase/base/src.bin/build/addtext.1.7/addtext.c
@@ -39,7 +39,7 @@ char *rtext={"text"};
 
 void print_info(FILE *fp,char *str[]) {
   int i;
-  for (i=0;str[i] !=NULL;i++) fprintf(fp,str[i]);
+  for (i=0;str[i] !=NULL;i++) fprintf(fp,"%s",str[i]);
 }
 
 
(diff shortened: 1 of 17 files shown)


kkotyk:
This was a good catch. Now that I read into this, it is incredibly sloppy to ignore the format of a string generated at runtime. I wonder if we should go through and catalogue any other warnings generated at compile time....

egthomas:
I've started doing just that on the `warning_cleanup` branch.  A lot of the compiler warnings are about set but unused variables which store the return value of various function calls.  There are also a ton of warnings about changing from one variable type to another in the dmap libraries.  It can get tedious fixing these and a better C programmer than me should probably deal with the trickier ones.

kkotyk:
I'll try look at adding to that branch when I get a chance.

PR #72, opened 2017-06-15
Updating hdw.dat files on master branch

egthomas:
This pull request (hotfix?) updates the Canadian hdw.dat files and adds the missing Longyearbyen hdw.dat file to the master branch.  It looks like a lot of lines were modified because there were a bunch of ^M escape characters from Windows in a few of the hardware files.  After verification from the USask and UNIS folks (@kevinkrieger , @kkotyk , @ecbland , etc) this should be merged as quickly as possible.

PR #78, opened 2017-06-21
Updating bks,kap,sto hdw.dat files on master branch

egthomas:
This commit updates the geographic longitude (sto) and altitude (bks,kap) of three radar sites. These files had been updated on the VT hdw.dat repository but those changes were not propagated to the old VTRST3.5 repository on which the new RST repository was based.  Similar to #72, the kap and sto files look like they were completely rewritten because there were a bunch of ^M characters that were removed when copy and pasting.


egthomas:
I should also note that I checked all of the hdw.dat files on the VT hdw.dat repository and these were the only other ones that were inconsistent with RST 4.0.

egthomas:
Any objections to merging this sooner rather than later?

sshepherd:
No objection

PR #80, opened 2017-06-26
Fixing bug in GridSeek when records are missing (trim_grid and grid_plot)

egthomas:
This commit fixes an issue identified by Maria-Theresia Walach at Lancaster University with `trim_grid` when there are records missing in a grid file.  Currently, `trim_grid` calls GridSeek to identify the starting record based on the command line option inputs. If a record is missing, rather than going to the next available record after the provided start time, GridSeek will point to the last record before the gap, ie earlier than the provided start time.  This problem also occurs in `grid_plot` as it is the only other piece of code which uses GridSeek.

To test, on the `develop` branch execute `make_grid -new -i 120 -tl 120 -c 20140104.*.cve.fitacf > 20140104.cve.grdmap` with the following files:
```
20140104.0401.00.cve.fitacf
20140104.0530.57.cve.fitacf 
20140104.0545.57.cve.fitacf 
20140104.0600.58.cve.fitacf
20140104.0801.00.cve.fitacf
```
There will be a data gap in the resulting grid file between 05:48 and 06:02 UT.  Try plotting with a start time of 06:00 UT to see what happens:
```
grid_plot -x -delay 0 -mag -rotate -fcoast -vecp -vkeyp -vkey color.key -time -raw -grd -grdontop -sd 20140104 -st 06:00 -new 20140104.cve.grdmap
```
Notice how the first figure to appear corresponds to the 05:48 record and not 06:02 - this is because of the bug in GridSeek.  Now try trimming the grid file to a 2-hour grid file from 06:00-08:00 UT with the following command:
```
trim_grid -sd 20140104 -st 06:00 -et 08:00 -new 20140104.cve.grdmap > 2014010406.cve.grdmap
```
Again, the 05:48 record appears in the trimmed grid file.  Now switch to the `trim_bugfix` branch and repeat the plotting and trimming steps - everything should now work as expected.

### DIFF
--- a/codebase/superdarn/src.lib/tk/grid.1.24/src/gridseek.c
+++ b/codebase/superdarn/src.lib/tk/grid.1.24/src/gridseek.c
@@ -123,13 +123,14 @@ int GridSeek(int fid,
       if (tfile>tval) fptr=lseek(fid,0,SEEK_SET);
     } else fptr=lseek(fid,0,SEEK_SET);
     if (atme!=NULL) *atme=tfile;
-    while (tval>=tfile) {
+    while (tval>tfile) {
       tptr=lseek(fid,0,SEEK_CUR);
       ptr=DataMapRead(fid);
       if (ptr==NULL) break;
       tfile=GridGetTime(ptr);
       DataMapFree(ptr);
-      if (tval>=tfile) fptr=tptr;
+      /*if (tval>=tfile) fptr=tptr;*/
+      fptr=tptr;
       if (atme !=NULL) *atme=tfile;
     }
     if (tval>tfile) return -1;


kkotyk:
Hey Evan, any chance you could hazard a guess as to why GridSeek was designed to point to the last record? Potential bug, or was it just not well thought out?

egthomas:
Not entirely sure.  Most of the seek functions appear to have the same logic section for looping through until the correct record is found.  Interestingly though, the IDL implementation of GridSeek disagreed with the C version (and other IDL versions).

kkotyk:
Alright, I was just curious.

egthomas:
So anyone had a chance to test this yet?

kkotyk:
I just asked Maria if she would test since she's been using trim_grid, but if not I will test for you this week.

mtwalach:
I've reprocessed all the 2014 data that I have to see if the records are now in order and trim_grid appears to be working as expected. Thanks for the fix @egthomas !

The only problem (as far as I am aware of) that I have had is that trim_grid can't deal with data gaps in files that span over more than 24h, but I guess it wasn't built for that. I process data in chunks slightly longer than 24h, so I don't end up with map gaps at midnight. So if there is a data gap somewhere in the grid file that I am chopping up, but there is data from the same time from the other day, trim_grid will use that record instead - perhaps something to be aware of in case anyone else is trimming grid files that are >24h! I don't know if this was an issue before or if this was introduced through the bug fix - either way I got around it by using -vb and double checking each date/time of each record it goes through.

As for other parts of rst which use GridSeek (or GridFSeek even), which is what I think @kkotyk was worried about, I don't know if this bug fix breaks them or if it was designed to point to the last record.

egthomas:
Thanks for the feedback @mtwalach!  Does trim_grid still exhibit the odd multi-day behavior if you explicitly set the `-sd`, `-st`, `-ed`, and `-et` options from the command line?

mtwalach:
Ah, I haven't tried using -sd and -ed, just -st and -et, so perhaps that can fix it.

mtwalach:
Just gave it a try: using -sd in conjunction with -st and -et appears to fix this - no need for -ed! Thanks @egthomas.

egthomas:
Awesome!  Glad it worked.

egthomas:
So is this ready to be merged?  Is it still necessary that someone else do so?

Issue #77, opened 2017-06-19
macOS installation


egthomas:
I think we're all aware that one of our biggest outstanding issues is the installation of the RST on systems running macOS.  Prior to the 4.0 release, @aburrell was eventually able to install the RST after a lot of headache and modifications to the build/environment scripts and I was unable to get `make.code superdarn rst` to progress beyond the fontdb.1.9 library.

Bill Bristow sent an email on Friday saying he was able to install the RST on his Mac desktop but had to fix numerous compiler warnings - specifically the ones related to printf/fprintf/sprintf statements (which have all been fixed in #75).  He also said that he needed to add a symlink to the X11 include directory in /usr/include: `ln –s /usr/X11/include/X11 X11`

So for any Mac users out there who want to try compiling the RST, I would suggest trying the `develop` branch which now has the printf warnings taken care of.  You may also want to try Bill's suggestion about adding the X11 symlink.

aburrell:
This is a good idea.  I needed this symbolic link for something previous, so already had it and completely forgot it existed.  Once we get it going we should try and track down a fairly new mac and do an install.

egthomas:
So has anyone else given this a try since the workshop?

agrocott:
I have just successfully* installed to a 2013 i7 Mac Mini running Sierra 10.12.5.  I had a couple of issues with header files not being found (cdf.h and a couple of png/libpng.h files) so I found where they were hiding and copied them into the directories that the compiler seemed to be looking for them - not very elegant, perhaps, but it seems to have done the job.

I also couldn't create the symlink to X11 in "/usr/include" due to Apple's "System Integrity Protection". Apparently you can turn it off, but instead I created the link in "/opt/local/include" which is not protected and seems to be in the compiler path.

*so far things seem to work but I've not done any extensive testing...

egthomas:
Thanks Adrian!  Do you know if you used the `develop` branch (with Bill's fixes) or the master / 4.0 version we sent around over the darn-users list?

agrocott:
I used the develop branch - I got the same print related compiler warnings with the master version.

aburrell:
Right.  I just finished testing this on another mac and doing a web search.  Best I can conclude is that the problem lies with the implementation of bash on iTerm.app.  I tried several things to fix this, without success.  What did work was doing the installation on XQuartz.  I just tested this on the master branch and once I fixed the fprintf statements that implicitly assume a formatting statement there were no issues with installation.

[...]

PR #84, opened 2017-07-11
Updated README

aburrell:
Updated installation instructions for mac, making it possible to avoid
all warnings during installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ For macOS it is also available through macports, as are all listed dependencies
 
    `XPATH, NETCDF_PATH, CDF_PATH`
 
+   If you are running macOS and run into issues with the X11 libraries, you may
+   need to add a symbolic link.
+
    If you have IDL, check to see that `IDL_IPATH` in `~/rst/.profile/idl.bash` is correct.
    (Note: for users without access to IDL, modifying the `IDL_IPATH` environment variable is
    not required).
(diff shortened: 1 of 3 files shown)


aburrell:
Address issue #77 .  Test installation (basically using xterm instead of term) should remove all of the warning issues encountered previously.

egthomas:
I think we need to change the opening line of the install notes from "Upon cloning the repository" to something like "Upon obtaining the software".  A lot of users will be installing from the zip/tar archive and won't know what "cloning" refers to.

egthomas:
Also, are steps 4a and/or 4b necessary with the other updated instructions?

aburrell:
I thought it would be helpful to keep them for people who may not want to use/install an xterm client, since once rst is installed it works fine in the default terminal application.

egthomas:
Piggybacking on this pull request to update the documentation in the README.txt file which describes the make.build / make.code / make.doc scripts.  Also deleted two unnecessary lines from the build.txt file.  I can break these out into a separate pull request if desired but it seemed like this might be an acceptable place to update other readme info.

egthomas:
Just made one more commit generalizing the first installation step and removing the tilde's from the example rst/ paths.  Not sure if I'm allowed to merge this but I think someone should go ahead and do so.

aburrell:
The changes look good.  I think it's better that you merge it than I, given that I started the pull request.

egthomas:
Don't need to ask twice!

PR #89, opened 2017-07-27
Updating aacgmeval and mlteval to use v2 coefficients by default

egthomas:
This pull request updates the `aacgmeval` and `mlteval` binaries to use the new v2 coefficients and procedures by default.  The old AACGM and MLT convert routines may still be used by setting the `-old_aacgm` and `-old_mlt` command line options for the respective eval functions.  Results can be tested against the output of Simon's online AACGM_v2 calculator.

### DIFF
--- a/codebase/analysis/src.bin/aacgm/aacgmeval.1.2/aacgm.c
+++ b/codebase/analysis/src.bin/aacgm/aacgmeval.1.2/aacgm.c
@@ -36,10 +36,55 @@
 #include "errstr.h"
 #include "hlpstr.h"
 #include "aacgm.h"
-
+#include "aacgmlib_v2.h"
 
 
 struct OptionData opt;
+
+/**
+ * Converts an input date from YYYYMMDD format to an epoch time in number of
+ * seconds since 00:00 UT on January 1, 1970.
+ **/
+double strdate(char *text) {
+
+    double tme;
+    int val;
+    int yr,mo,dy;
+
+    /* Calculate day, month, and year from YYYYMMDD format date */
+    val=atoi(text);
+    dy=val % 100;
+    mo=(val / 100) % 100;
+    yr=(val / 10000);
+
+    /* If only 2-digit year provided then assume it was pre-2000 */
+    if (yr<1970) yr+=1900;
+
+    /* Calculate epoch time of input year, month, and day */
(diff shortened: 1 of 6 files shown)


egthomas:
*cough*

ksterne:
_cough_ I was spending time testing and then waiting on things like #52, and then wading through the discussion of #90, #91, #93. But can understand the bump.  /_cough_

Now that I've got that coughed up and out, got around to testing here and see that the results match everything that I can try to do on:

http://sdnet.thayer.dartmouth.edu/aacgm/aacgm_calc.php#AACGM

keeping the default options of Coefficients and Geodetic.  I'm guessing the other options may not be applicable here?  Otherwise if I haven't missed anything with testing this looks good to merge.  So awaiting feedback and for sure that you're done with this.

ksterne:
I also didn't know of a good way to test the -old* options here, maybe you can provide me a way to get the geo to aacgm-v1 calculations?  Or remind me if I've forgotten?

egthomas:
Yeah, field-line tracing is something that wasn't available with the original aacgmeval function so it hasn't been added here.  I think the IDL cnvcoord/mlt functions should give the same result as the `-old` options but I wouldn't be willing to stake my life on it.  The simplest thing may be to switch back and forth between this branch and the develop branch and compare the output after recompiling (although a clean install of each would be preferable).

ksterne:
Hey @egthomas, that's a good suggestion, I hadn't thought of that.  Quickly looking at `aacgmeval` outputs it seems like the `-old` option is working correctly.

I know it's outside of this pull request, but looking at the other options, do you have a quick handle on what the format of the ASCII file should be if using the -f option?  So far it looks like it should be lat lon altitude.  But I tried to put a YYYYMMDD date string after these three and it seems as though the code ignored it as I got a warning about the date needing to be set, so using today's date.  If you don't have a quick answer for that then don't worry, but it's maybe worth noting that we don't know that in an issue.

egthomas:
Dates/times can only be set via the command line options for aacgmeval (since again the original version didn't care about the date).  That did make me realize the `-f filename` help in the documentation for aacgmeval was erroneously copied from mlteval, so the commit from a few seconds ago fixes that.

ksterne:
Sounds good.  I guess we'll leave the format of the ascii file for another time...if and when someone wants to use this.  Merging this now.

PR #93, opened 2017-08-14
Adding new elevation angle algorithm; keeping old with a new command line option

sshepherd:
A new elevation angle algorithm (elevation_v2) has been implemented and replaces the two existing algorithms: elevation() and elev_goose(). The latter was developed for the GBR radar which had a large interferometer offset along the arrays (X) and is _only_ called when the site is GBR, despite other radars having offsets in this direction. For most SuperDARN radars the offset is only in the boresight direction (Y) and the elevation() is exact. An approximation was made for radars with an additional offset in the vertical (Z) direction, however, the solution does not account properly for the 'coning angle', and small errors result. No solution was given for array offsets in all three directions. See doi:10.1002/2017RS006348 for further details.

To use the old algorithms set the -old_elev command line option.

The new algorithm has been tested on numerous radars and gives expected results, i.e., no change for radars with offsets only in Y and small changes for radars with an additional offset in Z (and no negative elevation angles).

One test remains for the phase sign. There are only two radars (GBR and SPS) in the network with the phase sign set to -1, which from the hdw.dat notes implies that the cables were connected incorrectly. It is not clear how this parameter should be treated, however it works correctly for the SPS radar, i.e., gives the same results as the old algorithm. The GBR radar hasn't collected XCF data since the early 2000s, prior to rawacf files, and testing is more difficult.

THE TEST:

Use make_fit to create fitacf files from rawacf files.
Compare the results to existing fitacf files with and without the -old_elev command line option set
The results should be identical for radars with X=Z=0. For other radars the results should be identical only when the -old_elev option is used.


egthomas:
I believe fitacfT.2.4 is only used for the test_fitacf function:

https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/testing/test_fitacf.1.1/test_fitacf.c

pasha-ponomarenko:
Good on you, Simon! 
;-)
Just wonder you decided to implement it inside the 2.5 framework which will be replaced in the near future...

pasha-ponomarenko:
...why you decided...

pasha-ponomarenko:
> One test remains for the phase sign. There are only two radars (GBR and SPS) in the network with the phase sign set to -1, which from the hdw.dat notes implies that the cables were connected incorrectly. It is not clear how this parameter should be treated, however it works correctly for the SPS radar, i.e., gives the same results as the old algorithm. 

Is this prm->phidiff you are talking about?  It looks like you apply it to the arrays' offsets (three varibles, accompanied by a somewhat cryptic reference to Ray) rather than to the phase shift itself (a single variable), as it is done in FITACF2.5. 
Is there a special reason for that?

sshepherd:
I started with fitacf.2.5 for several reasons, not least of which is because it is part of our standard code, which includes all of the AACGM-v2, gridding, etc. code. It probably isn't a good idea to include it with other code that is being tested. This way, if anyone actually tests it, and finds errors, then they know who to blame.

Yes, prm->phidiff, but you are mistaken. It is used to 'rotate' the solution by the arcsine of the amount by which the interferometer is offset in the vertical direction, which isn't correct because it includes any offset in X and doesn't account for the coning angle.

I have asked several 'senior' PIs if they can recall the exact details of this parameter. The only two radars in the network that use it are GBR and SPS.

sshepherd:
After glancing at the fitacf.3.0 branch I definitely made the right call by starting with fitacf.2.5. Instead of a simple function call that occurs in fitacf.2.5, the elevation angle determination seems to be spread across several functions and located in several files, with 'node' initialization, etc. Can we simplify the code in fitacf.3.0? If not, then Keith is likely the only one to who will ever be able to modify it moving forward.

sshepherd:
Is that function used at all, or can it be deleted (and saved in VTRST3.5)?
I think it must come after the fitacf library, and since it has the same .h
files they get overwritten in ~/rst/include/, which is fine, but a bit
silly that you have to know to edit the correct one or your changes don't
appear.

kkotyk:
Hmm? There is a lot of preprocessing done throughout fitacf.3.0, but elevation is done at one place, determinations.c:371. Pasha and I still have yet to write a white paper manual for fitacf.3.0. Fitacf.3.0 may be complex, but its not as complicated when one fully looks at it. I've had plans for some further optimizations, however I feel like we need to get the correctness down first.

egthomas:
> Is that function used at all, or can it be deleted (and saved in VTRST3.5)?
I think it must come after the fitacf library, and since it has the same .h
files they get overwritten in ~/rst/include/, which is fine, but a bit
silly that you have to know to edit the correct one or your changes don't
appear.

I'm not sure if anyone uses the test_fitacf binary, and even if they do, shouldn't it use the fitacf.2.5 or 3.0 library instead of some older variant of 2.4?  For a less drastic solution, could we just delete the include directory from fitacfT.2.4 if its .h files are being overwritten by those from fitacf.2.5 anyways?

sshepherd:
Keith, I did find the location of the elevation angle calculation in
determinations.c and it is similar to the function in fitacf.2.5 with the
exceptions of the elev_fit structure and some extra stuff at the end.

After looking at your refactoring of fitacf I have few questions. First, it
appears that you have implemented a linked-list data structure in place of
arrays. Can you elaborate on the advantages of using this structure over
arrays? My initial reaction is that a linked-list structure is overly
complicated for our purposes and potentially slower without random access
that arrays provide. I don't think we are necessarily memory limited so
making arrays large and checking that max_range is not bigger than the
array size would seem to be the better solution to me.

The second question is that you appear to be using an object-oriented style
of programming where there are numerous functions whose only purpose seem
to be setting an element to a particular value. While this practice seems
elegant and isolates the programmer from having to deal with the details of
the data structure being used, I expect that there is a performance penalty
associated with calling a function to do a very simple operation. Is that
desirable?

I am certainly no expert, but do wonder about these changes from the
previous code structure.

kkotyk:
Hey Simon. When I started this project, I was specifically told to use software engineering principles. Pasha can elaborate, but I was told that correctness and ease of understanding are valued more than performance for this version of fitacf. I find the use of a list apt here because it makes for a easy understanding of the data flow. To me it makes sense to start with a list of items and then successive prune away elements during filtering. When you are done filtering, the programmer knows that any data left to iterate over is good and valid data without having to check or set some sort of bad lags array or whatever. Now, I'm not saying optimizations can't be added. I am a better programmer now than when I started this project two years ago. I've been thinking about implementing an encapsulated array based list structure, but I want to get as close as possible to a fully correct program first.

pasha-ponomarenko:
@sshepherd, you asked for it:

>Yes, prm->phidiff, but you are mistaken. It is used to 'rotate' the solution by the arcsine of the amount by which the interferometer is offset in the vertical direction, which isn't correct because it includes any offset in X and doesn't account for the coning angle.

One more time I recommend you to read others' messages carefully and to avoid having just a qick glance and firing back within seconds. I am not mistaken. The procedure you mentioned above does not deal with the phase sign. This is done in `do_fit.c` through following statements (line 377-378):

>            if (iptr->prm.phidiff != 0)
>xptr[k].phi0 = xptr[k].phi0*iptr->prm.phidiff;

So, I repeat my question: 
>...It looks like you apply it to the arrays' offsets (three varibles, accompanied by a somewhat cryptic reference to Ray) rather than to the phase shift itself (a single variable), as it is done in FITACF2.5.
Is there a special reason for that?

Furthermore, you first wrote:
>After **_glancing_** at the fitacf.3.0 branch ... the elevation angle determination seems to be spread across several functions

and then the next day conceded

>I did find the location of the elevation angle calculation in determinations.c 

Wouldn't it be better to have a close look first and save everyone's time?  

One more thing. It does make sense to coordinate your actions with others, especially if they are directly related to what the others are working on, i.e. the elevation part of FITACF in this particular case. 
Form outside, it just looks like you are itching to implement your pet algorithm ASAP, no matter where and how. 
:-)

I believe that adding a new stuff to the to-be-obsolete FITACF2.5 is less than optimal, and I see no point of investing others' time into testing it while it is supposed to be re-introduced later on in FITACF3.*. With respect to its importance, as you know well yourself, the corrections for the current radars, with exception of ZHS, are minisque, so there is no real urgency in having it working right now and here. 

:-P

sshepherd:
fitacf.2.5/elevation.c line 60: elev_corr= prm->phidiff* asin(
prm->interfer[2]/ antenna_separation);

but you are correct, the phase is also flipped in do_fit.c. I was referring
to the elevation angle algorithm.

I was responding to your questions Pasha. I will ignore the insults, from
the WG chair no less...

I have plenty of pets, but do not include algorithms. I am happy to retract
the request, Pasha. The intent is to put new algorithms in the hands of
others so they can benefit from using them. That has been the intent with
AACGM, MLT, gridding, map potential, etc.. Again, we are happy to not have
to deal with the RST review process, but I expect that others would
appreciate the extra effort.

I think we can argue about "minisque", but again, I will defer to
leadership here and retract the request if that is the desire. ZHS? Perhaps
you mean ZHO?

I don't think a quick test of whether the new algorithm works or not is a
terrible waste of time. In fact I think it is better to test it with
fitacf.2.5 since there is only one thing that is changing.






On Tue, Aug 15, 2017 at 1:13 PM, pasha-ponomarenko <notifications@github.com

sshepherd:
Hi Keith,

Thanks for that perspective. I was under the impression that refactoring
the code meant only that functions were being reorganized, combined,
optimized, etc. and that the underlying structure would remain largely
intact. Changing more fundamental aspects of the code (the data structure)
is more significant than I would have expected and will likely have
implications for the maintainability and development of the code in the
future. Something to keep in mind is that most of the code was developed by
scientists and taking away their ability to modify it is likely not a good
thing. I am not saying this has happened, but something to keep in mind.

Correctness is definitely of importance, but performance is too if you want
people to use it.

In terms of ease of understanding, yes, the old fitacf was very complicated
and refactoring (organization, etc.) is a definite improvement. I am not so
certain that including more complicated data structures has the same effect.

No criticism intended, just offering another perspective.

Simon

egthomas:
Replying to myself here with regards to 79d6ac2 - this commit deletes the duplicate .h files from the fitacfT.2.4 directory, renames the `dofit.h` file (which was modified by AJ) to `dofitT.h`, and deletes the duplicate documentation files/code which were incorrectly copied from elsewhere (probably fitacf.2.5).  This should somewhat help with future confusion between the fitacfT.2.4 and fitacf.2.5 libraries.

(the reason this was done in this pull request rather than a separate one is because Simon had already modified the fitacfT.2.4 dofit.h file since it overwrites the fitacf.2.5 dofit.h file during compilation)

pasha-ponomarenko:
Hey Simon!
Insults? You must have missed the smileys! 
:-)
;-D

> I am happy to retract the request, Pasha. The intent is to put new algorithms in the hands of others so they can benefit from using them. 

Would it be more reasonable in the first place to discuss with the others how to make them happier rather than to force them to be happy your way? I am trying to promote a bit more cooperative, constructive and compassionate model of interaction within the WG with the main focus on common sense. With respect to your algorithm, at this stage I am not advocating any drastic steps in either direction.

egthomas:
Even after the #74 pull request is merged and FITACF3.0 is made the default library users will still be able to call the older FITACF2.5 library with `make_fit`, so I think it is valid to implement changes in version 2.5 if only as a starting point.  Just my two cents.

sshepherd:
An opportunity to discuss this issue (and general RST topics) at URSI was missed.

In terms of cooperation and making others happy... well, that's not really my intent. I am providing an algorithm that is dependent on fitACF but the algorithm only needs phi0 and it is more correct than the existing algorithm. I suppose I could ask if that would make everyone "happier", but I guess I just assumed that it would. "Better cooperation" probably could have been addressed better at  URSI as well but never came up.

The only issue I see with this algorithm is what to do about the phase sign parameter in the hdw.dat files. It is only set to -1 for GBR and SPS. When asked, Bill said it probably shouldn't be set to -1 for SPS. GBR hasn't collected XCF data since 2003, so the effect of this parameter is very small (minisque one might say :)

This code is ready to be merged unless there is further testing to be done.

pasha-ponomarenko:
@sshepherd, I just repeat that it makes no practical sense of testing your algorithm within FITACF2.5 just in order to repeat it again for FITACF3.0 some months later. That's is what you ask the rest of the WG to do for you, and I just don't want to set such a precedent.

sshepherd:
I'm not asking anyone to do anything Pasha. I have developed a more correct
algorithm and implemented it in the RST. Everyone is still using fitACF 2.5
and may continue to do so for a long time.

If there is indeed actually testing, other than "yes" it compiles, runs and
gives numbers. Then testing a new algorithm in an established environment
makes a lot of sense. I completely disagree with your objection.

This is a very simple thing to test, if you feel it needs testing. We use
the development branch here and so use this algorithm as our default.

sshepherd:
Approaching a month with no testing. Is that our limit before an auto-merge happens?

ksterne:
When did anyone ever mention auto-merging would be a thing?

To be truthful, it seems as though there's some issues here which resulted in "insults" to thrown around when this was first brought up.  I'm not stepping near a fire like that. 

I can see @pasha-ponomarenko's point though, why put effort into the old one when a new fitacf is being put forward?

egthomas:
I'll reiterate these points again:

1. FITACF 2.5 is our current default library on both the `develop` and `master` branches (for better or worse)
2. @kkotyk has gone to significant lengths to maintain backwards compatibility with FITACF 2.5 in the pull request adding FITACF 3.0 via command line options to `make_fit`

So the FITACF 2.5 library will not become obsolete once #74 is merged.  What's wrong with creating a new pull request in the future to add the revised elevation angle algorithm to FITACF 3.0 once that has been fully tested and merged?  Otherwise we're mixing separate features in the same pull request.

ksterne:
I guess this is more like a bugfix then?  I guess I hadn't thought of that but maybe it's something we need to duplicate a pull request into `master`.

I can take the point of trying to add on additional features to #74 and making a separate pull request there.  I'm not as familiar here, but once #74 is merged in, will we be back to the same issue that this pull request is fixing?  That back and forth is where it gets tricky.

I'll try to add this to my to-do list after several other time critical things (still doing plot generation after the big eclipse).  Probably I get to it this week is fairly low for me.

sshepherd:
Not really a bugfix, but rather a more general algorithm. The function
elevation_v2() replaces the functions elevation() and elev_goose(), and
gives usable results for ZHO. The old algorithms are still accessible but
require a command-line option.

pasha-ponomarenko:
@ksterne 

> I can see @pasha-ponomarenko's point though, why put effort into the old one when a new fitacf is being put forward?

Another important point is that FITACF2.5 is a reference version for testing further versions, and shifting the reference frame is not a good thing either. 

@egthomas 

>  What's wrong with creating a new pull request in the future to add the revised elevation angle algorithm to FITACF 3.0 once that has been fully tested and merged?

Doing the same testing job twice.

egthomas:
@pasha-ponomarenko There is a `-old_elev` command line option introduced by this pull request which maintains backwards compatibility with the original elevation algorithm, so no reference frame will be lost.

pasha-ponomarenko:
@egthomas, followed by `-old_w_l`, `-old_p_s`, `-old_v_err` etc if one eventually finds better ways to estimate these parameters? Where are we going to stop?

sshepherd:
Ok, now we have gone into the absurd... Pasha, it's pretty obvious that
some sort of line was crossed here, perhaps because someone else did
something in your fitACF territory? I don't know. The elevation angle
calculation independent of fitACF. So test it in the current version. It's
not that hard and won't change when it goes into v3. I was asked by several
people at the workshop to implement the new algorithm. So it did it in the
current RST. That's done. If you want to block it, that's your prerogative
as chair. All further inquiries about the status will be directed to the WG
chair. There are numerous other issues that have been sitting idle. For
some reason this issue elicits an immediate and negative response. If it
helps the WG to focus on other things and maybe move forward, I will
withdraw the algorithm.

egthomas:
So I've attempted creating a new fitacf.2.6 library which would include Simon's elevation angle algorithm while maintaining the functionality of the current 2.5 and 3.0 libraries.  To do so, one would need to adopt a similar naming convention (`fitacf_v2.6`) as we settled on for the 3.0 library (`fitacf_v3.0`).  That was fine for 3.0 since the code had been refactored and used different functions and header files.  However 2.5 and 2.6 rely on functions defined in the same header files, which is a problem since the `do_fit` function needs to be modified to handle the different elevation angle algorithms.  

I don't see a simple way forward to allow versions 2.5, 2.6, and 3.0 to all be called with `make_fit` using command line options without creating tons of mostly duplicate functions with modified names and header files.  I think such an approach would only introduce additional confusion regarding the different fitacf libraries.

I think the primary issue raised for not including the elevation angle algorithm in the current implementation of fitacf.2.5 during the March telecon was to maintain a static reference package.  This is a good idea, however I think there a couple of problems with it in practice.

First, we have kept the same "2.5" library version number since at least VTRST3.5 (and possibly an earlier 3.4 or 3.5 release by APL) despite a number of bug fixes and changes to the calculation of parameters such as the velocity error.  So someone could cite 2.5 as the version of fitacf they've used in a paper without readers being able to tell which era of fitacf.2.5 was actually used (pre-VTRST3.5, VTRST3.5, RST4.0, RST4.1? etc).

Second, and probably equally as important, the `FITACF_MAJOR_REVISION` and `FITACF_MINOR_REVISION` values in `fitacfversion.h` have been set to 5 and 0 respectively since at least RST.2.05 when the fitacf library version number was 1.22 (https://github.com/SuperDARN/rst-archive/blob/rst.2.05/code/src.lib/superdarn/fitacf.1.22/src/fitacfversion.h).  So there's currently no way to determine which version of fitacf was actually used from the contents of a fit file.  

Keith has addressed the second issue for fitacf.3.0 by changing the values to 3 and 0.  I think we need to do the same for fitacf.2.5 so we can differentiate between files produced using the two libraries (and not rely on user-defined file extensions).  And given the difficulties in maintaining multiple versions of the same library due to the compilation/linking procedure of the RST, I think the elevation algorithm should go into fitacf.2.5.  If one really wants to use the previous version of fitacf 2.5 without the elevation algorithm, they can download and compile RST 4.1 (or just use the `-old_elev` flag).

ecbland:
> THE TEST:
Use make_fit to create fitacf files from rawacf files.
Compare the results to existing fitacf files with and without the -old_elev command line option set
The results should be identical for radars with X=Z=0. For other radars the results should be identical only when the -old_elev option is used.

Tested this for SAS, ZHO and GBR and it produces the expected outputs (where `make_fit` was compiled with this branch and `develop`). 

I'm both excited and terrified to click the big green button on this one...anything I should be aware of before I merge it in? @sshepherd @egthomas @pasha-ponomarenko

PR #94, opened 2017-08-25
Adding .profile.tcsh equivalent of .profile.bash

egthomas:
As the title says, this pull request adds a tcsh equivalent of .profile.bash as requested by Rob Shore at BAS.  Since we already have tcsh versions of all of the scripts in the .profile directory it seems logical to include a .profile.tcsh file too.

### DIFF
--- a/.profile.tcsh
+++ b/.profile.tcsh
@@ -1,0 +1,16 @@
+# .profile.tcsh
+# =============
+
+setenv OSTYPE "linux"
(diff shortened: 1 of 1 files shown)


egthomas:
For what it's worth I've successfully tested compilation with this using tcsh on a Debian virtual machine.

sshepherd:
Looks ready to merge.

Issue #82, opened 2017-07-03
Loading custom colour tables


ecbland:
The way in which colour tables are loaded in e.g. `time_plot.c` and `field_plot.c` using `-pkey`, `-vkey` etc. has changed. Previously it was possible to load colour tables from any directory (incl. outside RST) by specifying the full path to the file, but now the code only looks for colour tables in the default directory, `$RSTPATH/tables/base/key/`. 

For example, line 643 in [time_plot.c](rst/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c) reads:
`if (key_path == NULL) key_path = getenv("COLOR_TABLE_PATH");`

Since there is no option to set `key_path`, RST cannot load colour tables unless they are placed in the default COLOR_TABLE_PATH directory. Is this the intended functionality?

egthomas:
I can't speak for @sshepherd since he was the one who made those changes, but my guess is the original intention was to remove the need to type out the full path in front of the color table filename each time one calls a plotting routine.  It sounds like `key_path` should also be added as a command line option (maybe with its default value set to `COLOR_TABLE_PATH`).

As a temporary workaround can you copy your external color tables into the `COLOR_TABLE_PATH` (ie `rst/tables/base/key`?

ecbland:
I agree that this bug can be fixed by adding `key_path` as a command line option - perhaps this was simply an oversight when the code was modified.

Yes, my temporary work-around was to copy my custom colour tables into `$RSTPATH/tables/base/key`.

sshepherd:
I'll try and have a look at this today (yesterday was a holiday of sorts).
Am I correct in thinking that you would prefer that we revert to how things
worked before I made the changes? I recall the changes were a result of not
wanting to specify a full path for a color table in the command line
options.

Simon

egthomas:
Couldn't we have separate options like `key_path` and `key_name`?  That way users who want to use the default RST color tables have less to type but still have the flexibility to use an external color table too?

ecbland:
I agree with Evan :)  Currently the "key_name" is given by the command line options `-vkey`, `-pkey` etc, so we'd just need to add `-key_path` and update the documentation for the affected routines.

sshepherd:
Hi Emma, Evan and others,

Sorry for the delay. I remember when I made this change. It seemed rather
unsavory to have to specify the full path of the color table, even the
default, so the environment variable seemed like a good option. I didn't
[...]

PR #96, opened 2017-08-29
Adding key_path options for custom color tables

egthomas:
This pull request addresses issue #82 raised by @ecbland.  I've added various `key_path` command line options to field_plot, grid_plot, map_plot, and time_plot which should allow the user to specify custom paths for any of the various color table options.  If a color table is specified but no path is provided, the COLOR_TABLE_PATH environment variable will be used as the default.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
@@ -209,8 +209,12 @@
 
 <option><on>-vkey <ar>vkeyname</ar></on><od>load the colorkey used to plot velocity vectors from the file <ar>vkeyname</ar>.</od>
 </option>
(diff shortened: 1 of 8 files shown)


ecbland:
Thanks @egthomas! It works well :) 

I noticed that I need to include the trailing slash `/` when specifying any of the paths (`key_path`, `vkey_path` etc). Perhaps I'm being pedantic, but could we include a statement which adds a trailing slash if necessary when the key path and filename are concatenated? I'm thinking of the novice user who may not figure this out. 

https://stackoverflow.com/questions/18027874/is-there-an-idiom-for-adding-a-trailing-slash-to-a-file-path

egthomas:
That's a great suggestion.  I've added checks for trailing slashes for the paths as well as some additional warning print statements if a particular color table cannot be found.  Hopefully that makes things more user friendly!

ecbland:
Awesome! Works well. The warning messages are a good addition too.

sshepherd:
Looks ready to merge.

sshepherd:
Hey @ecbland, can you merge this? Otherwise it's likely to sit here for another few months... ;>

sshepherd:
Thanks!

PR #98, opened 2017-09-02
Simplifying make.code and make.doc commands

egthomas:
This pull request removes the unnecessary `superdarn rst` arguments from calls to `make.code` and `make.doc`.  These were only being used to label the log files - those filenames have now been hardcoded as `"${LOGPATH}/superdarn-rst.build.${dtval}"`.  Users can also still compile a single binary or library matching a particular pattern by entering something like `make.code bin make_fit` or `make.code lib gtable`.  The relevant portions of the README files have also been updated to reflect these changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+[![DOI](https://zenodo.org/badge/74060190.svg)](https://zenodo.org/badge/latestdoi/74060190)
+
 Radar Software Toolkit
 ========
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ For macOS it is also available through macports, as are all listed dependencies
 	 ```
 
 5. To compile the html documentation, run `make.doc.rfc codebase superdarn` and
-   `make.doc superdarn rst` from the command line. You may need to modify the `URLBASE`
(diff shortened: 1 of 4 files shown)


kkotyk:
I think this PR is a good change to make. If I use make.code bin make_fit, does it recompile all the dependent libraries?

egthomas:
No, unfortunately it's not that clever.  I should also probably note that calling `make.code lib fit` will compile any libraries with names containing the string "fit" and not just the fit.x.xx library.  So it's not perfect but definitely useful when you don't want to recompile the entire RST.

egthomas:
Not sure why a merge conflict appeared - clicking on d0c1f42 doesn't show anything for me.  Oh well.

egthomas:
So anyone have a chance to take a look at this yet?

kkotyk:
Yes, this works for me in Docker.

egthomas:
Thanks!

ksterne:
I know I'm a little late to the game here, but no warning for people using the old method and a deprecation warning?  Probably not a big deal since it's updated in the readme, hopefully we don't get a bunch of questions on it though.

PR #99, opened 2017-09-15
Added option to plot elevation angles in time_plot

ecbland:
This pull request adds the ability to display elevation angles in `time_plot` with a new command line option `-e`. 

New command line options `-emin`, `-emax`, `-ekey` and `-ekey_path` are also provided. These are analogous to the options available for the power, velocity and spectral width parameters. 

I chose to keep the original behaviour of the `-a` ("plot all parameters") option, so only the power, velocity and spectral width will be plotted. All four parameters can be plotted using:
 `time_plot [options] -a -e [fitfile]`

When the elevation angle is not available, nothing is plotted. 

timeplot_elevation_example

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/doc/time_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/doc/time_plot.doc.xml
@@ -116,6 +116,11 @@
 <option><on>-wkey_path <ar>wkey_path</ar></on><od>load the spectral width colorkey from the custom path <ar>wkey_path</ar>.</od>
 </option>
 
+<option><on>-ekey <ar>ekeyname</ar></on><od>load the elevation angle colorkey from the file <ar>ekeyname</ar>.</od>
+</option>
+<option><on>-ekey_path <ar>ekey_path</ar></on><od>load the elevation angle colorkey from the custom path <ar>ekey_path</ar>.</od>
+</option>
+
 <option><on>-fkey <ar>fkeyname</ar></on><od>load the frequency colorkey from the file <ar>fkeyname</ar>.</od>
 </option>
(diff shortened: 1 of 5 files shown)


sshepherd:
Looks extremely useful, Emma!

egthomas:
Tested this and it works as described (for files with and without elevation data).  While testing this, I did notice two things about time_plot which bothered me:

1. all beams are plotted by default - how often is that useful?  should we just set it to something like beam 7 instead?
2. the default velocity scale is +/- 2000 m/s - again, should we change that to something more reasonable like +/- 1000 m/s?

Since these are unrelated to the new elevation angle plotting capabilities I'll leave it up to @ecbland whether she wants to include them here or if they should go in a separate pull request.

pasha-ponomarenko:
Good job, Emma!
I am not sure how difficult it is to modify colour tables in C, but the one for velocity can be confusing or even misleading for two reasons:
(1) high positive velocities are invisible on the white background;
(2) at low velocities, a change from positive to negative values is very sharp, like -1 m/s is red and +1 m/s is dark blue. 
My personal preference is a colour table #4 from IDL which does not include white at its extremes and gradually blends positive and negative velocities through nice shades of khaki, 
;-) 
which generally corresponds to ground scatter so that no extra GS shading is necessary.
image
Again, following Evan's suggestion, the scaling/colour issues might deserve to be considered separately.

egthomas:
@pasha-ponomarenko I might use a different word instead of "khaki" to describe the low velocity color...

pasha-ponomarenko:
It is mostly ground scatter, you know, so something like "fertilised pasture" will do!

ecbland:
Thanks for the input everyone! Let's move the discussion about other improvements to `time_plot` (and other relevant plotting routines) to a separate issue/pull request. I'm happy to take the lead on this.

PR #100, opened 2017-09-20
Feature/map potential_v2

egthomas:
This pull request introduces a series of upgrades to the Map Potential software, from the underlying libraries to the processing and plotting routines.  The list of major changes includes (but is not limited to):

1. Added PSR10 and CS10 statistical models for use with `map_addmodel`. These new models introduce dipole tilt dependencies as well as independent Southern Hemisphere patterns.  The CS10 model replaces RG96 as the default background model (framework has been added for the upcoming TS17 model but those coefficients have not been added yet).
2. Updated labels and annotations for graphical `map_plot` output
3. Added AACGM, MLT, and IGRF -v2 library functionality to all Map Potential routines (old-style AACGM coefficients can be used via the `-old_aacgm` option and dipole field approximation can be used instead of IGRF via the `-noigrf` option with `map_addmodel`)
4. Can now create an empty map file without needing to provide any grid-level data via the `-empty` option with `map_grd` (this can be used to produce Map Potential output for statistical model vectors only)
5. Can now use the Heppner-Maynard boundary suggested by the background model for periods where no data is available by setting the `-nodef` option with `map_addhmb`
6. Can now use separate order expansions of the background model coefficients at the `map_addmodel` stage and model+data at the `map_fit` stage using the `-o` option with each command (eg could perform a 30th order/degree expansion of model coefficients to see more model vectors and a 6th order/degree expansion for the final Map Potential solution)
7. Returned user control of the `error_wt` and `model_wt` options to `map_fit` which had been harcoded inside the CnvMapFitMap function
8. Moved the `calc_bmag` and various HMB functions into libraries to avoid duplication within various other binaries and functions
 
Testing by as many users as possible is strongly encouraged (particularly with Southern Hemisphere data).  As always, any bugs can be blamed on @sshepherd.


egthomas:
Pull request #100 - woot woot!

kkotyk:
Okay wow, this introduces a lot. Any strategies for how to begin testing this? What would be a good progression to start reviewing any main sets of files?

egthomas:
My suggestion would be to pick a day or two and just play around with the options.  Unfortunately I don't have much time at the moment to prepare detailed testing guidelines.  

I did also just notice that the Southern Hemisphere output is incorrect when using only model data; not yet sure if it's isolated to the map_addmodel stage or if the final map_fit stage is affected as well.

egthomas:
Ok, the SH bug in map_fit has been fixed in commit 5111768 by flipping the E-field sign within the `CnvMapVlosMatrix` function in `fitvector.c` according to the hemisphere (similar to the `solve_model` function within `map_addmodel.c`).

ksterne:
Could a test here be to use compiled code from #52 to make grid files and then to make map files with this branch and see the end product since these two pull requests are dependent on each other?

egthomas:
I've merged the `feature/map_potential_v2` branch into the `feature/grid_v2` branch on a separate `test_grid_map` branch (https://github.com/SuperDARN/rst/tree/test_grid_map) for anyone who wants to try them together without dealing with the couple of merge conflicts between the two.

ksterne:
Hey @egthomas, so merging #52 and this PR will create a merge conflict?  I'm guessing there's a plan going forward to resolve the merge conflict (I haven't looked at what it might be, so maybe it's trivial).  Would be better to make a pull request for the `test_grid_map` branch?

egthomas:
The merge conflict is trivial and can be easily dealt with once either branch is merged into `develop`.  I just wanted to save other users some minor inconvenience if they were interested in testing both together as you suggested.

ksterne:
Similar to the discussion in #52, why has the `map_grd`, `map_addhmb`, and `map_addmodel` directory version number remained unchanged here?

Wow, there's no mention of this pull request changing from `-new` to `-old` and not even a deprecation warning here.  I really cannot see merging this in without even a warning of the change since this will break all previously developed code.  If you do use `-new` on the `map_grd` and `map_addhmb` it just says "Grid file not found."  Also, `map_fit` just gives a "File not found." warning.  It would also be good to add in what file was it looking for (Grid file "xyz.grd" not found).  I can add these changes in if you'd like.

Inconsistently, the `map_addimf` and `map_addmodel` binaries don't seem to care as much about having `-new` or not.  Without digging to deeply, I can see that this is maybe this is more of a file format thing.

`map.major.revision` and `map.minor.revision` in the resulting file come out to be 1.13.  Is this to match `map_fit.1.13`?  Any thoughts on upgrading to 2.0?

Also, it seems as though there's still a major problem with at least one southern hemisphere case.  On the current `develop` branch:

image

And on this branch:

image

Shell script used to generate these files is roughly:
```
        map_grd ${switch} ${infile} > ${outfile}
        echo "  Adding HM boundary"
        map_addhmb ${outfile} > ${outfile}.tmp
        mv -f ${outfile}.tmp ${outfile}
        echo "  Adding IMF"
        # map_addimf -new -if ${imffile} -df ${delayfile} ${outfile} > ${outfile}.tmp
        map_addimf -if ${imffile} -d 0 ${outfile} > ${outfile}.tmp
        mv -f ${outfile}.tmp ${outfile}
        echo "  Adding model"
        map_addmodel -o 8 -d l ${outfile} > ${outfile}.tmp
        mv -f ${outfile}.tmp ${outfile}
        echo "  Fitting"
        map_fit ${outfile} > ${outfile}.tmp
        mv -f ${outfile}.tmp ${outfile}
```
where all of the variable names are taken care of outside of this.  Mostly showing this to show what options were used to create the file.  And at least at first glance this isn't a plotting issue as I am seeing that the cross polar cap potential in the map file is 1115 kV.  So is there something that I'm doing wrong here?  I'll admit I haven't looked at the `codebase/superdarn/src.bin/tk/testing/map_potential.1.0` directory.  I'll try to run through that next.

egthomas:
1. Directory version numbers have not been updated because it destroys the commit history.
2. We're merging this into `develop`, not `master` - I've been patiently waiting for a response to issue #7 for 3+ months now.  I think it does our community a disservice by assuming they can't eventually handle the `-old`/`-new` flag change.
3. The major/minor version number written to the map files can certainly be updated for the next release.
4. This is excellent - exactly the kinds of bugs that we're looking to find!  I'll get back to you on this.

ksterne:
A quick note about 2., I do agree that people are smart enough to figure it out.  But we should make sure to present the `-old`/`-new` flag as one of the major use changes with the next release.  It's more of a fact that it was not presented here and I only thought of it because of being apart of this team and knowing the discussion going on with #7.  Does this pull request then force the issue?  Another point related to this issue, this pull request is supposed to be teamed up with #52, but that pull request still uses `-new` flags.  So, if nothing else it would be good to have consistency between these two.

egthomas:
I thought I had fixed the southern hemisphere bug in commit 5111768 but apparently not.  If you change the call to CnvMapFitVector to use the absolute value of latmin (like in previous versions of the shf library) then it seems to work (more) correctly, but clearly the hemisphere corrections in CnvMapVlosMatrix based on the original code aren't working as I expected.  I'll look into this more closely tomorrow.

Yes, if we make the file format transition as advocated in #7 then we can add a statement about it to the release notes and/or readme file.  If we ever reach a decision on the issue it is minimal effort to change the new code back to using the `-new` flag or update the other binaries to use the `-old` flag.

All of the gridding routines in #52 have made the transition from the `-new` to `-old` flag as of commit b8cfd72 implemented on the `feature/grid_v2` branch 5-6 months ago so the two pull requests are consistent in that regard.

pasha-ponomarenko:
> I've been patiently waiting for a response to issue #7 for 3+ months now. I think it does our community a disservice by assuming they can't eventually handle the -old/-new flag change.

@egthomas, I do appreciate your patience! Seriously!
I think, everyone would agree that this switch is non-essential with respect to the data quality. On the other hand, it is potentially disruptive with respect to running the code. My initial question was: why has it been changed at all? I have just re-read the respective pull request and found, as I recalled before, that the justification is essentially based on the comment made by Rob Barnes. 

If it is that simple to revert back to the previous convention, I would do it now. Hopefully, this would expedite the testing/approval procedure. I appreciate that it is rather humiliating, but one would argue that this sort of convention issues should be discussed before their implementation and not the other way around. Of course, I am open to any other _contra_ and _pro_ arguments.

egthomas:
It's more than based on a comment by Rob as Simon and I explained in #7.

"It might be disruptive" is in my mind a poor reason not to implement not just this but any change in the software.  Again, in my opinion it is inaccurate to refer to dmap format files as "new" at this point.  By keeping the status quo we're catering to old timers and putting up an additional barrier for users who want to use any data from the last decade.  Is it a bad thing that people have to go back and look at what their scripts are doing once per decade?   Is our community that stuck in its ways and unable to adapt?  

If we had adhered to the "it might be disruptive" mantra we'd still be banging our heads against the wall trying to install VTRST3.5 and in the same poor software state as we had been for the previous ~5 years.

egthomas:
Also, I don't think the word "humiliating" has any place in a collaborative software development environment.  I'll leave it at that.

sshepherd:
I agree with Evan on both accounts. Make the change now or never.
Humiliating is a very poor choice of words.

Simon

pasha-ponomarenko:
Poor choice? I am not sure what do you mean, guys. I myself would feel a bit humiliated/ashamed/embarassed (Merriam-Webster) if needed to revert my code... 
:-(

egthomas:
southern hemisphere bug should be fixed as of commit 718badd (see attached figure):

pasha-ponomarenko:
>	"It might be disruptive" is in my mind a poor reason not to implement not just this but any change in the software. Again, in my opinion it is inaccurate to refer to dmap format files as "new" at this point. By keeping the status quo we're catering to old timers and putting up an additional barrier for users who want to use any data from the last decade. Is it a bad thing that people have to go back and look at what their scripts are doing once per decade? Is our community that stuck in its ways and unable to adapt?
If we had adhered to the "it might be disruptive" mantra we'd still be banging our heads against the wall trying to install VTRST3.5 and in the same poor software state as we had been for the previous ~5 years.

These are rather arbitrary statements, so I still see no actual necessity for changing the switch right now. Furthermore, it does cause problems even to the people who know about the modification (Kevin S).

In order to avoid further delays in testing and implementing more important changes to RST, as the DA-WG chair, I make an executive decision to stay with the `-new` flag. 

I take full responsibility for this decision. 

For track record purposes, I refer here to the original issue #7.

Evan, would you make the required modifications to your code please?

egthomas:
I can change the code, but I think it's laughable to suggest the switch from `-new` to `-old` flags has in any way delayed testing of this or any other pull request.  Case in point: the `feature/grid_v2` pull request  (#52) has been supposedly tested and deemed ready to be merged for months now yet it also made the `-new` to `-old` switch without any complaints.

I decided against doing this yesterday but now I'm going to go ahead and just say it.  What kind of working group environment are we trying to foster here?  One where it's considered "humiliating" or "embarrassing" to be asked to change your code?  I'm not ashamed of anything I've contributed to this repository and would never expect anyone else to feel that way about their work.  That kind of language does nothing but discourage others from contributing to the software discussion and/or development process.

pasha-ponomarenko:
@egthomas, I believe I made it clear that I meant no insult to anyone, leaving alone promoting unhealthy working environment. Would "feeling frustrated" work better?

Now, you asked for a descision, and I made it. Let's go back to work.

sshepherd:
I'll chime in here. I am suggesting that Evan, a postdoc who much better
things to do with his time, stop contributing to the RST. The WG has become
quite disfunctional. Significant contributions are met with resistance and
weak excuses for not being implemented. There is a lack of understanding of
the RST and little direction. I think Theo's WG needs a new structure.

Simon

ksterne:
I feel compelled to join this discussion since my name is being used.  My complaint about the problem wasn't that we were switching flags, but that in the pull request comments and notes here there is no mention or 'heads up' that we are switching flags.  @egthomas seems to have addressed this concern by noting that when we make the next release we document that the appropriate binaries have switched from `-new` to `-old`.  This isn't exactly my preference as better software development would add a deprecation warning to anyone that is still trying to use `-new`.  But I seem alone on that compromise so I'll let it drop.

Otherwise, I'll need to revisit #52 as the code I'm using to process things is still using the `-new` options for `make_grid` and `combine_grid`.  So something may be wrong with my testing there which doesn't necessarily make my tests here completely valid.

pasha-ponomarenko:
Well, what about using both flags for a time being but the non-flagged default calling "new" function? Plus a deprecation warning for the following release. That would not cause much problems and seems to be a good compromise.

egthomas:
Command line option error handling similar to what was implemented in #102 has also been added here with some additional info.  For example,

```
> map_addhmb -idl_rules 20161030.south.empty.cnvmap > 20161030.south.hmb.cnvmap
Option not recognized: idl_rules
Please try: map_addhmb --help
> map_plot -davitpy_drools 20161030.south.empty.cnvmap 
Option not recognized: davitpy_drools
Please try: map_plot --help
```

ksterne:
So, I started getting a seg fault here, but I think I've tracked that down to environment variables being off.

I'm going to manually check what environment variables the mapping binaries need and what they're getting set to in my environment.  Maybe I just need to re-source the environment variable setup (env_rst).

So, it seems to work doing it one by one, but the shell script I've got needed some more issues as I'm still getting seg-faults....ugh.  I thought I was getting somewhere here.

image

I copied the command line resulting file into the archive and made the above plot.  Seems as though there's something backwards.  And I redid it with the command line `-sh` and didn't see any change.  I'm not sure if it's something with the plotting or if it's in the data.

egthomas:
Can you try plotting it with `map_plot` and the following options?

`map_plot -x -cf my.options xxxx.cnvmap`

where `my.options` is a text file containing the following values:
```
-rotate -fit -flip -grd -grdontop -latmin 50 -tmlbl -time -dn -logo -imf -mag -hmb -coast -modn -pot -ctr -hmb -fcoast -fterm -coast -vmax 1000 -extra -frame -maxmin -vkey superdarn.key -vkeyp
```

egthomas:
Below are the steps for generating an empty map file and running it through the Map Potential procedure for an arbitrary date with only statistical model vectors (in this case CS10 using the trilinear interpolation option):

```
map_grd -empty -sd 20170101 > 20170101.north.empty.cnvmap
map_addhmb -nodef 20170101.north.empty.cnvmap > 20170101.north.hmb.cnvmap
map_addimf -if imf.txt -d 0 20170101.north.hmb.cnvmap > 20170101.north.imf.cnvmap
map_addmodel -cs10 -o 8 20170101.north.imf.cnvmap > 20170101.north.model.cs10.cnvmap
map_fit 20170101.north.model.cs10.cnvmap > 20170101.north.cs10.cnvmap
```

I've attached a text file containing fake IMF data called [imf.txt](https://github.com/SuperDARN/rst/files/1418952/imf.txt) for use with `map_addimf` which cycles through all clock angles and increases in magnitude each hour.  I've also attached a text file called [myoptions.txt](https://github.com/SuperDARN/rst/files/1418963/myoptions.txt) for use with `map_plot` which contains useful plotting/graphics options for viewing the Map Potential results:

```
map_plot -x -cf myoptions.txt 20170101.north.cs10.cnvmap
```

Note that if you want to test Southern Hemisphere results you'll need to set the `-sh` option with `map_grd` and the `-flip` option with `map_plot`.

pasha-ponomarenko:
@egthomas, thanks a lot! I will have a close look at it by the end of the week (weekend included).

pasha-ponomarenko:
Installed the branch, checked for both hemispheres. Seems to behave like expected. ;-)

pasha-ponomarenko:
I will try to generate some PS plots...

pasha-ponomarenko:
Something like this:

[20170322_0230SH.pdf](https://github.com/SuperDARN/rst/files/1428247/20170322_0230SH.pdf)

egthomas:
I think that looks reasonable, although sometimes it takes me a while to wrap my head around the southern hemisphere plotting conventions (ie through the earth vs below the earth, etc).

pasha-ponomarenko:
> sometimes it takes me a while to wrap my head around the southern hemisphere plotting conventions (ie through the earth vs below the earth, etc).


Well, I have got used to it since a while ago... 
;-)

I've had a second opinion from Sasha, and he also said that the reverse fitting of the statistical model makes sense with respect to IMF magnitude and orientation and . 
In a meantime, I unintentionally tested the option error feature too...  
I might play around with the maximum harmonic order a bit but what else could be useful to test here?

egthomas:
As far as testing without data you could try running the CS10 model without the trilinear interpolation option (`-nointerp`) to see the discrete patterns or you could try using the RG96 (`-rg96`) or PSR10 (`-psr10`) models instead.  You can also try cranking up the order of the statistical model expansion to something like 30 to add more model vectors for visualization while still doing the final `map_fit` with an 8th order/degree expansion.  That can be fun to test the fidelity of the reverse convection cells under northward IMF with the CS10 or PSR10 models.  Otherwise you can run some real grid files through the Map Potential process and make sure the results are sensible.

pasha-ponomarenko:
Being an RST dummy, may I ask you a favour to give me a hint or two on generating grid files from real data?

ksterne:
The system I created the file on is a GUI-less one (Ubuntu server), so maybe it would be better to compare data within the file?  Otherwise, I took the same map file and copied it to another system that does have a GUI and tried out the map_plot binary.  I guess that `map_plot` command was supposed to make a movie?  I'm getting a plot that's updating with each 2-minutes of data.  I'm also still getting the "+" marker on the dusk side (marked 18 mlt) which seems to be backwards with what you had.  Again, there might be something going on with old environment variables being set with the file I was generating.  I'll see if I can find some hour or two to try to reprocess the files.

egthomas:
The other odd thing is that in the original IDL figure you posted (on Oct 25) the vectors appear to be directed radially towards or away from the radar sites rather than in the direction of the contour lines.  In other words, they look more like grid vectors than fitted vectors.

egthomas:
@pasha-ponomarenko you can either call `make_grid` individually with several fitacf files and then use `combine_grid` to combine them into one daily file, or you can pass multiple files to `make_grid` with the `-c` command line option.

ksterne:
Hey @egthomas, I just got things working again and am fairly confident everything is good with using code from fresh installs.  I'm still seeing the positive of the PCP being on the dusk side, but this seems to be normal if the VT plotting tools are looking down from the north pole on things.  Something's not quite right though as it has convection going from 0 MLT *towards* noon... Did you see my question asking about the `map_plot` command?  I'm not as familiar with it and could use some help with it.

egthomas:
@ksterne can you provide more specific examples of the exact processing steps you're using? It would also be helpful if you could upload the original grid file and final map file you're getting so I can suggest some plotting options for map_plot (although I'm pretty sure we changed the '-def' option of 'map_plot' to do something meaningful).

ksterne:
Here's the map processing commands I'm doing:

```
		map_grd ${switch} ${infile} > ${outfile}
		echo "  Adding HM boundary"
		map_addhmb ${outfile} > ${outfile}.tmp
		mv -f ${outfile}.tmp ${outfile}
		echo "  Adding IMF"
		# map_addimf -new -if ${imffile} -df ${delayfile} ${outfile} > ${outfile}.tmp
		map_addimf -if ${imffile} -d 0 ${outfile} > ${outfile}.tmp
		mv -f ${outfile}.tmp ${outfile}
		echo "  Adding model"
		map_addmodel -o 8 -d l ${outfile} > ${outfile}.tmp
		mv -f ${outfile}.tmp ${outfile}
		echo "  Fitting"
		map_fit ${outfile} > ${outfile}.tmp
		mv -f ${outfile}.tmp ${outfile}
```

Here, `switch` is set to `-sh`; `infile` is the grid file that I've posted a link to below.  I've also posted a link to the imf file that I'm using for `imffile`.  Use whatever you'd like for `outfile`.

You can find the files at:

[vt.superdarn.org/share/rst_github_files/pr100/](http://vt.superdarn.org/share/rst_github_files/pr100/)

egthomas:
@ksterne Commit 8181879 should correct the southern hemisphere errors you're seeing.  Let me know if you're still seeing incorrect results.

ksterne:
Hey @egthomas, sounds good that there's a correction here.  Hooray testing!  Did you confirm that the grid file I provided looked right?  It could be a case of garbage in, garbage out.  Unfortunately I won't be able to resume testing until things get straightened out with our servers.

egthomas:
@ksterne the grid file looks fine to me.  The issue was that a lot of the map potential fitting code was hardcoded to treat everything as northern hemisphere data and ignored the southern hemisphere if/else branches by taking absolute values of latitudes.  I had tried to fix things so that the code worked a little more logically but ended up introducing some error for the southern hemisphere that I still don't fully understand. There are still a couple of places where I think the original code might be wrong but I I won't have time to look into that for a while.

pasha-ponomarenko:
@egthomas, I have a couple of questions:
(1) Where there any changes made to the way the velocity errors are treated? I recall that anything below 100 m/s was set to that value. Keeping in mind that the realistic v_e values for "good" ACFs are around ~5-10 m/s and those with 100 m/s are essentially garbage, this "equilisation" would make little sense...
(2) Based on your presentation here earlier this year, Kathryn inquired if there is now a way to select 1/2 hop data? I guess the answer is that one just limits the group range, and that's it.

I would love to play around with gridding/mapping in order to see the difference between v2 and v3 of FITACF, and I hope to be able to do it within the next month. 

BTW, is there anything really urgent to look at which lies within my limited abilities?

egthomas:
@pasha-ponomarenko :

(1) I have not adjusted the way velocity errors are handled in either the gridding or map potential process.  Below are the places where floors/bounds are applied to the velocity errors:

- Exclude fit-level data with velocity errors outside minimum and maximum values in `make_grid` (0 < v_e < 200 m/s by default, can be modified or disabled entirely via command line options):
https://github.com/SuperDARN/rst/blob/feature/grid_v2/codebase/superdarn/src.bin/tk/tool/make_grid.2.0/make_grid.c#L809
https://github.com/SuperDARN/rst/blob/feature/grid_v2/codebase/superdarn/src.lib/tk/filter.1.8/src/bound.c#L114-L120

- Apply floor of 100 m/s on fit-level velocity error before calculating weighted average in `GridTableMap`:
https://github.com/SuperDARN/rst/blob/feature/grid_v2/codebase/superdarn/src.lib/tk/gtable.2.0/src/gtable.c#L505-L506

- Apply floor of 100 m/s on grid vector velocity errors before the map potential fitting procedure in `CnvMapFitMap`:
https://github.com/SuperDARN/rst/blob/feature/map_potential_v2/codebase/superdarn/src.lib/tk/shf.1.10/src/fitmap.c#L107

(2) The fit-level data used as input to `make_grid` can now be filtered by minimum and/or maximum slant range values via the new `-minsrng` and `-maxsrng` command line options (which is probably as close as we can get to selecting for 0.5-hop scatter without elevation information [but the hooks are already there in the rpos library to use elevation angles instead of a virtual height model]):
https://github.com/SuperDARN/rst/blob/feature/grid_v2/codebase/superdarn/src.bin/tk/tool/make_grid.2.0/make_grid.c#L159-L174

I'd like to be able to store the slant range information directly in the grid files to allow post-processing but that would require either altering the existing grid file format (which would break the old ascii-style .grd file) or replacing one of the existing data fields.

ksterne:
@egthomas, our servers seem to be stabilizing, so I've gotten to testing this pull request again.  Looks like the last commit has fixed the southern hemisphere data.  The figure below with with data generated using this and #52.  It looks as though there's a small difference than your last plot in the cusp area, but generally it's pretty close.

image

I'm quickly taking a look at these same two days (2016103*) for the northern hemisphere to make sure nothing's been broken there either.  They look pretty good and reasonable at first go.  Otherwise, are we finally ready to merge this and #52?  @pasha-ponomarenko, do you have any other concerns before these two pull requests are merged in?

pasha-ponomarenko:
@ksterne, its 1am here, so I wonder how much you actually sleep... :-o
Anyways, my concerns and questions are more related to the original algorithms and statistical procedures rather than to their implementation. I think we (mostly you!) did as much as we can under the circumstances and can go ahead with merging these two branches. 
I also think that #74 is almost there. I just need to look at one or two things, and then we will be ready to go too.

egthomas:
@ksterne glad it's working for you now.  The patterns look slightly different because different solar wind conditions have been used to drive the CS10 model in each of our figures.  Also, if you're using the same parsed IMF text files note that you'll need to add an another column for Vx as that value is necessary to calculate Esw (rather than simply using Bt for the magnitude as in the RG96 and PSR10 models).

ksterne:
Hey @pasha-ponomarenko, thanks for the concern, but I'm on Pacific ST for the next few days.  I'm working on another project which is why I was working on this during later hours.

@egthomas, thanks for the note on the different input format for the IMF file.  I'll have to take a look at that sometime.

I'll work on merging #52 and this this evening.  @egthomas, do you have any notes on what will need to be changed to fix the merge conflict caused by merging both of these PRs?  I think you said it was pretty minor.

egthomas:
If this has working group approval then it would be much simpler for me to merge this pull request and #52 to deal with the merge conflicts.

pasha-ponomarenko:
I guess this is not the approval itslef, that has to be done by someone from a different institution, but its mere implementation. 
So, if @ksterne (or anyone else) has no objections for @egthomas merging these two PRs and resolving the arising conflicts, this would probably be optimal.

ksterne:
@egthomas, have at it!  I'll be glad to see these requests in the develop branch.  With these done, I'll turn my attention on #74.

PR #107, opened 2017-10-24
Fixing dlm return value of AACGM_v2_GetDateTime on error

egthomas:
This commit brings the return value of the DLM version of `AACGM_v2_GetDateTime` in sync with the IDL implementation. The IDL function prints an error message and returns -1 if the date/time has not yet been set while the C version always returns 0. 

https://github.com/SuperDARN/rst/blob/develop/codebase/analysis/src.idl/lib/aacgm.2.0/aacgm_v2.pro#L203-L207
https://github.com/SuperDARN/rst/blob/develop/codebase/analysis/src.lib/aacgm_v2/aacgm.1.0/src/aacgmlib_v2.c#L1137-L1145

This commit checks for year==-1 after the call to `AACGM_v2_GetDateTime` in the DLM and will print the same error message and return -1 if appropriate; otherwise the date/time values are stored in the keywords as normal.

### DIFF
--- a/codebase/analysis/src.dlm/aacgmdlm.2.0/aacgmdlm.c
+++ b/codebase/analysis/src.dlm/aacgmdlm.2.0/aacgmdlm.c
@@ -25,6 +25,7 @@
 #define AACGM_MISMATCHDIMS -1
 #define AACGM_MISMATCHELMS -2
 #define AACGM_BADTYPE -3
+#define AACGM_DATENOTSET -4
 
 static IDL_MSG_DEF msg_arr[] =
   {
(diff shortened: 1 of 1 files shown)


pasha-ponomarenko:
Is testing this consists of just checking if the right error message is generated?

egthomas:
This is what you get without DLMs enabled on the current develop branch when calling `AACGM_v2_GetDateTime()` without first setting the date/time values (ie using native IDL functions):
```
IDL> print,AACGM_v2_GetDateTime(year,month=month,day=day)
Date and Time are not currently set
      -1
```
This is what you get with DLMs enabled on the current develop branch (ie using C functions):
```
IDL> print,AACGM_v2_GetDateTime(year,month=month,day=day)
% Loaded DLM: AACGMDLM.
           0
```
And this is what you get with DLMs enabled on this pull request branch:
```
IDL> print,AACGM_v2_GetDateTime(year,month=month,day=day)
% Loaded DLM: AACGMDLM.
% AACGM_V2_GETDATETIME: Date and Time are not currently set
          -1
```
So yes, you should now be able to see that the IDL and DLM error messages and return values agree with one another.

pasha-ponomarenko:
Seems to work as intended:
`IDL> print,AACGM_v2_GetDateTime(year,month=month,day=day)

% AACGM_V2_GETDATETIME: Date and Time are not currently set

          -1
`

PR #111, opened 2017-11-23
File format transition

egthomas:
This pull request completes the old/new file format transition and also adds command line option error handling to the impacted binaries.


ksterne:
I guess we should go off of the commit names for all of the binaries to test `-old` here?

ksterne:
Bump, test here is to go off the commit names for all the binaries to test?

egthomas:
Sure, although I can't speak to whether every single binary actually worked in the first place.  At they very least now they should dump a warning message and exit if you try passing in the `-new` (or any other unrecognized) flag.

ksterne:
Yea, on that, it doesn't seem like `fitcomp` actually uses the flag, or read in arguments on it own.  It seems as though any arguments are passed in to the main function.  

`Test_fitex2` also doesn't behave as usual, though if you try to give it a `-new` it at least gives a warning, but `--help` doesn't offer anything (outside of this PR, but I'll start an issue to note it).  Same with `test_lmfit`. 

Otherwise these all look good.  I'll merge this in now.

egthomas:
A `doc/` directory with a properly formatted `.doc.xml` file is necessary to generate the command line `--help` output.  A lot (but not all) of the binaries with missing documentation/help info seem to be written by AJ.

PR #112, opened 2017-11-27
Adding checks for backwards compatibility for pre-v2 map files

ksterne:
I think this fixes the backwards compatibility issue of the new grid/map files.  My base test was to try to get an output of a map version 2 file with `map_ascii` and would get an error reading output with the 4.1 release branch.

I am unsure though if these changes break anything else in the code.


egthomas:
Tested and approved.

Issue #114, opened 2017-12-07
missing hdw.dat.svb


asreimer:
The `hdw.dat` file for SPEAR is missing from the this repository.

If you look at [the radars.dat file](https://github.com/SuperDARN/rst/blob/master/tables/superdarn/radar.dat) it references SPEAR.

However, [the tables/hdw  directory](https://github.com/SuperDARN/rst/tree/master/tables/superdarn/hdw) doesn't have the hdw.dat.svb file.

This issue also occurs for the [VT hdw.dat repo](https://github.com/vtsuperdarn/hdw.dat). [I've also filed an issue there.](https://github.com/vtsuperdarn/hdw.dat/issues/5)

egthomas:
Huh, didn't realize there was a hdw.dat file for SPEAR in pre-VTRST3.5 releases.  I'll create a pull request for a copy we have from version 3.3 of the RST.

asreimer:
Frankly, I was as surprised as you. Thanks for fixing this.

egthomas:
I think the brief but spirited discussion in the closed #115 pull request also brings this issue to a close - the `hdw.dat.svb` file will not be returned to the RST and the SPEAR entry has been removed from the `radar.dat` file on the 4.1 release branch.

PR #115, opened 2017-12-07
Adding hdw.dat file for SPEAR from RST 3.3

egthomas:
This pull request addresses issue #114 raised by @asreimer.  For some reason this file existed in versions 3.3 and earlier of the RST but not VTRST3.5.  This is just adding a missing hdw.dat file, so I don't seem the harm in committing it directly to the 4.1 release branch.

### DIFF
--- a/tables/superdarn/hdw/hdw.dat.svb
+++ b/tables/superdarn/hdw/hdw.dat.svb
@@ -1,0 +1,101 @@
+#  Hardware parameters for Svalbard radar.
+#
+# Each radar has a distinct set of hardware parameters that are used
+# by the radar control software and the analysis software. These
+# parameters are read in a distinct order and are assumed to have
+# specific units. If either the order of the parameters or their units
+# are incorrect, the processing and analysis software will produce
+# incorrect answers that may not be easily identified. It is the
+# responsibility of the SuperDARN P.I.s to insure that the hdw.dat files
+# for their radars are correct and that these files are updated as
+# required to accurately represent the physical state of the radar and
+# that copies of these files are retained under revision control by Rob
+# Barnes. Revision controlled versions of hdw.dat files are destributed
+# with SuperDARN radar control software and with analysis software.
+#
+# The hardware parameters are distributed as a string of values
+# delineated by one or more spaces. The following table specifies the
+# parameters, their units, and a brief description of their meaning.
+#
+# 01) Station ID (unique numerical value). Assigned by Rob Barnes.
+# 02) Last year that parameter string is valid. (4 digit year).
+# 03) Last second of year that parameter string is valid (range 0 to
+#     34163999 for non-leap years). The parameter string giving the current
+#     configuration is assumed to be valid until the last second of 2999.
+# 04) Geographic latitude of radar site (Given in decimal degrees to 3
+#     decimal places. Southern hemisphere values are negative)
+# 05) Geographic longitude of radar site (Given in decimal degrees to
+#     3 decimal places. West longitude values are negative)
+# 06) Altitude of the radar site (meters)
+# 07) Scanning boresight (Direction of the center beam, measured in
+#     degrees relative to geographic north. CCW rotations are negative.)
+# 08) Beam separation (Angular separation in degrees between adjacent
+#     beams. Normally 3.24 degrees)
+# 09) Velocity sign (At the radar level, backscattered signals with
+#     frequencies above the transmitted frequency are assigned positive
+#     Doppler velocities while backscattered signals with frequencies below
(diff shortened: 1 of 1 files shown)


egthomas:
I take back the part about the SPEAR hdw.dat file not being in VTRST3.5 - apparently @sshepherd removed it during his cleanup of hidden version files in commit 6fd1cfc before the 4.0 release.

ksterne:
I guess the thought process is that we don't have any svb data, so should we keep the hdw.dat file?  I'd think we could for legacy and it's still in the radar.dat file.

asreimer:
It is really weird that it's even there. What's the history on this?

From a user point of view, I don't care if it is there or not, except that it breaks expectations if something is in radar.dat but there isn't a hdw.dat file for it. Principle of least-surprise and all that.

pasha-ponomarenko:
SPEAR is (was?) a "heater" whose antenna has a desing which is very different to that of SuperDARN:

> The facility consists of 4 X 6 array of rhombically broadened crossed-dipole antennas, with a distributed transmitter system consisting of 48 X 4 kW transmitters. This configuration results in a frequency-dependent beam width of approximately 14° x 21° with a gain of 21 dB.

(JOURNAL OF GEOPHYSICAL RESEARCH, VOL. 117, A01307, doi:10.1029/2011JA017079, 2012)

In this respect the stored hardware information 

> 128 2999 31556736   78.150   16.050    50.0   25.0  3.24  1   10  0.478  1    1.5  100.0    0.0 100.0  2  75 16

makes no sense except, probably, the coordinates.  

I recall that the Brits were trying to use SPEAR for some sort of SD-like sounding, so this entry might represent some sort of a patch-up/placeholder in order to enable processing of their data with FITACF. 

It seems to be harmles enough to keep this hdw file, but I also think Simon had a good practical reason to remove it... We might ask Angeline if she can clarify this matter a bit more.

asreimer:
Ya SPEAR **was** a heater. Unfortunately it's gone. Lisa Baddelely might know more since she was PI until it was decomissioned.

If I remember correctly, the LYB radar is sitting where SPEAR was, or at least very close to that location (can't remember if that was just the initially planned location or not).

@aburrell, do you know anything about why SPEAR would be in the SuperDARN radar.dat?

egthomas:
As @pasha-ponomarenko suggested, my guess would be that the Leicester crew wanted to use SuperDARN / Go software with SPEAR and needed some pseudo-hardware information to get the code to work properly.  Since this was mistakenly removed the first time around, what's the harm in returning it to the RST?

ecbland:
Yes, SPEAR has been decommissioned and replaced with the new Svalbard SuperDARN radar at the same site. 

A message from Lisa:
> The hardware file for SPEAR was there as it was originally intended that SPEAR should also be able to function as a high power SuperDARN radar.  It did work but the problem was that the back lobe was almost as big as the front lobe when we lowered the pointing direction from heating mode (vertical/parallel to B) to SuperDARN mode (20 degrees elevation ish).  Hence we couldn't actually tell where our scatter was coming from. So - in conclusion you can probably remove it from the software. You could send an email to Tim Yeoman at Leicester just to confirm but I think there is no data around that would need the SPEAR hardware file.

And to add my two cents, I think the file should be removed because:
- "It is harmless" is not in itself a good argument to keep it, especially given the recent efforts to clean up other aspects of RST. In fact, there is potential for confusion between SPEAR (code SVB) and the new Svalbard SuperDARN radar (code LYR). 
- The PI (Lisa) has indicated that there is no data from SPEAR that would require the hardware file. And obviously no such data will be produced in the future because the instrument has been decommissioned.
- Most of the information in the SPEAR hardware file is meaningless
- Perhaps the SPEAR hardware file could be archived alongside the SPEAR data
- It sounds like SPEAR was never "officially" part of SuperDARN, so why should we support it?

egthomas:
Thanks for the info @ecbland!  In that case I can close this pull request / delete the branch, and remove the SPEAR line from the radar.dat file on the `4.1/release` branch.  That way we're consistent with no references to SPEAR anywhere in the RST.  Agreed?

egthomas:
Looking through my emails, I did contact Mark Lester about the SPEAR hdw.dat file back in March/April of this year but Tim/Julian never spoke up in defense of it.

ksterne:
@egthomas, I'd think if we're not going to have the SPEAR radar hdw.dat file then the entry should be removed from the radar.dat file.  That's how things came to be since there's a line in that file.  I do agree with the point that the name hdw.dat.svb could be confused with for the lyr radar.

sshepherd:
Agreed.

pasha-ponomarenko:
It was quick and efficient!
;-)
Good team work!

PR #110, opened 2017-11-17
4.1/release

ksterne:
This serves as the 4.1 release pull request.

I believe the first order of business is to get the 4.1 release branch to compile.  From a fresh install, I'm getting an error in the grid_plot.c code at:

https://github.com/SuperDARN/rst/blob/a51c6fa3706ef23569b05e6de9b720fc57195e0d/codebase/superdarn/src.bin/tk/plot/grid_plot.1.12/grid_plot.c#L904

Here there are only two arguments where the make_fov.c function lists 3 arguments should be passed.  This is also confusing since there are make_fov.[ch] files in grid_plot, fov_plot, field_plot, and map_plot.

Since this is probably related to #52 or #100, @egthomas, @sshepherd can you comment on why this is breaking potiently apply the bugfix directly to the 4.1/release branch?

Issue #95, opened 2017-08-29
4.1 release goals and timeline


egthomas:
Since we're coming up on 3 months post-4.0 I thought it might be a good idea if we start laying out our goals for the next release.  At the workshop we had discussed having version 4.1 ready around this time (September) but I think December is probably more realistic given our previous pace.  Or would we rather wait until next year's workshop for a new release?

Here are the key pieces I think we should have ready for the next release:
1) revised version of FITACF library
2) revised version of gridding software
3) revised version of Map Potential software
4) better Mac support

There are currently outstanding pull requests for items 1 and 2 on the `add_fitacf.3.0` and `feature/grid_v2` branches respectively, both of which I personally feel could use some additional testing from additional users.  @sshepherd  and @aburrell have gotten started on items 3 and 4 on the `feature/map_potential_v2` and `macOS_installation` branches respectively.  I know there is still a lot of work we need to do on the Map Potential branch; @aburrell - do you feel satisfied with the revised README instructions for Mac installation?

There are a few other issues we should also try to address, including (but not limited to):
5) paths for custom color tables (apologies to @ecbland for breaking this in the first place)
6) `old` vs `new` flags
7) licensing
8) etc

ksterne:
Are you thinking of trying for a release a week or so prior to the fall AGU meeting (say Dec. 5th)?  I'm on board with this (if it wasn't obvious from the off-git discussion back in June).  Would you also be in favor of making a `4.1-release` branch about 4 weeks prior to release time (say Nov. 7th) to allow the code to stablize and be better tested?  I haven't looked at it recently, but the merge process should be a lot cleaner this time.

That sounds good on goals for the 4.1 release, is it reasonable that these will be finished by early November?  Do we think 3 would be ready in time for December?  Maybe taking a step back, are goals 1-4 too much of a jump between versions (or more so is too much changing at once)?  Comments on goals (relating to how they were numbered):

1. It seems as though #74 has gotten into some deep discussions on how this code should be implemented and interfaced with dating back to June.  I'll admit I haven't kept up with this one as my focus has been trying to work with #52.

2. Speaking of, it seems as though this is ready to go.  I'm not sure what the hold up is on this.  More testing?  I've tested this a few times and haven't found anything wrong.  This pull request seems to have been ready for a few months though maybe the hold is relating to the median filter discussion dating back to a month ago.

6. It seems as though this is somewhat a broader discussion that might be solved in a working group meeting.

[...]

PR #118, opened 2017-12-12
Adding new cnvmap file format compatibility to native IDL and DLM routines

egthomas:
This pull request adds compatibility to the native IDL and C-based DLM routines for handling the new cnvmap format files (which have new data fields such as Vx and tilt) while also maintaining backwards compatibility with the older format files (ie those without Vx and tilt).  @sshepherd already did this for the C code but not the IDL/DLMs.

I think we should hold off on merging the release branch today so that we can include this compatibility fix and ensure consistency across both IDL and C when handling the newer map_potential_v2 type files.

### DIFF
--- a/codebase/superdarn/src.idl/lib/main.1.25/cnvmap.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/cnvmap.pro
@@ -303,8 +303,8 @@ function CnvMapRead,unit,prm,stvec,gvec,mvec,coef,bvec
       
   if (sclid[25] ne -1) then prm.imf_model[0]=*(sclvec[sclid[25]].ptr)
   if (sclid[26] ne -1) then prm.imf_model[1]=*(sclvec[sclid[26]].ptr)
-  if (sclid[27] ne -1) then prm.imf_model[1]=*(sclvec[sclid[27]].ptr)
-  if (sclid[28] ne -1) then prm.imf_model[1]=*(sclvec[sclid[28]].ptr)
+  if (sclid[27] ne -1) then prm.imf_model[2]=*(sclvec[sclid[27]].ptr)
+  if (sclid[28] ne -1) then prm.imf_model[3]=*(sclvec[sclid[28]].ptr)
 
   prm.hemisphere=*(sclvec[sclid[29]].ptr)
   if (sclid[30] ne -1) then prm.noigrf=*(sclvec[sclid[30]].ptr)
--- a/codebase/superdarn/src.idl/lib/main.1.25/cnvmap.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/cnvmap.pro
@@ -85,8 +85,11 @@ pro CnvMapMakePrm,prm
                   Bx: 0.0D, $
                   By: 0.0D, $
                   Bz: 0.0D, $
-                  imf_model: strarr(2), $
+                  Vx: 0.0D, $
+                  tilt: 0.0D, $
+                  imf_model: strarr(4), $
                   hemisphere: 0, $
+                  noigrf: 0, $
                   fit_order: 0, $
                   latmin: 0.0, $
                   coefnum: 0L, $
@@ -178,9 +181,14 @@ function CnvMapRead,unit,prm,stvec,gvec,mvec,coef,bvec
           'IMF.Bx', $
           'IMF.By', $
           'IMF.Bz', $
+          'IMF.Vx', $
+          'IMF.tilt', $
           'model.angle', $
           'model.level', $
+          'model.tilt', $
+          'model.name', $
           'hemisphere', $
+          'noigrf', $
           'fit.order', $
           'latmin', $
           'chi.sqr', $ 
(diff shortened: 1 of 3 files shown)


ksterne:
Really? Hours within the release time?  I feel like this is starting to be scope creep for what we outlined in #95.  I am not in favor of waiting on this pull request for a new release and this can be added with the next release.  @pasha-ponomarenko, any thoughts?

If we are going to delay the release waiting for this pull request, I'd propose we delay until Jan. 16th.  This is starting to get way far away from our original goal of Dec. 5th.

Aside from all of that, do you have a good test for this as I'm not as familiar with using the IDL part or exercising the cnvmap routines?

egthomas:
Knowingly releasing a broken product today would not make a good impression on the rest of the SD community. What's the point of a testing window if the fixes that arise from testing can't be implemented until 6 months later?

ksterne:
I get the point of releasing something that's knowningly broken.  Though how much is this used?  I'd think this has been in the testing realm since #100 was merged a month ago and it's just now coming to light?  That seems like it's pretty low use.

Maybe pushing off the release until mid-Jan is a bit far, but I'm not sure much will happen this week or next for testing.  So at earliest, the first part of January seems reasonable.

Thinking of testing, do you have some example code to test this out?  I haven't run into this in my current testing of #110.

ksterne:
Sounds good all. Sorry, I get this now is a way to take pre-grid v2 files and make map v2 files...right?  I'm more on board with holding off, though it also doesn't seem like good practice to continually miss deadlines.  I stick by that this feature or functionality wasn't outlined in #95 so it's a little extra that we're doing here.  I'll try to get to testing this at some point...hopefully before the holidays, but anyone else feel free to test.

So, are we good to move the release date back to Jan. 16th?  Maybe that's too much time and Jan. 9th would be better?  I like that date more to try to get this wrapped up.  Thoughts?

pasha-ponomarenko:
9 January seems OK for me.

egthomas:
No, it's not related to grid files, only map files.  Basically the map files produced with RST 4.1 will contain some additional data fields that were not present in map files from RST 4.0 and earlier.  @sshepherd and @ksterne made the necessary changes so the C I/O routines could handle both pre- and post-4.1 files, but the same adjustments were never made to the IDL code.  So for example, the IDL convection map plotting software for the VT website would not be able to read all of the model, tilt, and solar wind information from a post-4.1 map file.

pasha-ponomarenko:
>the prm structure will be missing crucial Vx, tilt, IGRF, and model bin information 

So, I can see sensible values in `vx`, `bx`, `by`, `bz` and `tilt` fields of the `prm` structure. I wonder which one contains the **_model bin info_** though...

egthomas:
The prm.imf_model string array contains either the model bin ranges or I think the exact values used for the trilinear interpolation available with the CS10 model.

pasha-ponomarenko:
In that field I can see four parameters: Bang, Esw, tilt and model (CS10). 
Is this correct?

egthomas:
Yes I think so.

pasha-ponomarenko:
OK, I was able to read old map files without any problem, and the `prm `structure has  `vx`, `bx`, `by`, `bz`, `tilt` and `imf_model` fields in it.

pasha-ponomarenko:
Should anything else be tested here?

egthomas:
No that pretty much covers it.

pasha-ponomarenko:
Is there an IDL version of map_plot in RST?

egthomas:
No, there aren't any IDL plotting tools in the RST.

pasha-ponomarenko:
So, is it done now and can be merged? What do you think, @ksterne ?

ksterne:
Just getting around to testing this.  So, I've modified the [map plotting tool](http://vt.superdarn.org/tiki-index.php?page=DaViT+Map+Potential+Plot) on the VT website and nothing seems to break.  My initial impression is the that things would break when plotting the map2 files.  Right now I've only got 2017032[34] map2 files processed for testing.  I see it's still got the RG96 stat model in, but I wonder if that's hardcoded in the plotting software.

I'll try out running the IDL example manually, but @pasha-ponomarenko, if you think this is good to go, don't wait on me.

pasha-ponomarenko:
> I'll  try out running the IDL example manually...

We'd better wait for the outcome because the VT website is the main (or even the only?) user of these IDL routines.

egthomas:
Pretty sure the VT IDL plotting software has the RG96 label hardcoded (since until now that has been the only model available).  It will need to be updated once we start using CS10 or TS17 as the default for map files.

ksterne:
I can confirm this seems to work.  Good catch on this @egthomas, even if the timing was frustrating.  I'm able to run the example code you provided and print out values for `prm.bx`, `prm.by`, `prm.bz`, `prm.tilt`, `prm.vx` and `prm.imf_model`.  In addition, I've updated the map plot tool on the VT website to look for and print out the imf_model info a little bit better.  I'll go ahead and merge this in.

As a sidebar, finally got the Vx info included the IMF file which wraps up a note from @egthomas, in #100.</sidebar>

PR #121, opened 2017-12-22
Fixing array indexing in IDL versions of RadarDecodeRadarPrm and FitCnxRead

egthomas:
I swear this is a bugfix, everyone!  Put down the pitchforks!

This pull request fixes several typos in the indexing within the IDL version of RadarDecodeRadarPrm.  Before, if you didn't have DLMs enabled then functions like FitRead would always return the lvmax value instead of the actual combf string.  Also, if the ifmode value was not present (which appears to sometimes happen with older FITACF library versions) then the nrang value of the prm structure would be zero, causing problems with interpreting the radar data and crashing my higher level code.  You can try to test this by disabling the DLMs - or you could just compare the indices of the variables being checked for -1 to those being stored in the prm structure.

### DIFF
--- a/codebase/superdarn/src.idl/lib/main.1.25/fitcnx.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/fitcnx.pro
@@ -195,7 +195,7 @@ function FitCnxRead,unit,prm,fit
   else fit.qflg[slist]=1
   if (arrid[6] ne -1) then fit.gflg[slist]= (*(arrvec[arrid[6]].ptr))[*]
   if (arrid[7] ne -1) then fit.p_l[slist]= (*(arrvec[arrid[7]].ptr))[*]
-  if (arrid[9] ne -1) then fit.p_l_e[slist]= (*(arrvec[arrid[8]].ptr))[*]
+  if (arrid[8] ne -1) then fit.p_l_e[slist]= (*(arrvec[arrid[8]].ptr))[*]
   if (arrid[9] ne -1) then fit.p_s[slist]= (*(arrvec[arrid[9]].ptr))[*]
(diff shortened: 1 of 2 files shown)


ksterne:
Good find on the bugfix! Sorry it was a little late before the holiday shutdown happened here at VT.

Just doing a visual inspection on this since not sure on how to disable the DLMs, it looks good.  I was going to have a question about what if the ifmode value is -1, but I see the sclid array is just an array of results from trying to find the variables in the record.  Mind if I add a few to high-level describe what's going on here so then it's clearer next time?  Or otherwise could you add them?

egthomas:
Sure, feel free to add some comments.  Unfortunately I don't have time to dive back into this at the moment.

ksterne:
Also looks like there may be another one at:

https://github.com/SuperDARN/rst/blob/03821dd7513edb96bcc1ccef57c7e8af52d83c1c/codebase/superdarn/src.idl/lib/main.1.25/fitcnx.pro#L198

Worth an extra PR or just direct edit to release branch?

egthomas:
Good catch!  I'm not sure that anyone uses that function but it certainly doesn't hurt to fix it.

ksterne:
I'm not familiar with it either, just did a quick visual scan of incrementing numbers in that directory and found that one.  I'll do at direct commit to the release branch unless someone objects.

Otherwise I've added comments here and looks like it compiles for me.  Are you good with the comments?  If so, I'll merge this in.

egthomas:
Maybe add the commit to this pull request so that they're somewhat grouped together, rather than directly to the release branch?  I think I can change the title of this pull request.

ksterne:
OK, I've pushed it here.  I'll merge it if the checks and my compilation are good.

pasha-ponomarenko:
Belated Season's Greetings!
;-)
I am not sure how many people use the native (non-DLM) IDL code, but it is always good to have it right.
Good job!

PR #122, opened 2018-01-09
Obiligatory master merge back into develop after 4.1 release

ksterne:
This is a capture of all of the extra features and bugfixes introduced into the `4.1/release` branch that was just merged into `master`.  This will get the `develop` branch up to speed with the `master` branch.

If no one else wants to hit the button, I will in an hour or so after I've gone and done a few other things.


pasha-ponomarenko:
So, now I can send out a reference to the develop branch accompanied by the release notes?

pasha-ponomarenko:
I think we may add Maria-Theresia as well.

ksterne:
Good e-mails @pasha-ponomarenko and @egthomas, though probably would have been good to do a bit more description on the new data products here (grid2, map2 standard, fitacf 3.0 is optional).  Or maybe we'll have to wait until SD '18 for a presentation at which point we're announcing RST 4.2?

pasha-ponomarenko:
Well, I attached a fitacf3_readme.pdf to the first message.

PR #123, opened 2018-01-15
Fixing error check in gridtogrdmap and maptocnvmap

egthomas:
This commit fixes a bug in the way `gridtogrdmap` and `maptocnvmap` check for errors which was added in the warning_cleanup branch prior to the 4.1 software release. The error check added to the end of these binaries looks for a non-zero return value from GridFwrite and CnvmapFwrite respectively, when it should actually check for values less than or equal to zero. The check for non-zero return values was only appropriate for the OldGridFwrite and OldCnvMapFwrite functions in `grdmaptogrid` and `cnvmaptomap` respectively. This commit makes the necessary changes so that `gridtogrdmap` and `maptocnvmap` do not always return a "GridFwrite/CnvmapFwrite failed." error message.

As they stand, `gridtogrdmap` and `maptocnvmap` are currently unusable.  Whether this warrants a fix directly to the master branch is left to the working group.

### DIFF
--- a/codebase/superdarn/src.bin/tk/reformat/gridtogrdmap.1.7/gridtogrdmap.c
+++ b/codebase/superdarn/src.bin/tk/reformat/gridtogrdmap.1.7/gridtogrdmap.c
@@ -100,7 +100,7 @@ int main (int argc,char *argv[]) {
               yr,mo,dy,hr,mt,(int) sc,grd->vcnum);
     }
     s=GridFwrite(stdout,grd);
-    if (s !=0) {
+    if (s <=0) {
         fprintf(stderr,"GridFwrite failed.\n");
         if (fp !=stdin) fclose(fp);
         exit(-1);
--- a/codebase/superdarn/src.bin/tk/reformat/maptocnvmap.1.6/maptocnvmap.c
+++ b/codebase/superdarn/src.bin/tk/reformat/maptocnvmap.1.6/maptocnvmap.c
@@ -104,7 +104,7 @@ int main (int argc,char *argv[]) {
               yr,mo,dy,hr,mt,(int) sc);
     }
     s=CnvMapFwrite(stdout,map,grd);
-    if (s !=0) {
+    if (s <=0) {
         fprintf(stderr,"CnvMapFwrite failed.\n");
         if (fp !=stdin) fclose(fp);
         exit(-1);


pasha-ponomarenko:
I think the hotfix is a viable option here.

ksterne:
Yea, agreed that this looks hotfix worthy as it looks like `gridfwrite` and `cnvmapwrite` both return 0 if they're good and -1 if something breaks (if that as it's an `exit()` call).  Go for it @egthomas. I don't think there's any testing necessary.

pasha-ponomarenko:
Ok then!

ksterne:
Or, a bit confused. If it is zero then it shouldn't go into the failed message...right?  Or shouldn't it be less than instead of less than or equal?

egthomas:
Gridwrite will return the value of either DataMapWrite or DataMapSize (depending on whether writing to a file or stdout I think):

https://github.com/SuperDARN/rst/blob/bugfix/error_check/codebase/superdarn/src.lib/tk/grid.1.24/src/writegrid.c#L293-L297

DataMapWrite will return -1 if it fails or the size of the record it writes if it succeeds:

https://github.com/SuperDARN/rst/blob/bugfix/error_check/codebase/general/src.lib/dmap.1.25/src/dmap.c#L866-L880

DataMapSize will always return the value of size of the record being considered (I think):

https://github.com/SuperDARN/rst/blob/bugfix/error_check/codebase/general/src.lib/dmap.1.25/src/dmap.c#L632

So I have it set to check for <=0 to catch the -1 if DataMapWrite fails and also 0 in case the size of the record is zero (which shouldn't happen).

egthomas:
I take that back, I think you're right that a zero returned by Gridwrite earlier in the function should not cause a failure message.

ksterne:
OK, sounds good. I was also confused because I was looking at a doc code at:

https://github.com/SuperDARN/rst/blob/bugfix/error_check/codebase/superdarn/src.lib/tk/grid.1.24/doc/src/GridFwrite/GridFwrite.c 

which is down in the doc directory and has a weird recursive call to itself.  But I think this is just a red herring...right?

egthomas:
Yeah, that file is just an (outdated) example.

egthomas:
I think we need to release a `4.1.1` version of the RST to restore this functionality to users (assuming my bug fix was correct).

pasha-ponomarenko:
Just arrived to Newcastle, NSW, to work with Colin Waters for a couple of weeks. Will try to have a closer look on Monday...

Issue #34, opened 2017-03-10
RadarLoadHardware return value


ksterne:
Looking through some old notes, I saw https://github.com/SuperDARN/rst/blob/94d90a09d0af48db34d0f1623a9bddfbcc05b143/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro#L102-L107  was changed.  I was under the understanding that RadarLoadHardware returned a zero if good and so the condition needed to be if it was not equal to zero.  However, in looking in the `codebase/superdarn/src.idl/lib/main.1.24/radar.pro` file, it seems as though the RadarLoadHardware routine will return zero if there's an IO error.  I'm not the best at IDL, so please correct me if I'm wrong.  

Otherwise, should we cleanup how this is noted or more of a comment than "SGS for DLM" on this line?  Or more so, can we correct the header documentation to this function as it notes that it returns a value of zero?

sshepherd:
This is an issue that needs to be sorted out later for the DLMs. The
problem is that the DLM version returns the opposite of what the C version
returns. I make the change myself when needed but haven't touched anything
in the DLMs, yet.

egthomas:
I think Simon may have meant the DLM and IDL versions return different values (rather than DLM and C).  Looking at the C code, RadarLoadHardware in radar.c of the radar.1.21 library will return -1 if an error occurs:

https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.lib/tk/radar.1.21/src/radar.c#L162-L163

This means that for a user with DLMs enabled, a call to RadarLoadHardware in IDL will return a -1 if an error occurs.  Otherwise if DLMs are not enabled, I'm not sure what RadarLoadHardware returns for an error since the iofail point at the end of the function isn't actually tied to an ON_IOERROR call.  Maybe 0 since a return value isn't set anywhere else?

egthomas:
Ok, so my previous post was not entirely accurate.  The IDL version of RadarLoadHardware will return a zero if everything works correctly:

https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L248-L300

However the DLM version of RadarLoadHardware will return the number of radars loaded if it works correctly:

https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.dlm/rposdlm.1.9/rposdlm.c#L385-L403

[...]

PR #125, opened 2018-01-31
Changing DLM version of RadarLoadHardware to return 0 on success to match native IDL function

egthomas:
This pull request addresses issue #34 originally raised last year by @ksterne.  The DLM and IDL versions of RadarLoadHardware now both return zero on success.  The relevant error checks in the IDL rbpos and make_grid routines have been corrected to reflect this change.

### DIFF
--- a/codebase/superdarn/src.dlm/rposdlm.1.9/rposdlm.c
+++ b/codebase/superdarn/src.dlm/rposdlm.1.9/rposdlm.c
@@ -400,7 +400,7 @@ static IDL_VPTR IDLRadarLoadHardware(int argc,IDL_VPTR *argv,char *argk) {
 
   IDL_KWCleanup(IDL_KW_CLEAN);
 
-  return (IDL_GettmpLong(s));
+  return (IDL_GettmpLong(0));
 
 }
 
--- a/codebase/superdarn/src.idl/app/make_grid.1.0/make_grid.pro
+++ b/codebase/superdarn/src.idl/app/make_grid.1.0/make_grid.pro
@@ -460,7 +460,7 @@ endif
 
 ; Load the hardware information for the radar network
 s = RadarLoadHardware(network, path=envstr)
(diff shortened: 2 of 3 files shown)


ksterne:
Hey @egthomas, is there any kind of test you could throw our way?  I'm struggling to remember if there was one that I was looking at with #34.  Likely not since I'm not too familiar with IDL/DLMs.  I'm guessing that you've found no value in knowing how many radars were loaded in with https://github.com/SuperDARN/rst/blob/82b4ac03873b521b4de51f7ad20a961e92a544ff/codebase/superdarn/src.dlm/rposdlm.1.9/rposdlm.c#L92-L157

or there's no where else that seems dependent on r/s being returned instead of zero?

egthomas:
All of the IDL code I've found which calls RadarLoadHardware expects a zero upon success (e.g., beamread in fitlib.pro and rbpos in rbposlib.pro).  You should be able to compare the DLM and IDL output by temporarily renaming the RadarLoadHardware function in radar.pro to something like RadarLoadHardware_IDL.  That way, a call like `s=RadarLoadHardware(network,path=getenv('SD_HDWPATH'))` will use the DLM while `s=RadarLoadHardware_IDL(network,path=getenv('SD_HDWPATH'))` will use the native IDL code, and you can compare the return value of the two sets of code with/without the change in this pull request.

egthomas:
So it's been over a month since the telecon and three months since this pull request was opened.  The changes here to three lines of IDL code are much more benign than some of the other open pull requests (eg adding the TS18 statistical model).  Any objections to merging this?

sshepherd:
No objection.

pasha-ponomarenko:
No worries!

ksterne:
Sorry I wasn't able to get to this as maybe promised.  Several radar trips plus writing a report have taken up much of my time.  I'll pull it locally and see if we've got any problems here.

PR #129, opened 2018-02-09
4.1.1 patch to fix gridtogrdmap and maptocnvmap binaries

egthomas:
As previously suggested in #123 I think we need to create a patched version of v4.1 which includes the fix to the gridtogrdmap and maptocnvmap binaries.  Therefore this pull request is to merge the current develop branch into master.  I've also drafted a new release tag:

https://github.com/SuperDARN/rst/releases/tag/untagged-b6fea7859e14184f0a63


ksterne:
So what was the feeling here again?  Were we thinking we'd more pull this into develop and do a release in a couple of months or keep going with this one as a hotfix into master?

ksterne:
Sorry, I mean should we just keep this in develop or continue with this pull request?  I say this because I feel like we should take care of this pull request before the others especially since there are a few commits that are now included in the `develop` branch.

pasha-ponomarenko:
I believe that the consensus at the telecon was that it does not seem to be that urgent because noone except Evan has complained yet about problems with  gridtogrdmap and maptocnvmap. On the other hand, this request has snowballed somehow absorbing several other things during its lifespan. 
So, I have no preference either way.

My efforts to generate the minutes of the telecon in a timely manner have been somewhat compromised by an injured right pinky which is heavily bandaged now...
:-(

egthomas:
Yes, I agree with Pasha's recollection - we can either close this pull request or leave it open for the next eventual release (be it 4.1.1, 4.2, etc).

egthomas:
I'm actually in favor of closing this pull request so we don't continue to get spammed with emails every time a commit is added to the develop branch.  yay/nay?

pasha-ponomarenko:
As I said before, I am OK either way. It is more of a git technical nature so it's up to Kevin/Keith(?).

ksterne:
I'm in favor of closing this as the original issue doesn't seem to be "hot" enough for a hotfix to the master branch and this pull request either keeps including additional commits or discourages any new pull requests into the `develop` branch.  I'll let it to @egthomas to close it since you opened it up.

egthomas:
Closing this for now due to email spam.

PR #130, opened 2018-02-09
Adding old meteor_proc software from BAS

egthomas:
This pull request adds the meteor wind processing software (meteor_proc) which originated at British Antarctic Survey (BAS) and was distributed in older versions of the RST.  The following changes have been made to the 1.10 version of the software I started with:

- Added functionality to read fitacf format files (previously could only process fit or cfit files)
- Increased maximum size of some arrays from 16 to 24 elements to support radars with more than 16 beams
- Added Chisham virtual height model functionality
- Updated documentation (html and command line)

The underlying algorithms used for calculating the meteor wind velocity components remain unchanged.  The only other refurbishment the code needs are the default meridional beam values for the newer radars (ie those constructed since the early 2000s) set in `meteor_proc.c`.


kkotyk:
Was there a reason this stopped being included? Does this piece of software need to be tested as if it were a new algorithm, or has it been used and accepted by the community in the past?

egthomas:
I'm not sure why it stopped being included.  The algorithm was used for several published papers and the current and former BAS PIs (Mervyn and Gareth) both said it should be returned to the RST.

pasha-ponomarenko:
Should we ask Gareth to test it?

aburrell:
I did ask him and he said yes.  Someone else should probably ping him at this point.

Cheers,
Angeline

egthomas:
This branch now compiles as of version 4.2 and is in the hands of Gareth or whoever else to test/approve.

ecbland:
Update: Gareth has indicated that the `meteor_proc` software is NOT ready to be put back into RST due to the discovery of some new issues, so I'm closing this PR.

Issue #131, opened 2018-02-12
20070117 make_grid errors


mts299:
Hi, 

I am a new programmer/ data analyst at SuperDARN Canada, and I am currently working with the RST code to generate convection maps.

I noticed for the data set 20070117.C0.rkn.fitacf returns an error message "Error mapping beams". Any insight on why it would return this message? dmapdump shows there is data in the file. 
Also, the return value for this error is 0, do we not want something more meaningful? 

Another problem is reading in 20070117.C0.han.fitacf: 
~> make_grid -xtd -cn A -i 120 -minrng 2 -vemax 1000000 20070117.C0.han.fitacf > 20070117.han.A.grid
Error reading file.

Which I assume is because there are scan flags = -1 so when I change the options to deal with it as instructed in issue #85 , I get a segfault instead:
~> make_grid -xtd -cn A -tl 100 -i 120 -minrng 2 -vemax 1000000 20070117.C0.han.fitacf > 20070117.han.A.grid
Segmentation fault (core dumped)

Cheers,
Marina

egthomas:
Hi Marina - The "error mapping beams" message appears when the GridTableMap function called by make_grid returns -1 instead of 0:

https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.bin/tk/tool/make_grid.2.0/make_grid.c#L869
https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/gtable.2.0/src/gtable.c#L555

Unfortunately my memory is hazy and the comments I added don't explain why this might happen.  Probably failures in GridTableFindBeam or GridTableAddBeam.  I totally agree the error message should be more informative.

When using the `-tl` and `-i` flags with make_grid I would probably recommend setting each of those to the same value (rather than 100 s for one and 120 s for the other).  If that's not causing the seg fault, could you try setting the `-vb` verbose flag to see which record the software is breaking on?  That way we might be able to isolate whichever operating parameter or data value is causing the issue.

mts299:
Hi Evan, 

Thanks for the quick reply. I will take a look at the code and let you guys know if I find anything. 

Sorry, that was a typo. Both options are set at 120, and I still get a segfault. 
With verbose I get output that looks like this: 
Storing:2007-1-17 13:46:0 13:48:0 pnts=0 (10)
Storing:2007-1-17 13:48:0 13:50:0 pnts=0 (10)
[...]

PR #132, opened 2018-02-24
Check absolute value of scan flag in various C routines

egthomas:
As originally identified by @mts299 in #131, this pull request modifies several binaries and libraries to check against the absolute value of the `prm->scan` flag for values of 1 to handle cases where the scan flag has been set to -1 (for backwards scanning sequences) rather than the usual value of 1. This should fix the behavior where make_grid never finds a scan flag of 1 and eventually 1000 beam soundings are read, causing the code to stop and display an error message.


egthomas:
@mts299 - did this solve your issues in #131 (other than the problems with the blacklisted files)?

mts299:
@egthomas - Yes it did. 
 Thank you!

egthomas:
Well, it doesn't seem like the 4.1.1 patch will go into effect any time soon so feel free to merge this pull request.

PR #133, opened 2018-02-24
Fix bugs in IDL version of make_grid

egthomas:
This pull request fixes several bugs identified by @billetd within the IDL version of make_grid which prevent the software from executing.  These bugs were my fault when porting the IDL code from my local machine to the RST.  The negative scan flag issue corrected in the C code in #132 is also fixed here.

### DIFF
--- a/codebase/superdarn/src.idl/app/make_grid.1.0/fitscan.pro
+++ b/codebase/superdarn/src.idl/app/make_grid.1.0/fitscan.pro
@@ -181,7 +181,7 @@ function FitReadRadarScan, unit, state, RadarScan, prm, fit, tlen, lock, channel
             if tlen eq 0 then begin
                 ; Set flg equal to one if scan data according to scan flag
                 ; was successfully stored in RadarScan structure
-                if prm.scan eq 1 then $
+                if abs(prm.scan) eq 1 then $
                     flg = 1
             endif else begin
                 ; Set flg equal to one if scan data of length tlen was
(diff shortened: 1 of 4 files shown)


egthomas:
@billetd - any other changes you think should be made here?  I know you made a few other changes to the timing logic of make_grid in your local code.

billetd:
The only other change I've experimented with is replacing line 469 of gtable.c with:
`ptr->st_time=scan->st_time;`
This seems to make the start times behave better when using the `-i` flag for custom integration times and has no effect to standard behaviour. I can't comment on whether this breaks anything else though, although all it seems to prevent is the original code from rounding up the start time to the next scan.

egthomas:
@billetd - what are your thoughts on approving this pull request?  It sounds like you've been using the proposed changes for quite some time without any issues.

At a minimum the fixes to the IDL make_grid code are absolutely necessary for it to work correctly.

billetd:
As you've said, I've currently ran into no issues using this. When -i and -tl are not used, the standard two minute records are still created as normal with no obvious changes, so I'm happy for this to go live if nobody else has found a problem.

ksterne:
@billetd, if you're our best tester and can say this works...don't be shy to hit that merge button!

billetd:
@ksterne, I would if I had write access!

egthomas:
@billetd sorry, didn't realize you weren't a "developer" member - check your inbox, an invite should be en route.

ksterne:
Seems as though @egthomas is the only one who can correct that.  @egthomas, maybe worth adding a few more of us to maintainers for the group if that's possible?

PR #136, opened 2018-03-08
Updating Falkland Islands hdw.dat and radar.dat files

egthomas:
This pull request replaces the old Falkland Islands radar (FIR) hdw.dat file with the new one distributed by Gareth Chisham over the darn-users listserv.  The `radar.dat` file has also been updated to change the FIR radar status flag from -1 to 1.

For some reason there are a bunch of `^M` return characters at the end of the each line in a lot of the hdw.dat files so git thinks the entire file has changed when really it's only the last few lines.


ksterne:
Should note the location and altitude changed ever so slightly here on the first line.  Probably doesn't change things much, but going to recommend a complete data reprocess to DDWG.

ksterne:
Also changed scanning boresight by 0.1 degree.

ksterne:
OK, for real last comment (sorry for the spam on e-mail subscribes), also increased number of ranges on first line.

PR #137, opened 2018-03-09
Fixing more compiler warnings and errors (dmap, zlib, etc)

egthomas:
This pull request reduces the total number of compiler warning messages on the develop branch from 285 to ~~191~~ 150 by fixing numerous typos and pointer issues, particularly within the dmap library and functions using zlib.  This pull request also clears up 87 `note: expected ... but argument is of type ...` messages associated with some of those compiler warnings.

Of the ~~191~~ 150 remaining compiler warnings, here is the breakdown:

- 83: variable ... set but not used
- 33: ignoring return value of ...
- 24: ... defined but not used
- ~~12~~ ~~3: ... may be used uninitialized in this function~~ (fixed in #149)
- ~~12: unused variable ...~~
- 10: dereferencing type-punned pointer will break strict aliasing rules
- ~~10: format '%X' expects argument of type 'Y', but argument Z has type ...~~
-  ~~4: ... declared inside parameter list~~
-  ~~2: its scope is only this definition or declaration, which is probably not what you want~~

### DIFF
--- a/codebase/general/src.lib/dmap.1.25/include/dmap.h
+++ b/codebase/general/src.lib/dmap.1.25/include/dmap.h
@@ -158,7 +158,7 @@ int DataMapSetFreeArray(struct DataMap *ptr,char *name,int type,int dim);
 
 int DataMapSize(struct DataMap *write);
 
-char *DataMapEncodeBuffer(struct DataMap *ptr,int *size);
+unsigned char *DataMapEncodeBuffer(struct DataMap *ptr,int *size);
 
 int DataMapWrite(int fid,struct DataMap *ptr);
  
@@ -170,7 +170,7 @@ int DataMapFwrite(FILE *fp,struct DataMap *ptr);
 
 struct DataMap *DataMapFread(FILE *fp);
 
-struct DataMap *DataMapDecodeBuffer(char *buf,int size);
+struct DataMap *DataMapDecodeBuffer(unsigned char *buf,int size);
 
 struct DataMap *DataMapReadBlock(int fid,int *s);
 struct DataMap *DataMapFreadBlock(FILE *fp,int *s);
--- a/codebase/general/src.lib/dmap.1.25/include/dmap.h
+++ b/codebase/general/src.lib/dmap.1.25/include/dmap.h
@@ -58,7 +58,7 @@ struct DataMapFp {
   int size;
   union {
     FILE *f;
-    gzFile *z;
+    gzFile z;
   } fp;
 };
 
--- a/codebase/general/src.lib/dmap.1.25/include/dmap.h
+++ b/codebase/general/src.lib/dmap.1.25/include/dmap.h
@@ -151,7 +151,7 @@ void *DataMapStoreArray(struct DataMap *ptr,
 int DataMapRemoveArray(struct DataMap *ptr,char *name,int type,int dim);
 
 void *DataMapFindArray(struct DataMap *ptr,char *name,int type,int dim,
-		       int **rng);
+		       int32 **rng);
 
 
 int DataMapSetFreeArray(struct DataMap *ptr,char *name,int type,int dim);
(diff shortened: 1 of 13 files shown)


pasha-ponomarenko:
Usual no-crash compiling test here?

egthomas:
@pasha-ponomarenko The Travis CI integration should automatically handle the compilation testing for us.  You can verify this by clicking on the "Show all checks" link next to the bold "All checks have passed" text and then click on the details of the commit and/or pull request build tests to see the compilation log (e.g., https://travis-ci.org/SuperDARN/rst/builds/351707185).

This is also my final RST commit / pull request for the foreseeable future.  Of the remaining 165 compiler warnings identified in my edited post above, most are pretty benign (variables set or defined but not used, etc), however the "dereferencing type-punned pointer" warnings are outside my wheelhouse and the remaining 3 "may be used uninitialized" warnings are in @sshepherd's AACGM_v2 code.  If someone else wants to take a crack at some of the remaining errors please feel free.  Otherwise there are plenty of other open pull requests that could use testing/approval.

egthomas:
Ok, that was my final commit (previously identified and fixed in davitpy by @asreimer and @jspaleta in https://github.com/vtsuperdarn/davitpy/pull/40).

codebase/general/src.lib/dmap.1.25/include/dmap.h
```diff
@@ -151,7 +151,7 @@ void *DataMapStoreArray(struct DataMap *ptr,
 int DataMapRemoveArray(struct DataMap *ptr,char *name,int type,int dim);
 
 void *DataMapFindArray(struct DataMap *ptr,char *name,int type,int dim,
-		       int **rng);
+		       int32 **rng);
```
kkotyk: What's the reason this particular variable needs be labeled int32?

egthomas: Because that's how rng is declared in DataMapFindArray in `dmap.c`:

https://github.com/SuperDARN/rst/blob/bfa8cb951ddf93126d3a244cbedccdbaac045f9f/codebase/general/src.lib/dmap.1.25/src/dmap.c#L636-L637

And earlier in `dmap.h`:

https://github.com/SuperDARN/rst/blob/bfa8cb951ddf93126d3a244cbedccdbaac045f9f/codebase/general/src.lib/dmap.1.25/include/dmap.h#L93-L100

As noted in https://github.com/vtsuperdarn/davitpy/pull/40, DataMapFindArray isn't actually used anywhere in the RST but it has been cited as causing compilation problems in other situations.

egthomas: Well, looking at the DataMapArray struct, dim is also labeled int32, so maybe that's not the best justification.  More importantly I think, all other functions in dmap.c and dmap.h label rng as int32 rather than int.

kkotyk: I would agree. If you ever want to break your brain, look at rtypes.h. It actually does make sure that variables are all the same size on any platform, but I think it was written before the std types lib was available. I would love if we could get rid of that file.

egthomas: As a C dummy, what would prevent us from doing so?  time/effort?  Or is it a compatibility issue with the ROS side of things?

kkotyk: I dont think it would be too hard to do in RST. You could easily have a script that would search and replace int, short, char with int32,int16,int8 or whatever. As long as we make sure that rtypes define the variables to be the same sizes as those in the std types lib, I think it would work and be compatible?

kkotyk: I agree, and yes this look fine to me. There is more we need to consider for a future PR on changing types.

egthomas:
For a while there I was excited that something might get approved.  What a tease!

ksterne:
Two radar trips in three weeks has gotten in my way.  Try for tomorrow or more likely next week after debriefing from major antenna work in Wallops. Anyone else feel free though.



--------------------
....thumbed from my phone; auto-correct may have improved this message...

----- Reply message -----

PR #138, opened 2018-03-22
Adding the TS18 statistical model and making it the default in map_addmodel

egthomas:
As the title suggests, this pull request adds the recently accepted Thomas and Shepherd [2018] (TS18) statistical model coefficients to the tables/superdarn/model directory and replaces CS10 as the default model used by map_addmodel.  I believe the DOI will be 10.1002/2018JA025280 but can list the correct one if it changes during the publication process.


ksterne:
Any figures that I could compare against that show differences between CS10 and TS18?

egthomas:
Why yes, actually:

http://superdarn.thayer.dartmouth.edu/WG-sched/models_dial.html

At some point this will become a publicly accessible page on our website so the address might change.

ksterne:
Hey Evan, the website's pretty neat!  Is there anything with running code/binaries from this repo that can be shown?  I guess otherwise, just take a map file and put the magnitude into the website to see that they match?

egthomas:
Yes, this can be compared to those figures using the `-empty` option with map_grd and the `-nointerp` option with map_addmodel. Otherwise, this code has been running in real-time using DSCVR data as the solar wind input since early December on the Dartmouth SD home page.

egthomas:
This second (and hopefully final) pair of commits adds the alternative TS18-Kp model, which as the name implies is a statistical model parameterized by the Kp index (and IMF clock angle) rather than Btot or Esw.  To add this functionality I had to modify the map data structure and cnvmap file format to add an additional Kp field.  map_addimf should also now accept Kp as either a command line input using the `-kp` flag or as another column on the custom input text file.  The Kp model does not have any interpolation applied between model bins and is generally not recommended to be used over the standard TS18 model.  But we thought it would be best to get the coefficients and functionality into the RST so they don't potentially disappear into the ether like the RG05 model coefficients.

egthomas:
Here's the link to the article on the JGR website:

https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1002/2018JA025280

egthomas:
In the process of writing a more user-friendly statistical model solver, a bug was identified in `map_addmodel` which causes incorrect selection of the dipole tilt bin for the PSR10, CS10, and TS18 models.  This bug is also present in the current version of `map_addmodel` in RST 4.1.

ksterne:
@egthomas, seems as though you got to this before I did.  Though since you know it better you were probably able to work through it quicker.  I had been looking at an issue that was popping up in `map_addmodel` off and on over the last few weeks.  It seems as though f5e0a03 corrects the issue I was seeing (seg faulting on random days).  Maybe worth elevating this commit to a hotfix into the current master?

egthomas:
Hm, that sounds like it may be unrelated to yesterday's.  The bug I corrected was causing periods with dipole tilt < 0 to be identified as negative, when really it should have been `tilt < -10 = negative`, `-10 < tilt < 10 = neutral`, and `10 < tilt = positive`.

Can you provide any more information about when `map_addmodel` is causing seg faults for you?

ksterne:
If it's too much derailing, I can kick this over to an issue.  This does seem to fix it at least for one date.

So here's the commands I'm using where I find a segfault on the current `develop` branch:

- `map_grd 20171002.north.grid2 > step1` where the grid2 file has been generated using the current `develop` branch
- `map_addhmb step1 > step2`
- `map_addimf -if 20171002_omniex.asc -d 0 step2 > step3` where the 20171002_omniex.asc file is the 1 minute OMNI IMF data and then split and parsed by DaViT code
- `map_addmodel -o 8 -d l step3 > step4` On develop this results in segfault, but after f5e0a03 I'm not getting a segault.

I did some debugging with print statements to try to figure out where the segfault was coming from and the best I could nail down it was somewhere in:

https://github.com/SuperDARN/rst/blob/81fdd658157e8d7c599278d6892c3f15f5c59cd3/codebase/superdarn/src.bin/tk/tool/map_addmodel.1.14/map_addmodel.c#L747-L787

and more so in the 'else' section of that nested for loop.  The very last I was trying to figure out was if one of the model variables hadn't been set yet.  Looking at it now though, I'm wondering if denom was set to 0 or something too.  Sound familiar to what you found?

egthomas:
@ksterne I think the issue of `map_addmodel` causing seg faults should now be fixed.  It occured when trying to apply trilinear interpolation to the CS10 model coefficients when the solar wind conditions are in a By+ and Esw>4.1 bin (which exists), but the adjacent Bz-/By+ bin for Esw>4.1 doesn't exist.

egthomas:
@ksterne - I tested the other two dates you mentioned (20171117 and 20171120) and those will no longer seg fault on this branch when using the CS10 model.  Note that this branch replaces CS10 with TS18 as the default statistical model, so you need to include the `-cs10` option when calling `map_addmodel` to verify.

egthomas:
Adapting this from a comment in #100:

```
map_grd -empty -sd 20170101 > 20170101.north.empty.cnvmap
map_addhmb -nodef 20170101.north.empty.cnvmap > 20170101.north.hmb.cnvmap
map_addimf -if imf.txt -d 0 20170101.north.hmb.cnvmap > 20170101.north.imf.cnvmap
map_addmodel -ts18 -o 8 -nointerp 20170101.north.imf.cnvmap > 20170101.north.model.ts18.cnvmap
map_fit 20170101.north.model.ts18.cnvmap > 20170101.north.ts18.cnvmap
```

The above commands can be used to create an empty map file with something like the attached 
[imf.txt](https://github.com/SuperDARN/rst/files/2039340/imf.txt) file for adding fake IMF data which cycles through a variety of magnitude and clock angle bins.  I've also attached a text file called [myoptions.txt](https://github.com/SuperDARN/rst/files/2039350/myoptions.txt) for use with `map_plot` which contains useful plotting/graphics options for viewing the Map Potential results:

`map_plot -x -cf myoptions.txt 20170101.north.ts18.cnvmap`

The results can be verified by comparing to the static patterns in the supplementary info of the paper or on the Dartmouth website:  http://superdarn.thayer.dartmouth.edu/convection.html

A simpler way to test this may be to look at the convection model solver first (#159), since it will provide simple MLAT/MLT/potential output for a single set of solar wind/geomagnetic conditions in plain-text format which should be easy to contour in the programming language of choice.  That pull request actually needs to be approved first since it was based on this branch (`feature/ts18_model`).

ksterne:
Thanks for the great example here.  That's what I was looking for since I'm not up on this as much as you are.  I did have to add `-delay 0` to the `map_plot` options as they were wheeling by too fast.

Throwing this at some actual data, I've getting:

image

image

image

See any issues?  Looks pretty good to me.

So wait, you don't want this merged in until #159 is merged in?  I guess we'll wait until that's merged before merging this in then.

egthomas:
Those instantaneous Map Potential patterns look reasonable when compared to the underlying model bins being used (whether the data should reflect the model or the model should reflect the data is another question entirely).  Yes, I would like to see the model solver merged into this branch first because it depends on the TS18 functionality added by this branch.  Also note that there is new Kp index functionality added to `map_addimf` and the map file format to support the TS18-Kp model.

egthomas:
Similar to the new `solve_model` binary introduced in #159, I've added a check for missing/undefined PSR10 model patterns to `map_addmodel`.  Even so I would discourage trying to use the PSR10 model for operational purposes due to the limited input data coverage and missing model patterns.

ksterne:
I think this looks good here.  Thanks for adding in the `map_addmodel` check as well.  Though that has shed light on the fact #159 should have been reconfigured to create a library for at least the `determine_model` function as now there are duplicate copies of that code in this branch.  At least I think I have that analyzed right; @egthomas, does that seem right?

I bring it up because if a modification is to happen in the future, then both files will need to be modified if that's true.  Maybe a library isn't the right lingo/type, but it seems as though `map_addmodel` and `solve_model` could use the same code.

ksterne:
I tried a couple of days and combinations, but it doesn't seem like I could trigger the psr10 model to go undefined as easily as I could in #159.  So I think this is good to go aside from what I mentioned before with the duplicate `determine_model` functions.  Let me know if that doesn't seem right.

I can completely understand if you'd rather proceed with this merging and I can create an issue noting the duplicated code (again, if that's a correct understanding).  It's not a problem with functionality, but more so of keeping track of things in the future.

egthomas:
@ksterne that can be addressed in a future pull request.  Consolidating duplicate functions in the RST into a separate library was only just recently done for the `addhmb` library when I had more time to work on this.  There are still dozens of other places (particularly in the plotting code) where duplicate functions exist in different directories.

kkotyk:
It appears you guys were waiting for #159 to be accepted. It appears that has happened. Are you performing any more testing @ksterne. As a side note, I think this level of testing and staying focused here is excellent, and would like to see more of this in other PRs.

sshepherd:
Let's get this approved. Another month has gone by with no activity...

ksterne:
My apologies for being slow here.  The last couple of weeks have been filled with hosting a student, vacation, and catching up with my other half.  Thank you for being patient with me.  I think the last comment you made @egthomas is good and understand that there's probably a bigger effort/job to eliminate the duplicated code in other places.  I'll create the issue to capture the knowledge gained here.

Think it's good to go, so going to hit merge here!

egthomas:
Thanks @ksterne!

Issue #116, opened 2017-12-07
test_raw and dattorawacf buffer overflow with 1993-1995 dat files


egthomas:
I don't have much time to look into this with AGU approaching but I'm consistently encountering buffer overflows when trying to process early (~1993-1995) dat files using the `test_raw` and `dattorawacf` binaries.  I think it might be happening in `OldRawRead` of the `oldraw.1.16` library although am not entirely sure.  Is anyone else experiencing these issues?

A few of the files I've tested include (obtained from the BAS data mirror):

1993093023t.dat
1993101221g.dat
1993101221k.dat
1993101221t.dat
1994010100g.dat
1994010100k.dat
1994010100t.dat
1995010100k.dat
1995010100w.dat

pasha-ponomarenko:
I did recall a problem related to reading/converting dat files in the past, but it looked more like a segfault. Is it only for the 1993-1995 data?

egthomas:
The half dozen dat files I tried from the year 2000 worked without any buffer overflow errors.  Beyond that I haven't narrowed down a specific problem interval.  It seems like the routines make it through all of the records in the file but crash on the next iteration of the loop (ie when it should hit a null or something and exit).

egthomas:
I think I've tracked this down to the `int rlen` parameter in the OldRawFp structure:

https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/oldraw.1.16/include/oldrawread.h#L45

Sometimes OldRawRead appears to figure out that the record length at the end of the file is zero and not crash (post-1995 .dat files?) but other times it doesn't realize the file is about to end and you encounter a buffer overflow (~1993-1995 .dat files).  I tested this by using the IDL OldRawOpen and OldRawRead routines (without DLMs) and checking the rechdr.size (or rawfp.rlen) values to see if they are zero.  Sure enough, the 1993-95 files appear to have "empty" records at the end of the .dat files - the file pointer doesn't seem to return a null but the record header information is all zero / missing so it crashes when trying to access the full set of `(struct OldRawFp *ptr,struct RadarParm *,struct RawData *)` here:

https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/oldraw.1.16/src/rawread.c#L49

[...]

PR #139, opened 2018-03-28
Check for zero-length records in dat files

egthomas:
This pull request addresses issue #116 where early dat files caused buffer overflows due to zero-length records at the end of the files.  Fixes are made to `OldRawRead` in the IDL `oldraw.pro` file and `OldRawReadCurrent` in the C oldraw library.  As noted in #116, we may also need to add similar checks to the `cmpraw` binary and `OldRawOpenFd` in the oldraw library.

Strangely, the rawfp structures returned by the native IDL and DLM versions of OldRawRead do not match.  Not sure that it's worth focusing any time/resources on addressing but thought it should at least be pointed out.

### DIFF
--- a/codebase/superdarn/src.bin/tk/testing/cmpraw.1.6/raw_open.c
+++ b/codebase/superdarn/src.bin/tk/testing/cmpraw.1.6/raw_open.c
@@ -75,7 +75,7 @@ struct rawfp *raw_open(char *rawfile,char *inxfile) {
   }
   fstat(ptr->rawfp,&ptr->rstat);
 
-  if (ConvertReadShort(ptr->rawfp,&num_byte) !=0) {
+  if (ConvertReadShort(ptr->rawfp,&num_byte) !=0 || num_byte <= 0) {
     close(ptr->rawfp);
     free(ptr);
     free(inbuf);
@@ -105,7 +105,7 @@ struct rawfp *raw_open(char *rawfile,char *inxfile) {
      the file */
 
   
- if (ConvertReadShort(ptr->rawfp,&num_byte) !=0) {
+ if (ConvertReadShort(ptr->rawfp,&num_byte) !=0 || num_byte <= 0) {
     close(ptr->rawfp);
     free(ptr);
     free(inbuf);
--- a/codebase/superdarn/src.bin/tk/testing/cmpraw.1.6/raw_read_current.c
+++ b/codebase/superdarn/src.bin/tk/testing/cmpraw.1.6/raw_read_current.c
@@ -85,7 +85,7 @@ int raw_read_current(struct rawfp *fp,struct rawdata *raw_data) {
    memset(raw_data,0,sizeof(struct rawdata));
    fp->rlen=0;
    do {
-     if (ConvertReadShort(fp->rawfp,&num_byte) !=0) {
+     if (ConvertReadShort(fp->rawfp,&num_byte) !=0 || num_byte <= 0) {
        free(inbuf);
        return -1;
      }
--- a/codebase/superdarn/src.idl/lib/main.1.25/oldraw.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/oldraw.pro
@@ -438,7 +438,7 @@ ON_IOERROR,iofail
   readu,rawfp.rawunit,rechdr
 
   ; Make sure the record size is non-zero
-  if rechdr.size eq 0 then return,-1
+  if rechdr.size le 0 then return,-1
 
   num_byte=rechdr.size
   rawfp.rlen=rechdr.size
--- a/codebase/superdarn/src.idl/lib/main.1.25/oldraw.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/oldraw.pro
@@ -436,7 +436,10 @@ ON_IOERROR,iofail
   OldRawMakeOldRawRadarParm,rparm
 
   readu,rawfp.rawunit,rechdr
- 
+
+  ; Make sure the record size is non-zero
+  if rechdr.size eq 0 then return,-1
+
   num_byte=rechdr.size
   rawfp.rlen=rechdr.size
   rawfp.ptr+=rechdr.size
--- a/codebase/superdarn/src.lib/tk/oldraw.1.16/src/rawopen.c
+++ b/codebase/superdarn/src.lib/tk/oldraw.1.16/src/rawopen.c
@@ -84,7 +84,7 @@ struct OldRawFp *OldRawOpenFd(int rawfd,int inxfd) {
 
   fstat(ptr->rawfp,&ptr->rstat);
 
-  if (ConvertReadShort(ptr->rawfp,&num_byte) !=0) {
+  if (ConvertReadShort(ptr->rawfp,&num_byte) !=0 || num_byte <= 0) {
     close(ptr->rawfp);
     free(ptr);
     free(inbuf);
@@ -140,7 +140,7 @@ struct OldRawFp *OldRawOpenFd(int rawfd,int inxfd) {
      the file */
 
   
- if (ConvertReadShort(ptr->rawfp,&num_byte) !=0) {
+  if (ConvertReadShort(ptr->rawfp,&num_byte) !=0 || num_byte <= 0) {
     close(ptr->rawfp);
     free(ptr);
     free(inbuf);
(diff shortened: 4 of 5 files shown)


ksterne:
hey @egthomas, not a rush on this so get to it when you can.  I noticed a similar issue with the num_byte check.  I should have probably checked here first...but at least I'm confident in that part of this pull request now.  Do you have any examples you might give on getting to the IDL portion here?  Otherwise, I can try digging through and understanding that file change as well (maybe it's simple enough).

egthomas:
@ksterne sorry, I'm not sure what you're asking me for.  The C and IDL changes are both simply to allow opening/reading of early `dat` files without encountering a seg fault.  You can bypass the DLMs to test any IDL code by renaming the routine or function (e.g., `function OldRawRead,rawfp,prm,raw` -> `function OldRawRead_test,rawfp,prm,raw`).

ksterne:
Sounds good, I realized I don't have a way to test the IDL portion anyways (I think) since I don't have IDL setup on my VM.  Is it worth putting some kind of error message in the if statements here to be a little bit more informative?  For now when I run it on my test file, the only info I get back is `File not found.` which makes it seems like I have a bad input filename, not that the file is actually bad/corrupted.  I can add in these prints to stderr if you'd like since I'm asking extra.

pasha-ponomarenko:
I can help with the IDL part.

egthomas:
@ksterne what test file / commands are you using?  I don't think you should see a `File not found` message on either the develop or this branch (at least related to the issue being addressed).

ksterne:
I'm seeing `File not found.` on 2001102410w.dat when running `make_fit -old 2001102410w.dat w_fit`.  I encountered this while running a big reprocessing script related to #162.  There's a bunch of other dat files that are failing/seg faulting that ultimately stopped the script.  I tested two other files though 199901010[02]w.dat and both of those seg fault on the `develop` branch but pass here.  Though not sure on the data quality but getting over 2.5 MB files, so they appear to be OK.

I feel like getting the `File not found.` message seems right though as the oldraw library will return NULL at https://github.com/SuperDARN/rst/blob/049d0a487cabe1b18754dd3c6b7188b5c578488d/codebase/superdarn/src.lib/tk/oldraw.1.16/src/rawopen.c#L87-L92 
which then returns back to
https://github.com/SuperDARN/rst/blob/049d0a487cabe1b18754dd3c6b7188b5c578488d/codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c#L180-L185
which will print that error message and then exit.  Maybe it is something with the 2001 file, but I think this should happen with the other two if the num_byte condition is being met.  I haven't confirmed that they are meeting this condition though and maybe it's happening elsewhere.

egthomas:
@ksterne this pull request is meant to address the specific issue described in #116 where all early (~1993-1995) dat files caused a buffer overflow when trying to read or convert them.  It sounds like those two 1999 dat files you identified suffered from this same issue where the number of bytes was not correctly set in the header for the final record in the file.  If other isolated dat files from later dates are still causing errors then they are outside the scope of this pull request.  

For your particular example, the value of `num_byte` in the first record of `2001102410w.dat` is set to -29921 which is causing the initial call to `OldRawOpen` in `make_fit` to return a NULL and print the `File not found.` error message.  So a specific file corruption issue and not meant to be addressed by this pull request.

ksterne:
I certainly wasn't trying to cover too much here.  After looking at the two 1999 files, I figured there's another problem with the 2001 file that I agree is outside the scope of this request.  But what is the value of `num_byte` for your test cases?  The condition in `rawopen.c` that I've highlighted should still be met.  Do you think it's outside the scope of this PR to include a print message noting an issue with reading the number of bytes?

egthomas:
The 1993-1995 (and your two 1999 files) fail at the very end of the file within `OldRawRead` (ie later in `make_fit`).  Your 2001 file fails on the very first record within `OldRawOpen`.  That's why the one returns a `File not found` error message at line 183 while the others don't.

pasha-ponomarenko:
OK, the native IDL routines are reading all the data records in 
1994010100g.dat
1994010100k.dat
1994010100t.dat

On a side note, I motice that the date is 1996/12/31 instead of 1994/01/01, and this happens in both native IDL and DLM outputs. Yet another issue to resolve...

pasha-ponomarenko:
The remaining files form the initial test pool,
1993093023t.dat
1993101221g.dat
1993101221k.dat
1993101221t.dat
1995010100k.dat
1995010100w.dat
produce the expected dates.

egthomas:
@pasha-ponomarenko thanks for testing those files in IDL.  Is this ready to be merged then?

ksterne:
@egthomas, I'd hit merge, but I was wondering what you thought on adding an error message here.  I was thinking something along the lines of `fprintf(stderr, "_libraryname_:_functionname_:Error reading record\n");`  That way if a file starts having errors, its at least apparent to someone (maybe).  I can add these in if you thing that would be good.

egthomas:
It seems like that particular dat file (`2001102410w.dat`) should be blacklisted since it is corrupted.  I also recall @kevinkrieger volunteering to do a mass-quality control check of the dat and rawacf files on the data mirror.  Maybe we should wait on the outcome of that before we start adding lots of error messages for files that users shouldn't be trying to process in the first place.

pasha-ponomarenko:
I merge it. Good way to start Friday before the long weekend.
;-)

ksterne:
Good to hear here.  I'd moved on from the `2001102410w.dat` file though @egthomas and meant more for the 1993 to 1995 files as well as the 1999 file I mentioned here.  Not sure that we'd need to wait on a mass data quality check to know what check is being done on a function call.  I get there might be other cases where that if condition is met, but then we can modify things in the future.  At least this way if we never come back to this part of the code, we've captured some kind of knowledge or put a place in the code that's helpful.  I'll see about a new branch and pull request to add in some here.

asreimer:
@ksterne I think that's a good idea. I can try to help with testing when you get that put together. Just let me know.

pasha-ponomarenko:
One new issue is the discrepancy between the dates in the filename and those in the prm structure, but I guess it should be passed to DD-WG. I will talk to Kevin Krieger.

egthomas:
@ksterne I tend to disagree with adding a warning message for the case addressed by this pull request.  That would only cause confusion when users see a warning message but their file was processed correctly.

sshepherd:
I agree with Evan on this point. We should have a program that checks for
consistency of files that are located on the mirror and not add additional
error checking to processing routines for files that do not belong in the
distribution stream.
Simon

kevinkrieger:
Hi everyone, sorry for being late to the party. Early last week I implemented dmap integrity checking on every file that is uploaded to the mirror. So far it has caught the files listed below, all with the error "INTIAL INTEGRITY: Initial integrity check shows record size mismatch". I'm not sure of the details at this point, but a check for other issues can probably be done at the same time (like timestamps being reasonable, beam numbers being reasonable, etc...). 

Now deciding what to do with these files is the next step I suppose. There were talks about cutting the file short at the first record with a problem.

`
'20180401.1401.00.zho.rawacf.bz2', '20180402.0801.00.zho.rawacf.bz2', '20180422.2322.57.ade.a.rawacf.bz2', '20180422.2322.57.adw.a.rawacf.bz2', '20180423.0001.00.ade.a.rawacf.bz2', '20180423.0001.00.adw.a.rawacf.bz2', '20180423.1602.00.ade.a.rawacf.bz2', '20180423.1602.00.adw.a.rawacf.bz2', '20180424.2001.00.ade.a.rawacf.bz2', '20180424.2001.00.adw.a.rawacf.bz2'
'20180624.0601.00.hkw.rawacf.bz2'
'20180114.0201.00.unw.rawacf.bz2', '20180502.1001.00.zho.rawacf.bz2', '20180502.2328.25.ade.a.rawacf.bz2', '20180502.2328.25.adw.a.rawacf.bz2', '20180503.0035.00.ade.a.rawacf.bz2', '20180503.0035.00.adw.a.rawacf.bz2', '20180503.0045.00.ade.a.rawacf.bz2', '20180503.0045.00.adw.a.rawacf.bz2', '20180504.1801.00.ade.a.rawacf.bz2', '20180504.1801.00.adw.a.rawacf.bz2', '20180504.1845.21.ade.a.rawacf.bz2', '20180504.1845.21.adw.a.rawacf.bz2', '20180504.2001.00.ade.a.rawacf.bz2', '20180504.2001.00.adw.a.rawacf.bz2', '20180505.2201.00.sys.rawacf.bz2', '20180508.0800.23.unw.rawacf.bz2', '20180528.0601.00.hkw.rawacf.bz2', '20180530.0601.00.hkw.rawacf.bz2'
'20180601.0601.00.hkw.rawacf.bz2', '20180620.0601.00.hkw.rawacf.bz2', '20180620.2110.13.mcm.a.rawacf.bz2', '20180623.2201.00.hkw.rawacf.bz2'
`

PR #144, opened 2018-04-11
Add option error handling to remaining binaries

egthomas:
This pull request adds command line error handling to the remaining binaries in the RST (similar to what was implemented for binaries that used the `new`/`old` flags).  A couple of other fixes were made to documentation that I noticed while testing the changes, and I renamed the `eval` binary to `rst_eval` to avoid a naming conflict with the POSIX `eval` built-in.


mts299:
Should I just check how it handles miscellaneous input or is there more tests cases of error handling that it does?

egthomas:
It's really just checking whether you enter an unrecognized command line option (like `-garbage`), in which case it should print a warning message about the unrecognized option and exit.  Nothing as clever as making sure the input file is in the correct format.

mts299:
error handling and added documentation looks good to me. Going to merge it now.

PR #151, opened 2018-05-04
Add 16-bit png support to fbuffer library

egthomas:
This pull request adds support for 16-bit png images to the fbuffer library in png.c by using the png_set_strip_16() to strip pixels down to 8 bit. Have also changed a call to png_get_color_type() to png_get_bit_depth() for determining whether an image is 16 or 8 bit.

This change can be tested using the `pngd` binary for displaying a png-format image on an X-terminal:

`pngd awesome_superdarn_figure.png`

A response to the SuperDARN software survey requesting graphics support for more user-friendly formats such as png or pdf prompted this fix.

### DIFF
--- a/codebase/base/src.lib/graphic/fbuffer.1.19/src/png.c
+++ b/codebase/base/src.lib/graphic/fbuffer.1.19/src/png.c
@@ -1,4 +1,4 @@
-/* ppm.c
+/* png.c
    ===== 
    Author: R.J.Barnes
 */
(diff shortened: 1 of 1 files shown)


kkotyk:
I checked this out and it seems to open pngs whereas it throws errors on current develop

egthomas:
@kkotyk on a roll today!

kkotyk:
Yep just trying to crush a few of these.

Issue #113, opened 2017-12-05
Binary help messages


ksterne:
As found in #111, there are a few binaries that could use some work to make them a little more user friendly at the command line.

`Test_fitex2` also doesn't behave as usual, though if you try to give it a `-new` it at least gives a warning, but `--help` doesn't offer anything (outside of this PR, but I'll start an issue to note it).  Same with `test_lmfit`. 

`lagfr_fix` doesn't seem to have any help notation at all.

`make_fitex` doesn't seem to have any help notation at all.  Same for `make_fitex2` and `make_lmfit`

I get that these were more recently binaries that were written by folks that aren't with SuperDARN anymore, but if they're in our package, we should acknownledge where we need to do some work.

egthomas:
Personally I think `lagfr_fix` should be removed entirely.

pasha-ponomarenko:
@egthomas, is this code allows for adjusting the first range gate? 
I recall finding out in 2011-12 that the horizontal gaps in the range time plots are due to inaccurate timing between the emission and sampling. A procedure to estimate the delay is implemented on VT website, and this code might be developed in order to re-write the correct value into the par structure of an offending rawacf file. 
This problem became rather widespread with switching to digital radars so it requires a more general solution like introduction of a first range gate offset in the hdw (analogous to tdiff).

ksterne:
I think there was a different set of code that was used to fixed data from 2011-2012, but I imagine that the 'lagfr_fix' code here works much the same.  Why do you think that it should be removed though @egthomas?

egthomas:
My main concern stems from the fact that the routine is entirely undocumented.  Who wrote it?  (It wasn't actually distributed in pre-VTRST3.5 releases of the RST so I doubt it was Rob.)  What is its purpose?  When should it be used?  What do the command line options control?  Why does it only modify lagfr/frang and not any other operating parameters in the prm block?  

If the code is meant to address a specific issue with the data then I'm not aware of that issue being documented anywhere online (maybe in a workshop presentation or mailing list discussion? neither of which are publicly accessible).

In other words, users shouldn't have to dig into the source code to understand what a routine does if we're distributing it to the general community - we should apply some level of quality control before allowing code into the general software distribution.  Things shouldn't get a pass just because they were added before this working group started maintaining the RST.

ksterne:
[...]

PR #154, opened 2018-05-08
Remove some test fit binaries

ksterne:
As noted in discussions in #113, some of the test_* packages should be removed as they aren't well supported or used outside of VT.  This pull request removes the `test_fitacf`, `test_fitex2`, `test_lmfit` packages.  I guess there are two basic tests with this:

- does it compile? (Should)  Make sure to remove binaries in ~/bin/ directory and then see they don't come back after recompiling.
- does anything outside of rst you're using use those packages? (Doesn't seem like it)

I was able to point our website code away from the current branch and to older binaries from RST-4.0.  So, there shouldn't be any impact on the vt.superdarn.org website tool that utilizes these binaries.


ksterne:
I can try to add that in here if there's other interest in doing that.  Otherwise, I was just trying to show we could remove these without breaking things on the VT website as I'd noted in #113.

egthomas:
Merging this for the sake of progress leading up to the workshop, with a strong recommendation for the removal of the fitacfT.2.4 library in the near future.

PR #156, opened 2018-05-13
Removing fitacfT.2.4 library

egthomas:
As the title indicates, this pull request removes the `fitacfT.2.4` library which was only used by the recently removed `test_fitacf.1.1` binary in #154.  This should help reduce confusion / compilation issues when working with the different fitacf libraries.


ksterne:
Whoops!  I should have known to look in the libraries part as well in my pull request.  Maybe when I'm back  on more steady internet I'll test this out as a for sure check.  I suspect it's probably pretty good though.

kkotyk:
This seems fine to me. Im merging it.

PR #158, opened 2018-05-18
Fix trim_fit to allow removal of first record in fit file

egthomas:
This pull request addresses a bug found by @sshepherd where `trim_fit` will not remove a record if it is the first record in a fit file and is the only one which needs to be removed.

I've also fixed a seemingly unimportant bug where the microseconds were not being correctly read by the `FitGetTime`, `RawGetTime`, and `IQGetTime` functions called by `FitSeek`, `RawSeek`, and `IQSeek` respectively.  Previously the microseconds were always set to zero by these functions because they were looking for `"time.us"` with a `DATASHORT` rather than `DATAINT` type (as the microsecond value is defined in [rprm.h.](https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/radar.1.22/include/rprm.h)).

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/trim_fit.1.20/trim_fit.c
+++ b/codebase/superdarn/src.bin/tk/tool/trim_fit.1.20/trim_fit.c
@@ -361,6 +361,9 @@ int main (int argc,char *argv[]) {
                             prm->time.sc+prm->time.us/1.0e6);
 
 
+    /* rewind file pointer in case we want to trim first record */
+    rewind(fp);
+
     /* skip here */
 
     if ((stime !=-1) || (sdate !=-1)) { 
--- a/codebase/superdarn/src.lib/tk/fit.1.35/src/fitseek.c
+++ b/codebase/superdarn/src.lib/tk/fit.1.35/src/fitseek.c
@@ -130,7 +130,8 @@ int FitSeek(int fid,
       if (ptr==NULL) break;
       tfile=FitGetTime(ptr);
       DataMapFree(ptr);
-      if (tval>=tfile) fptr=tptr;
+      /*if (tval>=tfile) fptr=tptr;*/
+      fptr=tptr;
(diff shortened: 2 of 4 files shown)


kkotyk:
I'm trying to figure out how to test this, but it seems like you can only extract records down to the minute. Is there a particular file I can use to extract only a single record?

egthomas:
[trim_test.zip](https://github.com/SuperDARN/rst/files/2254291/trim_test.zip)

Try testing this pull request against the attached FITACF file.  It should have one record from 20140327 23:59:46 UT followed by continuous records from 00:00:01-00:29:53 UT on 20140328 from FHW.  If I'm remembering the issue correctly, the current develop branch version of `trim_fit` can't remove that first record by setting something like `-st 00:00`, but with this pull request that first record will be removed.

kkotyk:
Can you give me an example of how to run this correctly? Its not trimming anything for me.

egthomas:
`dmapdump 20140328.fhw.keith.fitacf > keith.fitacf.dump` (this should show the first record corresponds to the previous day, ie 20140327)

On the `develop` branch:
`trim_fit -sd 20140328 -st 00:00 20140328.fhw.keith.fitacf > 20140328.fhw.trim.fitacf` (this will try to remove any records from before 00:00 UT on 20140328)
`dmapdump 20140328.fhw.trim.fitacf > trim.fitacf.dump` (this should show the first record still corresponds to the previous day)

Then on the `fix_trim_fit` branch:
`trim_fit -sd 20140328 -st 00:00 20140328.fhw.keith.fitacf > 20140328.fhw.fix.fitacf`
`dmapdump 20140328.fhw.fix.fitacf > fix.fitacf.dump` (this should show the first record is now at 00:00:01 UT on the correct day)

kkotyk:
Yes this working. I never had a start date set when I tested which Im guessing why it wasnt trimming anything.

PR #157, opened 2018-05-17
Adding documentation for make_fitex, make_fitex2, and make_lmfit binaries

egthomas:
This pull request addresses @ksterne's remaining concerns from issue #113 by adding documentation for the `make_fitex`, `make_fitex2`, and `make_lmfit` binaries.  I can't speak to the actual functionality of any of those binaries (although I think there are some fundamental issues with `make_fitex` and the `fitacfex` library stemming from subsequent changes to the fit data/software structure).  Also note that I haven't added documentation for the `lagfr_fix` binary as no examples have been provided for how/when it should be used.

PR #161, opened 2018-06-11
Updating make_fitex and the fitacfex library to no longer immediately seg fault

egthomas:
Following discussions with Ray at the SD workshop last week, this pull request updates the `make_fitex` binary and `fitacfex` library so that calls to `make_fitex` no longer immediately throw a seg fault.  This was occurring because the two underlying .c files were never updated along with the rest of the RST to follow the pointer/structure convention currently being used by the other fitting libraries.

It should be emphasized that I have not modified any of the underlying algorithms and have only made the bare minimum changes to ensure successful execution of `make_fitex` (which I believe was originally designed to fit data collected with tauscan pulse sequences).  Also note that documentation for the `make_fitex` binary was added separately in #157.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/make_fitex.1.4/make_fitex.c
+++ b/codebase/superdarn/src.bin/tk/tool/make_fitex.1.4/make_fitex.c
@@ -54,9 +54,9 @@
 #include "errstr.h"
 #include "hlpstr.h"
 
-struct RadarParm prm;
-struct RawData raw;
-struct FitData fit;
+struct RadarParm *prm;
+struct RawData *raw;
+struct FitData *fit;
 
 struct OptionData opt;
 
@@ -89,7 +89,11 @@ int main(int argc,char *argv[]) {
   time_t ctime;
   int c,n;
   char command[128];
- 
+  char tmstr[40];
+
+  prm=RadarParmMake();
+  raw=RawMake();
+  fit=FitMake();
 
   OptionAdd(&opt,"-help",'x',&help);
   OptionAdd(&opt,"-option",'x',&option);
@@ -128,7 +132,7 @@ int main(int argc,char *argv[]) {
        fprintf(stderr,"File not found.\n");
        exit(-1);
      }
-     status=OldRawRead(rawfp,&prm,&raw);  
+     status=OldRawRead(rawfp,prm,raw);
   } else { 
     if (arg==argc) fp=stdin;
     else fp=fopen(argv[arg],"r");
@@ -137,7 +141,7 @@ int main(int argc,char *argv[]) {
       fprintf(stderr,"File not found.\n");
       exit(-1);
     }
-    status=RawFread(fp,&prm,&raw);
+    status=RawFread(fp,prm,raw);
   }
 
   command[0]=0;
@@ -152,12 +156,12 @@ int main(int argc,char *argv[]) {
 
 
   if (vb) 
-      fprintf(stderr,"%d-%d-%d %d:%d:%d beam=%d\n",prm.time.yr,prm.time.mo,
-	     prm.time.dy,prm.time.hr,prm.time.mt,prm.time.sc,prm.bmnum);
+      fprintf(stderr,"%d-%d-%d %d:%d:%d beam=%d\n",prm->time.yr,prm->time.mo,
+	     prm->time.dy,prm->time.hr,prm->time.mt,prm->time.sc,prm->bmnum);
 
 
 
-  FitACFex(&prm,&raw,&fit);
+  FitACFex(prm,raw,fit);
     
   if (old) {
     char vstr[256];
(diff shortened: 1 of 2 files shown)


pasha-ponomarenko:
On our way back from France, changing flights in Toronto...

ksterne:
@pasha-ponomarenko, hope your travel back went well!  Definitely way outside of the scope of this pull request but relates to something that we talked about at the workshop, what's the difference between make_fitex and make_fitex2?

egthomas:
Here is the relevant explanation from [AJ's 2013 Radio Science paper](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/rds.20031):

> FITEX2 represents an iteration on an earlier routine called FITEX. FITEX was originally developed as a new method to fit velocity data from a particular pulse sequence, known as tauscan, which is described in Greenwald et al. [2008]. Similar to FITACF, a weighted linear least‐squares fit to the log of the ACF lag powers is used to resolve SNR and spectral width. However, the phase fitting differs significantly. Instead of the linear phase fit, 120 phase variation models spanning the 180° range at 1.5° intervals are calculated and compared to the phases of the actual data. The number of models used is a trade‐off between the computational burden of doing the comparison, which increases with the number of models, and the resolution of the velocity determination. These predetermined models are then fitted to the actual sawtooth pattern (e.g., Figure 3a) without the need for an initial guess on the number of 2π cycles across the ACF lag range. Once this comparison is complete, the model that produces the lowest root‐mean‐squared (RMS) error is chosen to be the best fit. If this best fit produces an RMS error more than three standard deviations below the mean error across all models, then it is determined that the fit is valid (“good”), and v is calculated from the model. The major disadvantage of FITEX is that it only provides ≈ 25 m/s velocity resolution because of the discrete steps in the phase slopes of the 120 models. The result of the model comparison with the data in Figure 3 is shown in Figure 5.

> The subsequent version, FITEX2, was developed to solve Doppler velocity to arbitrary resolution. For improvement of Doppler velocity resolution, the best model phase fit from FITEX is used as an initial guess for the bisection method algorithm described in Press et al.[1992]. This provides arbitrary velocity resolution, while only increasing processing time by the equivalent of 5–10 model comparisons. For the data in Figure 3, FITEX2 returns v=−916 m/s, w= 73 m/s, and SNR = 20 dB, which are essentially the same as determined by FITACF. The FITEX criteria for a good fit are also used in FITEX2.

ksterne:
Thanks for the info here, I'm guessing this probably can't be condensed down enough to provide some kind of context and added to #157 could it?  Otherwise, I agree here that this fixes the immediate seg faulting issues.  I'll leave it open for comment on my question and then otherwise merge it in.

kkotyk:
Should reviewers be assigned to this?

pasha-ponomarenko:
@egthomas, any suggested reviewers?
Any volunteers?
I guess this is just to chech that it does not segfalut any more.

pasha-ponomarenko:
>which I believe was originally designed to fit data collected with tauscan pulse sequences... 
@egthomas, should it be tested specifically against ACFs obtained using tauscan? I recall that tauscan "kills" some of the "bad" lags in real time during or right after averaging the pulse sequences, which is normally done by FITACF.

egthomas:
@pasha-ponomarenko - when I asked Kjellmar about this he indicated that the original tauscan code created some kind of modified RAWACF file which stored additional information in the lag table.  Over a decade ago though, this code was accidentally erased when Rob upgraded the software at the WAL radar.

I've actually been meaning to ask about this for a while now - is rawacf data collected using current implementations of tauscan still incompatible with the standard fitting software, and thus require the FITEX or FITEX2 library to be used?  Or does the current version of tauscan no longer produce a different style RAWACF file, and therefore can be processed using the standard routines?  

Going back to @ksterne's original question, I guess there are two fundamental questions related to this pull request:

1) What is the difference between FITEX and FITEX2?  (which I think is answered in AJ's paper)
2) What is the relationship between FITEX/FITEX2 and tauscan?

pasha-ponomarenko:
>does the current version of tauscan no longer produce a different style RAWACF file, and therefore can be processed using the standard routines?
 
>What is the relationship between FITEX/FITEX2 and tauscan?

These compatibility/interdependence issues require further investigation.
As we know, tauscan is an alternative type of pulse sequence, while fitex/fitex2 is an alternative way of phase fitting. On the surface, the algorithms are independent of each other, but they seem to be forcibly married in Kyellmar's software. Again, there is an issue of the on-site "bad" lag rejection by tauscan that can generate side effects if the respective raw data are processed by FITACF. 

This is all I can tell at this stage.

egthomas:
I agree it's worth further investigation, but in the meantime this pull request simply restores functionality so the FITEX routines can (eventually) be compared to the other fitting libraries.

pasha-ponomarenko:
OK, if there are no volunteers, I will have a go. What branch I need to download?

pasha-ponomarenko:
I presume, fix_fitex

ksterne:
@pasha-ponomarenko, my note kind of hides it a while ago, but I already found this was good.  I left it unmerged for any relevant discussion on fitex/fitex2.  Maybe it's worth moving this to an issue as I think we've got enough for this pull request here.  I'll merge it in now.

pasha-ponomarenko:
OK, both fitex and fitex2 generate fitacf files from WAL raw data (ground scatter) without any problem. The data look OK except that most of the fitex velocity values are significantly larger. This issue should be considered later, but for this pull request the testing seemd to be finished.

pasha-ponomarenko:
Gee, just missed it by 30 seconds...

ksterne:
Sorry I stole it out from under you.  Currently transferring some of this info in high-level to help out with #157.

PR #164, opened 2018-06-15
Add dependencies for CentOS 7

asreimer:
I just successfully installed RST4.1 on CentOS 7. It required installing the following dependencies:

hdf5-devel
libpng12-devel
libX11-devel
libXext
libXext-devel
netcdf
netcdf-devel
ncurses-devel
zlib-devel

I've added a column for CentOS 7 to the README.md file.

Long term, I'm not sure this is sustainable. I'm about to install RST on a Fedora 26 machine. Where should I put the dependency installation info? I don't think the README can handle another column in the table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,18 +5,18 @@ Radar Software Toolkit
 
 In addition to this code, you will need the following packages:
 
-Debian 8.7 | Mint 18.1 | OpenSUSE 42.2 | Ubuntu 16.04 | macOS 10.12.4
----------- | --------- | ------------- | ------------ | --------------
-gcc | libc6-dev | gcc | libhdf5-serial-dev | libhdf5
-libhdf5-serial-dev | libncurses5-dev | hdf5-devel | libncurses-dev | libnetcdf
-libnetcdf-dev | libnetcdf-dev | libpng16-devel | libnetcdf-dev | libcurses
-libncurses | libpng16-dev | libX11-devel | libpng12-dev | libpng16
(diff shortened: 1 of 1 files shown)


asreimer:
Fedora 26 dependencies:

hdf5-devel
libpng-devel
libX11-devel
libXext
libXext-devel
netcdf
netcdf-devel
ncurses-devel
zlib-devel

egthomas:
Maybe we shift the installation dependencies to a github wiki page and add links to the readme?  Or we could create a separate file for listing the dependencies?  I agree the table in the readme has become unwieldy.

asreimer:
I think that's a good idea. What about moving __all__ of the installation instructions into an installation page and then putting a link to it in the README file? Or the dependencies can still be listed in the README, but only using their generic names. Specific dependency names for specific OSes would then be in the Wiki as you suggested.

So how do I contribute to the Wiki? I don't see an edit button when I visit the Wiki. It looks like there isn't the same pull request functionality for the wiki as for this repo.

egthomas:
@asreimer that may have been a permissions issue - let me know if you can edit the Wiki now.

kkotyk:
If this is being converted to a wiki with changes to the README, can I close this PR @asreimer since entire set of instructions will change?

asreimer:
Just leaving a note here about Fedora 28 dependencies. I'll modify the Wiki later:

For CDF library install, needed to install: ncurses-devel
For rst install, needed to install: libpng-devel libX11-devel hdf5-devel libXext-devel netcdf-devel

PR #165, opened 2018-07-02
Add warning message for bad num_bytes

ksterne:
This is a follow-on from #139 though it doesn't relate to original issue that it fixed in #116.  I'm catching up to where some of the discussion was and realized that the last record is supposed to return `num_byte` as zero.  So I've added in a check here to check for that value being less than zero.  My examples for tests here are:
```
make_fit -old 2001102410w.dat w_fit
cmpraw 2001102410w.dat 2001102412w.dat
```
I understand the `2001102410w.dat` is likely corrupted, but adding this kind of warning message (I've been debating upgrading it to an error) will help in detecting problem files in the future.  The `develop` branch though will end the `make_fit` call with a `File not found.` which leaves a user (and some light developers) to believe they might have made a typo.


sshepherd:
I will reiterate my opinion that adding error messages to make_fit is
misguided. You are working with a corrupted file that does not belong on
the data mirror. A much better solution is to create a program that checks
for consistency of files _before_ they make it on to the mirror.

Simon

ksterne:
I understand I'm working with a _likely_ corrupted file.  And I agree, a program should probably be created that checks for the consistency for files...oh wait, there's this thing called `test_raw`.  Maybe that program should be checked out before calling to re-create it.  For now, I'd like to add these warning messages for the future whenever another corrupt file is encountered.  That way we can quickly and easily identify a possible issue with a file without having to run `test_raw` because I don't always run that program on every file before trying to use it.

The spirit of this pull request is to try to make the code more user friendly and readable.  There seemed to be a good bit of call for that at the Workshop.

sshepherd:
I disagree. test_raw (or whatever) _should_ be run on _every_ file _before_
it goes onto the mirror. We need to do more quality control up front.
Adding friendly messages simply slows the code down. Our processing code
should be light-weight for speed. Minimize or eliminate error checking,
which is done as part of the data distribution, preferably _before_ files
make it into the distribution stream.

Simon

ksterne:
@sshepherd, tell me the last time a _new_ dat file went onto a mirror.  This code doesn't touch anything rawacf related.  If you'd like, I can do the performance test to see how much adding this code in slows things down.  Is that necessary here?

Also, I mean trying to use it as in opening it up.  File corruptions on local file systems which people have a copy of rst are separate from mirrors.

I absolutely agree with the point you're making, but that's separate from what this can cover.

egthomas:
There may not be any new dat files being added to the mirror, but does this mean we start adding unique warning messages to the rawacf I/O routines each time we encounter a corrupted rawacf file?  Already adding warning messages in 4 different places for a single corrupted dat file seems like overkill.

sshepherd:
Fair enough that we have not had "new" dat files in over a decade, however,
I think we should have some sort of overarching philosophy or guidelines
here, which I suggest as streamlining the processing code and adding
quality control layer, which can check files and provide you all the
messages that you want. As for the file corruptions on local files systems,
that should be taken care of locally. You are suggesting a modification to
the core code because "maybe" you have a corrupted file? Again, as a
quasi-mirror (lots of data flow through VT) there should be some quality
control, but that should be separate from our standard processing tools. It
may be a small (imperceptible even) increase, but this won't be the last
request to add even more "checking". So, I am suggesting that we add this
layer of quality control, which should flag such a file, and these types of
messages will be unnecessary.

ksterne:
@egthomas, to answer your question, yes!  I believe that adding warning messages to make the code easier to debug and reduce the amount of time to figure out where things are going wrong is a good thing.  Do you oppose this goal and think the code should be more difficult to read/debug and that time should be wasted figuring out that a file is breaking at the same point in the future?

Do you have a problem with the number of messages (4 to 1)?  If so, I can reduce the scope of this pull request to only fit the one place that the one file I've found.  I did four messages to follow on from the `num_byte` checks that you put in for #139.

@sshepherd, I agree with you 100% that quality control is a goal, but that's better served for the DDWG.  What we can do here though is make the rst code more useful as I've noted previously in this comment.  To answer your question, yes.  I am suggesting this pull request for when files go bad (bits can go bad on hard drives for no reason).  Otherwise a user is left thinking the file wasn't found, that they made a typo that they weren't giving the binary the correct arguments or in the right format.  My thought is this warning message helps a user figure out something is wrong with the file instead of bad input.

With the logic that we're making modifications to core code because of corrupted files, then we should revert #139 and blacklist the files identified in #116 as they must be corrupted or bad.

To address streamlining concerns, I made a bash script to repeatedly run the `make_fit` command in my initial comment 500 times.  The average time it took to run the loop was:
- `develop` branch: 13810781 ns
- `fix_oldraw_warningmsg` branch: 12656264 ns

I ran these a few times and they were all about the same though these are the lowest values.  And yes, the `develop` branch code ran slower.  I can post my bash script if you'd like to do the same.

So, the difference is on the order of milliseconds or sub-milliseconds.  Is there another part here that I'm missing with the code not being streamlined?  If it's on extra lines of code, then there's a trade off on usability/documentation vs. lines of code.  I don't think adding this in opens the flood gates for us to add lots of other print warning and error messages.

egthomas:
> Do you oppose this goal and think the code should be more difficult to read/debug and that time should be wasted figuring out that a file is breaking at the same point in the future?

https://en.wikipedia.org/wiki/Straw_man

ksterne:
@egthomas, yea OK that was a little bit much there.  But I think it's because I don't understand why adding in these print statements doesn't seem like a good idea.  Maybe you can elaborate some more if my point below doesn't make sense.

I think it seems like good practice to add in a warning/error message where a known problem has been found as it's possible that problem could happen again.  Granted, I also understand it may never happen again.  But if it does, then at least we're not trying to remember where in the code it breaks.  Or it may help eliminate places of failure when debugging in the future.  (Maybe I'm just repeating the straw man argument as before)  Does that make sense?

sshepherd:
I stand by my objection to modifying the core processing code.

asreimer:
Hey @ksterne,

I'll try to test this soon (tonight/tomorrow). I have reviewed the code and do not see anything that you changed that would break anything.
 
I'm going to step in it here...

I want to comment about the discussion had here. I think we should remember that this is a public forum and that how we interact reflects on the larger community. This is what I have observe as an outside observer in the last day:
1) @ksterne is clearly communicating and asking questions trying to understand concerns. He's even addressed concerns raised and asked for clarification, but has been met with silence or unfair comments about __how__ he asked a question.
2) @sshepherd and @egthomas are defending a seemingly arbitrary stance on what is a good or bad pull request. When concerns are addressed, they double down. It really looks like a bad faith approach to the discussion. 

__TL;DR version__: My view is that @ksterne has addressed @sshepherd and @egthomas concerns, but they are being arbitrary in what constitutes a __bad__ pull request. I, like @ksterne, would like to know what is a good or bad pull request so that I can understand why certain pull requests in the past have been merged, but this one suddenly cannot be.

## test_raw vs. warning messages
I agree with @sshepherd and @egthomas that there should be quality checks upstream of the "core processing code". However, I think @sshepherd and @egthomas are being arbitrary about what constitutes __bad__ changes to the core processing code that they like or don't like. As @ksterne already pointed out, why did #139 get merged if changes to the core processing code are bad? You could easily ask the same thing about any pull request. It seems to me that this is mostly a philosophical/stylistic preference rather than one rooted in practicality.

@sshepherd stated that he doesn't want error messages added to make_fit, but @ksterne added error messages to the raw file reading codes, not to the higher level make_fit code. The changes were made in the I/O parts of RST. Just wanted to clarify that. It seems to me that test_raw would need to make changes to the I/O parts of RST in a similar way? A brief read of the test_raw code shows that it calls the raw file I/O routines in a similar way as make_fit... 

@sshepherd mentions he doesn't like changes because speed should be a priority. I agree with this to a point. Obviously there are other things that should be prioritized other than pure speed, otherwise we should be focused on rewriting huge amounts of RST in Fortran and using a better file format than dmap, etc. It's all about a weighted balance of priorities. @ksterne showed that the changes he made do not affect the speed in a meaningful way, but then in response @sshepherd simply digs his heels in before disengaging from the discussion. That's not helpful.

## What is a bad change/value in enduser useability?
What value is there in useability of the code? There are members of the community that spend more time than others working with the processing code (@ksterne is one of them). I don't think making the code easier to use for them, at the cost of a couple milliseconds, is a bad thing. As @ksterne correctly pointed out, this one pull request does not open the flood gates for a ton more warning messages.

@egthomas, that strawman comment is not fair. At all. @ksterne has raised valid points that you haven't addressed, while he has addressed yours. Sure, he could have asked those questions a bit different, but I think it's pretty clear he's approaching this discussion in good faith. From my reading of this thread, you and @sshepherd have made up your mind in a seemingly arbitrary way about what change is allowed and isn't allowed and are not clearly communicating why you oppose this change compare to others, such as #139. Is it that #139 doesn't print warning messages? That's the only meaningful difference I can see between this and #139. And your decision seems arbitrary because you haven't been as clear as @ksterne has been.

## The complaints
1) This pull request doesn't slow the code down in a meaningful way, so that's @sshepherd's main complaint addressed and made moot.
2) Having people run test_raw potentially requires modifications to test_raw for people to know what the problem with a raw file is. These modifications would look something like what has been done in this pull request anyway.
3) I'm not really sure what @egthomas's complaint is. It seems to be the same as @sshepherd's, but that's already been addressed.

## Summary
The checks that @ksterne put in are definitely useful for a user because they help the user figure out if they made a mistake in their input or if there's an issue with the file, all while have an almost immeasureable affect on the speed of the core routine.

What I see is this pull request making things easier for the people who have to work with processing files on a daily basis without any practical negative impacts to the codebase. I think we should keep in mind that people are already stretched on time that would be better served getting things done for the community. I see value in this pull request without any practical detriment.

Also, I think people should try to actually test a pull request before simply casting it aside.

I'll test the code and return with more comments.

pasha-ponomarenko:
Just come to the office -- it was the Canada Day (long) weekend here...

I see no harm in adding this message as long as it helps to identify a problem and does not lead to a crash.

egthomas:
@asreimer I'm not going to try defending any points which I haven't been making.  I think there has been a general misunderstanding between pull request #139 and this one, likely because both are dealing with the value of `num_byte` but for very different purposes.  If you go back to the end of #139, @ksterne was originally proposing adding error messages for the 1993-1995 dat files which the software could not (at the time) read yet contained valid data.  I raised a concern about that, and it appears @ksterne changed the thrust of this branch/pull request.  That's perfectly fine, and is wholly separate from this pull request.

> However, I think @sshepherd and @egthomas are being arbitrary about what constitutes bad changes to the core processing code that they like or don't like. As @ksterne already pointed out, why did #139 get merged if changes to the core processing code are bad? You could easily ask the same thing about any pull request. It seems to me that this is mostly a philosophical/stylistic preference rather than one rooted in practicality.

It's probably worth pointing out that @sshepherd and I have not actually seen each other in several weeks, often have philosophical differences with regards to software implementation, and have not been coordinating our points in this discussion.  Continuing to measure this pull request against #139 is not a fair comparison, and it has nothing to do with who the commit authors are or if it's judged "good" or "bad" by someone.  And I'm certainly not arguing against adding warning messages to the code, which again is not the sole difference between this one and #139 and is not why I raised a (single) concern earlier.

The past procedure for dealing with corrupted files by the SuperDARN community has been to blacklist them as they are identified, thus removing them from use by the larger user base.  What has been proposed in this pull request is a shift in that approach which I feel should be recognized.  My question earlier in the discussion of this pull request was simply whether we are going to start applying similar customized error messages for each new corrupted raw data file we encounter.  I personally believe that for this **particular** case of a file that should and will be blacklisted (2001102410w.dat), adding warning messages for a potentially non-repeatable file corruption issue does not increase code readability.

> As @ksterne correctly pointed out, this one pull request does not open the flood gates for a ton more warning messages.

But that's precisely what was advocated for by @ksterne in response to my question:

> There may not be any new dat files being added to the mirror, but does this mean we start adding unique warning messages to the rawacf I/O routines each time we encounter a corrupted rawacf file?

> @egthomas, to answer your question, yes!

So no, my concern has not been clearly addressed.  Anyways, assuming this pull request does go forward then I would suggest a more accurate error message describing the integrity test being performed (something like `num_byte < 0 in record header, potentially corrupted` rather than `Issue reading num_byte`).

asreimer:
Ok @egthomas that's much more helpful and makes things a bit clearer than before. I'm going to push back a bit on it being unfair to compare #139 with this pull request though. I think it is totally fair! Yes, they are different in some ways, but they are still both being used to address problems with the raw files. Technically, a __better<sup>TM</sup>__ fix than what was implemented in #139, which is consistent with the thinking laid out here, would be to fix the old raw files so they don't have the zero length records at the ends of the files. So, I think it's these subtle inconsistencies that make it hard to understand the objections to @ksterne's proposed changes.

__Re: the shift__
I think this change is or will be necessary to some extent if the DAWG are going to convince the DDWG to run `test_raw` anyways on all the raw files. Clearly there is some overlap between DAWG and DDWG in regards to software. Changes will probably have to be made to the raw file I/O code in RST to accommodate their needs. Changes like checks in the raw file I/O that provide verbosity in regards to problems with the data files. Understanding what is wrong with a particular raw file beyond the catch all "it is corrupted" could be important. For example, it could be possible to fix the num_bytes in the dmap file. It could also be used to help identify an operational issue with a particular radar or storage media, etc.

One last thing I'll point out: the Saskatoon group wrote their own dmap file I/O code in python because the C code in RST had little/no error checking and would cause all kinds of problems in the real time displays. So maybe this means the DDWG might just use that tool to check the raw files on the mirrors? It is verbose about what is actually wrong with the input file, which from an operational point of view is valuable diagnostic information. If so, maybe that makes this who pull request moot? If anyone here is interested in what they did, they should talk directly with the Saskatoon group because I'm fuzzy on some details. @ksterne, maybe it's worth revisiting the python version of the dmap library for operational stuff if things can't be done within RST.

ksterne:
I thought I was clear in my original note on this pull request that this one is really just a follow on to #139.  What I really mean is I discovered this while working on #139.  The issues that are addresses are completely different.  As I noted, I had to catch up a little here as I hadn't realized (most...I've only tested a handful) dat files ended with the `num_byte` variable being zero.  That's why I ended up having to add in f14dc67.  I won't claim to be an expert on dat files, so maybe this assumption is not true and please correct me if I'm wrong.

@egthomas, thank you for elaborating or mostly reiterating your point here.  I understand this is a change from how things have been done in the past, but I think this seems to be a grey area with how I stumbled upon this issue.  I'll reiterate a point that the file I've put in the example doesn't look (at first glance) like it's corrupted with the `develop` branch when using these routines.  I only encountered this file as the code prior to #139 caused a seg fault.  Now, the processing just ends with a not-so-helpful `File not found.`  So, if I'd started processing the 20011024 file now, I could overlook it without prior knowledge about the file.

I completely understand that this error may not come up again, but since it has once, the possibility isn't zero that it will come up again.  I tend to agree with @asreimer here that we could be helpful to the DDWG group if we could present to them what we're finding or have the rst code a little bit helpful on noting errors.  That's not to say I think a goal should be future proofing for all possible errors, but where it might make sense when we find a place where the code could provide helpful feedback to the user.

To clear up some confusion on my part and relating back to #139 though, I think I didn't see this pull request as a big issue since the test of `num_byte` is less than or equal to zero.  @egthomas, were you finding the files in #116 had `num_byte` less than zero?  Otherwise, could the condition just have been `== 0`?  This goes back to my assumption that non-corrupt dat files end with `num_byte` being zero which I accept could be wrong.

Thanks for the feedback on the warning message @egthomas, I've added that in as that captures more of what this is checking.

kkotyk:
So despite there being a thesis of comments here, I still do not understand the final decision on this PR. Are we going with it, or are we not?

ksterne:
@kkotyk, it seems as though @pasha-ponomarenko is good with it and I go with him as chair for DAWG.  Apologies to you and anyone else combing through the discussion here, I know I did not help in some points to be straightforward.

@egthomas, are there large outstanding issues with this pull request?  I understand the concern of the RST repo isn't necessarily for finding file issues, but I don't think there is another codebase in the DDWG that is looking for file errors.  And if not, then the reading/opening routines here could be very useful to the DDWG for finding files that have errors.

I think one of the biggest reasons I made this PR was to make the error/warning output more useful than noting `File not found.`

ksterne:
Any motion towards this pull request?

mts299:
I will help test if you need someone too. Just let me know some examples of tests and expected outcomes :)

ksterne:
@mts299, that would be great.  Look back to the original posting where I included a test.  The difference between `develop` and this branch is `develop` ends with "File not found" where as this branch prints out a message.

sshepherd:
But that file has been blacklisted since it has been shown to be corrupt,
correct?

mts299:
I cannot find that file on our server. Would any file with zero bytes be okay to test with? 

I know we have a few that I was playing around with.

mts299:
@ksterne Got a hold of the dat file, ended it up it was on our mirror not in the removed folder :/ 
Goes to show these files sometimes makes it into the mirror.

Did your test and this is what I get: 
`make_fit -old 2001102410w.dat w_fit` 
WARNING:rawopen:OldRawOpenFd:num_byte < 0 in record header, potentially corrupted file.
File not found.

`cmpraw 2001102410w.dat 2001102412w.dat`
WARNING:raw_open:*raw_open:num_byte < 0 in record header, potentially corrupted file.

Just some comments: 
1. Might want to put some spaces between the ':' and the message to make it easier to read 
2. Do we want it to be a WARNING? This seems to be more of an ERROR?
3. If we throw an ERROR, do we still need a file not found? It is a bit miss leading as the file was found but there is an Error or Warning that it is corrupt.  
4. Also, is this something to be added to other binaries? I know we have some zero byte arrays in our fitacf files on our servers and probably some corrupt rawacf files on the mirror. 

Test with corrupt rawacf, this might be a different problem: 
`make_fit 20070117.1001.00.han.rawacf >  han.fitacf` 
Segmentation fault (core dumped)

ksterne:
Thanks for taking a look at this @mts299.  Glad you were able to get ahold of the files and sorry I wasn't quicker to help out on that.  I've pushed a commit that should address your first comment.  I think that makes more sense if I'm printing that out to the screen; I think my mind was more buried in logging files when I initially wrote that.

I've noted this as a warning as a way to soften concerns that @egthomas and @sshepherd may have expressed in this pull request or #139 that throwing an error may be too strong as there may be usable data up to the point that the file goes bad.  I think it's also possible that some files may end with num_bytes being zero and that it's completely normal.  I think it's better at this point to stay a warning.  If that's the case, is your 3rd comment still good here?  I might be able to eliminate the `File not found` message, though I think it may require playing around with some error code return values which have likely not been very well defined.

To the fourth comment, it's possible there are other places where there are assumptions on things being found or valid values, but I wasn't exactly trying to address that here.  I don't know that I'd looked to trying to solve or address the seg fault with the corrupted `20070117.10*han.rawacf` file.  It's likely that's a different issue that's not addressed here as the file changes here are to places which deal with the old file format (dat).

Thanks for the review and let me know your thoughts here.

mts299:
>  I think it's also possible that some files may end with num_bytes being zero and that it's completely normal.
I think this gets back to the `File not found` error message since we ruled it as a warning. This `File not found` still seems to be miss leading. This could be another PR if it is out of scope or more time consuming to remove it. 
 

> It's likely that's a different issue that's not addressed here as the file changes here are to places which deal with the old file format 
Sounds good, I will make this a separate issue and investigate it more.

ksterne:
@egthomas, I've merged develop into this branch now and looks like it's passing now.  Hopefully I didn't make a mess with merging back...doesn't look like it.

@mts299, I've had a look at trying to get rid of the `File not found` as I agree that can be confusing.  That was why I thought adding the extra message in here would be helpful.  It doesn't seem like it will be too easy to modify this though as the `File not found` message is coming in because it's seeing the NULL return from the `OldRawFp` function which is a struct of `*OldRawOpenFd` type.  Seems as though the best way I can do this is to return `ptr` back up to `make_fit` and then free it there.  I see there's a `ptr->rawread` that might be some kind of flag (at least it has some -1 returns from the `rawreadcurrent` file.

Can you check this out to see if I'm introducing any other memory leaks?  I'd probably need to do the same freeing of memory for any place that is calling the `OldRawOpen` is being called (basically anything starting with raw/dat as a data source).  Could be about a dozen binaries from first glance.

mts299:
@ksterne 
`make_fit -old 201102410w.dat w_fit`
Segmentation fault (core dumped)

`cmpraw 2001102410w.dat 2001102412w.dat`
WARNING : raw_open : *raw_open : num_byte < 0 in record header, potentially corrupted file.

Soooo in `make_fit` I would say there is a memory leak.

ksterne:
@mts299, I was really asking whether or not you thought it was worth it to edit the binaries that feature return value I was messing with in order to eliminate the `File not found` in the `make_fit -old` call.  I'm guessing from the latest post that you're thinking it is worth it.  It's a little away from the original thought I had for this, but I think it is a good solution for making things less confusing.

mts299:
@ksterne Sorry about that! Miss understanding, I must have gone into the testing mode in terms of trying to find current memory leaks. 

I still think the message is worth it from how you describe the data might still be "good". I cannot see how you can be introducing more memory leaks as long as you clean up the ones your pass around and create in the functions you are altering. (From what I can see in the code you touched for `make_fit` you did that.)

Question, why must you return the pointer? Could you not modify the pointer (as in pass it into the function) but have the function return a return value code? 

This is something I have been thinking about for a while but never got around to saying "why do we not use errno codes and have some SuperDARN warning/error codes?" 

There might be a nice way to orchestrate a nice modular way to handle all of this, but you like said 

>  I'd probably need to do the same freeing of memory for any place that is calling the OldRawOpen

mts299:
@ksterne Sounds good! 

> That sounds like a big goal/mission for this repo in the future though
I have been making some notes on how to make the software cleaner/better for future development potentially an RST 5.0 :) But I agree, something for the future to do.

Let me know when you want me to test again :+1:

ksterne:
Alright @mts299, I think I've got all the binaries updated here.  To add to your list, need to run:

`make_info -old -rawa 2001102410w.dat`

Shouldn't be any `file not found` or seg faults anymore.  Let me know what you find out.

Maybe I'm introducing a merge conflict here with other PRs, but we'll have to see about that.

mts299:
@ksterne There is typo in your example test you picked up an "a" with the option -raw 

However, with that said did some testing and her is my findings: 

Ones that worked:
- [x] `cmpraw 2001102410w.dat 2001102412w.dat`
- [x] `make_info -old -raw 2001102410w.dat`
- [x] `cat_raw -old 2001102410w.dat 2001102412w.dat > new.dat`
- [ ] `cat_raw -old 2001102410w.dat 2001102410w.dat > new.dat`
- [x] `raw_cp -old 2001102410w.dat`
- [x] `dattorawacf 2001102410w.dat > w_rawacf`
- [x] `make_fitex -old 2001102410w.dat > w_fitex`
- [ ] `rawstat -old 200112410w.dat`
- [ ] `trim_raw -old 200112410w.dat w_dat`

Failed tests: 
`cat_raw -old 2001102410w.dat 2001102410w.dat > new.dat`
Segmentation fault (core dumped)

And this command seems to make the original file empty :/

However! It works if you choose a different file for the second one:
`cat_raw -old 2001102410w.dat 2001102412w.dat > new.dat`
WARNING : rawopen : OldRawOpenFd : num_byte < 0 in record header, potentially corrupted file.

`rawstat -old 200112410w.dat` 
Segmentation fault (core dumped)

`trim_raw -old 200112410w.dat w_dat`
Segmentation fault (core dumped)

`rawstat` and `trim_raw` before the changes did `file not found` error now it segfaults :/ 

Questions: 
1. make_fit, raw_cp tend to create empty files or files dmapdump cannot read? I thought the point was to use the data that was there? Or do we expect users to try and trim the files after seeing the warning message? 

2. On `cmpraw 2001102410w.dat 2001102412w.dat` it gave the Warning error. However, I did not have the second file. Should this not return a `file 2001102412w.dat not found `? This issue may be out of scope for this PR but something to think about.

ksterne:
@mts299, I'm fairly confused on the results for `rawstat` and `trim_raw`.  I'm getting the expected warning message I put in.  Maybe try making sure you're on this branch (with the latest pulled) and recompile?

For the error to `cat_raw` that seems more like user input error checking and I'll hide behind the "out of scope of this PR" for that.

To the first question, I'm not sure I ever thought of this with the perspective of gaining some usable data.  It's possible the way I'm exiting is preventing data from being written.  But, the main point of this PR was to add a more helpful message as an output.  I guess my counter-question is does the `develop` branch versions of these binaries create non-empty files? I didn't think so, but I can't remember now.

P.S. I'm not sure if my clicking your checkboxes carries through (interesting feature for github).

mts299:
@ksterne Sorry, I just realized I was giving `rawstat` and `trim_raw` a dat file ... unless that should work? 
 
> For the error to cat_raw that seems more like user input error checking and I'll hide behind the "out of the scope of this PR" for that.

I would too. 

>  guess my counter-question is does the develop branch versions of these binaries create non-empty files? I didn't think so, but I can't remember now.

Sorry, I don't think I understand the question? I believe these binaries will produce some output given a proper file. 

>  perspective of gaining some usable data. 

I think I am going back to the idea of Error vs. Warning. Where an error I would expect no output and a Warning would be "hey your file may be corrupt but I am going to make a rawacf anyways"  
And I think this goes back to your original response when I asked why it was a warning:
> I've noted this as a warning as a way to soften concerns that @egthomas and @sshepherd may have expressed in this pull request or #139 that throwing an error may be too strong as there may be usable data up to the point that the file goes bad. I think it's also possible that some files may end with num_bytes being zero and that it's completely normal. I think it's better at this point to stay a warning. If that's the case, is your 3rd comment still good here? I might be able to eliminate the File not found message, though I think it may require playing around with some error code return values which have likely not been very well defined.

I think if we want to carry on as a warning then something should be produced but that maybe another PR as I imagine it is not a trivial thing to do in the code? Like a subbranch of this branch and push that PR into this branch? thoughts? 

> I'm not sure if my clicking your checkboxes carries through (interesting feature for github).

I never realized you can click them! This would make sense why `rawstat` was checked off. I really like the checklist for things that need to be done or test cases so the developer has a nice list to look at when fixing or cleaning things up :p I also use it in my jupyter notebook for daily TODO's

ksterne:
@mts299, yea, the `-old` should work well with dat files.  So, should try what I mentioned before on recompiling and such.

I feel as though the error vs. warning part is getting to be a little over my head here since I'm not as familiar with the older data and don't work with it very often.  Maybe @egthomas can help out here with how this code behaves with older data.

Keep in mind, we've *very* likely identified a corrupt file that should be not in distribution with the `2001102410w.dat` file.  I went back and look at some of the source of all of this in #116.  The original problem came from some random files where the reading at the end of the file wasn't working well.  That's been cleaned up with #139.  Since I was looking through #116, I did pull one of the files there and tried `make_fitex -old 1994010100k.dat k_fitex` (note a change in your command what has the '>').  The result was a decent sized file.  So, that does work here.  There's also no file that's generated with `raw_cp 2001102410w.dat` unless you do a `>` redirection.

I get these are small spot checks, but I don't have the time here or think it's worth it to try to run down all of the cases with older data that I'm not familiar with.  I really had found one specific file which is corrupted due to what could be a random issue and thought this PR would help identify any corrupted files in the future.  I get it would be better to run this across all pre-20060701 data for all of the binaries which are effected, but I just don't think that is worth the effort and I don't have the time to create that kind of testing code.  If that is the case, I'll close the pull request and remove the branch since I think it's getting down a dead end on a good total solution.

And I know I'm not taking it more than face value and I'm not taking anything to heart by closing and removing here.  I thought this would be a fairly simple and it's scope-creeped a little (though I think for the good and appreciate the guidance on it!) but I don't know enough of these routines to probably make the best solution.

mts299:
@ksterne Sorry for being pedantic, I guess I don't know the history of this PR. However, I do think it is important. I think I need a bit more understanding and time to try and figure out why things are going wrong on my side.

If anyone else wants to test these cases that would be helpful just to make sure it is my mistake and not the codes. Then we probably could merge this PR :)

@egthomas is there any reason why this cannot be an error message?
I do believe this is something that will get extended into other file formats because I do find zero byte arrays in rawacf's and fitacf's.

ksterne:
@mts299, not at all a problem.  I think if I'm expressing frustration (know that it's not at you or anyone, maybe a tired and cranky day) is that I intended this to be a simple thing and I didn't think it would be something that could set the precedent for all future error/warning messaging.  Not that I think you're explicitly trying to do that, but that's an underlying question here that keeps popping up.  I would propose the DAWG discuss in the future on error/warning messaging and get a plan to go forward on.  I'd be happy to change this around if this pull request didn't match that plan.

And not a problem on not fully grasping this, this pull request is long enough by itself.  And then it seems from another pull request that solved a troubleshooting problem Evan found almost a year ago.  So, it's a lot to digest.

Certainly let me know if there's something you're not getting with the `raw_stat` and `trim_raw` binaries.  Maybe this is a case where it's working for me but not you even though the code is the same.  Maybe it's worth diving into each binaries' directory and running a `make clean` and `make` there to see what warnings/errors there might be in the compilation?

mts299:
@ksterne no worries, I was not trying to make it set the precedent. I was just trying to get some discussion going about error/warning messages as I do think they are important. Again, I like this PR as it starts the trend of thinking about how to implement more clear messages. By no means do we have to do it all right here. 

Definitely, a discussion to have with the next meet up with DAWG. 

With that said I was asking why not an error vs a warning just in case there was a clear answer in the previous conversations. I couldn't find one but if we leave it as a warning for now and it works I am good to merge it :) 

I will try to make clean and maybe testing it on another machine to see if that will work. Again,

codebase/superdarn/src.bin/tk/tool/make_lmfit.1.0/make_lmfit.c
```diff
@@ -163,6 +163,11 @@ int main(int argc,char *argv[])
 
   if (old) {
       rawfp=OldRawOpen(argv[arg],NULL);
+      /* Error code for num_bytes less than 0 */
+      if (rawfp->rawread==-2) {
```
mts299: I think I found my problem. What if rawfp returns NULL? This condition with segfault.
Possible solution 
//TODO:  using error codes will make this simpler
if (rawfp != NULL)
    if (rawfp->rawread == -2) 
           free(rawfp);
           ......
else
    fprintf(stderr,"file not found\n");
    .....

codebase/superdarn/src.bin/tk/tool/make_fitex.1.4/make_fitex.c
```diff
@@ -134,12 +134,17 @@ int main(int argc,char *argv[]) {
...
   if (old) {
-     rawfp=OldRawOpen(argv[arg],NULL);
-     if (rawfp==NULL) {
-       fprintf(stderr,"File not found.\n");
-       exit(-1);
-     }
-     status=OldRawRead(rawfp,prm,raw);
+    rawfp=OldRawOpen(argv[arg],NULL);
+    /* Error case where num_bytes is less than 0 */
+    if (rawfp->rawread==-2) {
```
mts299: Same problem here.

codebase/superdarn/src.bin/tk/tool/trim_raw.1.15/trim_raw.c
```diff
@@ -193,6 +193,11 @@ int main (int argc,char *argv[]) {
 
   if (old) {
     rawfp=OldRawOpen(argv[arg],NULL);
+    /* Error case where num_bytes is less than 0 */
+    if (rawfp->rawread==-2) {
```
mts299: same here.

mts299:
@ksterne I found a bug and maybe my problem. Though I shouldn't be getting a "file not found" error when I have a file. So I will keep looking into why `rawfp` is returned NULL in the first place. 

Sorry for giving you more work, but should be a simple fix if you see my suggestion.

ksterne:
Not a problem @mts299!!! This seems like a great find.  Maybe it's easy enough that I could move the check for NULL above the `rawfp->rawread==-2` check?  Or maybe a cleaner/slicker check is to do (in psuedo-code):

```
if rawfp==NULL {
print; exit;
} else if rawfp->rawread==-2 {
free; exit;
}
```

Does that seem ok @mts299?  Or is mixing up the checks for the whole `rawfp` pointer and the `rawfp->rawread` variable within that point not so kosher programming style?  If not, I think I could easily flip the condition checking here so that NULL is checked for first and then the -2 error code.

Are you still using the same `2001102410w.dat` file or are you using a different file when you're getting the `rawfp` returned as NULL?

mts299:
@ksterne That looks good to me :+1: 
No, that is fine for the style. 

I was still using the same 2001102410w.dat that was not empty. But I should probably check with a different file. Sorry got distracted by other projects.

mts299:
Do not ask me how but it works now. I have a feeling it has something to do with the build system and linking the binaries to the newly built code. But that is just suspicion and out of scope for this PR.

@ksterne did you make that fix with the bug I found? Also, do we want someone else to test it?

mts299:
@ksterne Looks good, it seems all of the binaries were changed correctly and no segfaults on my end. 

I feel confident now just confused how my binaries were geting mixed up with other versions.

Good to merge then?

mts299:
I am merging and closing this PR if there are no complaints :+1:

mts299:
@ksterne - are you okay with updating the summary? 
`num_bytes == 0` vs `num_bytes < 0` coule be a typo, can you clarify @ksterne because I actually don't know. 
@egthomas now that I do have your attention, what is the difference between this being a `Warning` vs `Error` message? @ksterne and I left it for now, but I too would like some clarification on this :)  

For the `rawread` we decided to add the extra error value field in because there is no way to determine the difference between `file not found` vs. `Warning: num_bytes < 0` which if both messages are given to a user this can be confusing as I mentioned in my testing of the code. I also mentioned the more elegant way of dealing with this is to use `errno codes`, a standard way of error handling in C. However, that would be a larger task to take on so we left it for a topic to discuss with the DAWG group.

ksterne:
@egthomas, I can definitely work on a summary of this pull request.  What delivery method were you thinking for the summary?  Just an addition post on this pull request?

egthomas:
Just noticed while compiling the develop branch that this pull request has introduced numerous compilation warnings of the type:

```
make_fit.c:192:31: warning: comparison between pointer and integer [enabled by default]
      } else if (rawfp->rawread==-2) {
```

@mts299 this working group doesn't have any official guidelines for warning vs error messages.  I'm not sure it would be productive to rehash the earlier discussion from this thread.

@ksterne yes, an additional post on this pull request would be useful so that all of the information is in one place.

mts299:
That should be a quick fix to do with a simple cast. Nice catch @egthomas. 

@egthomas - I will add the point of clarifying a warning vs. an error in the next DAWG discussion regarding error handling.

mts299:
@ksterne @egthomas Did one of you fix the warning messages?

codebase/superdarn/src.bin/tk/tool/trim_raw.1.15/trim_raw.c
ksterne: Fixed between f978ced and e436a98.

codebase/superdarn/src.bin/tk/tool/make_fitex.1.4/make_fitex.c
ksterne: Fixed between f978ced and e436a98.

codebase/superdarn/src.bin/tk/tool/make_lmfit.1.0/make_lmfit.c
ksterne: Fixed between f978ced and e436a98.

ksterne:
So, I know this is much too late and I probably won't capture the updates in a summary here but here goes. I'm not sure I totally remember the difference of `num_bytes = 0` and `num_bytes < 0`.  I think this goes back to #139, but I thought either could be a corruption in a reading value.

The bulk of the updates were to binaries in commits on Nov. 8, 2018.  In testing @mts299 found that while the initial warning would note an issue with reading or converting the num_bytes value, the code would still print out `File not found.`  So the code was modified to use the `rawread` value as an error code return from the library when there was an issue.  That way, all of the binaries that are using these libraries could check the returned pointer's `rawread` value for the error code.  I don't think I've seen documentation for possible values of `rawread`, but from looking at the code it seems safe to use it this way.  The majority of the commits are updates to each of the binaries to look first for the returned pointer to not be `NULL` and then to look for the returned error code (-2).

If the error code is found then the returned pointer is freed and the binary is exited right away.  That way the binary doesn't continue to run and produce the `File not found.` output that confuses what went wrong with in the code.

@mts299, let me know if I missed something there.  @egthomas, let me know if there's anything else that seems like I could add here.

egthomas:
@ksterne note the compilation errors introduced by this pull request still have not been fixed, eg

```
make_fit.c:192:31: warning: comparison between pointer and integer [enabled by default]
      } else if (rawfp->rawread==-2) {
```

PR #166, opened 2018-07-11
Fixing bug in map_addhmb.c

egthomas:
This pull request fixes a bug in `map_addhmb.c` of the `map_addhmb` binary identified by Bill Bristow when the boundary latitude is specified on the command line, causing the routine to never return.


ksterne:
Is this hotfix worthy into `master` as well?

egthomas:
> How to test it?

Call `map_addhmb` with either the `-lat` option (to set a single HMB latitude for the entire file) or the `-lf` option (to read the HMB latitudes from a text file), eg:

```
map_addhmb -lat 64 19970415.map > 19970415.hmb.map
map_addhmb -lf lat.dat 19990830.map > 19980830.hmb.map
```

https://superdarn.github.io/rst/superdarn/src.bin/tk/tool/map_addhmb/index.html

> Is this hotfix worthy into master as well?

I flat out broke several other functions in the 4.1 release that we never made hotfixes to the `master` branch for, so following that precedent then no I don't think so.

pasha-ponomarenko:
Here are the JPEGs:
original

modified
20160308 hmb

PR #169, opened 2018-07-18
testing git mv keeping history

mts299:
@egthomas @ksterne, Please look at make_fitex, is this what you are looking for regarding keeping the history? 

renamed:
codebase/superdarn/src.bin/tk/tool/make_fitex.1.4 to codebase/superdarn/src.bin/tk/tool/make_fitex


egthomas:
Huh, that's pretty neat.  What git commands did you use to rename the `make_fitex` directory?  Also, why are so many other commits appearing / files showing changes?

mts299:
So I used this script for the first attempt: https://gist.github.com/emiller/6769886#file-git-mv-with-history

pretty much this script uses git's --filter-branch option which allows you to rewrite the history so in this case change all make_fitex.1.4 to make_fitex, thus the abundant amount of commit messages.

kkotyk:
Any downsides or things we need to discuss with this, or can we start with stripping version numbers? should we have one branch where are all the renaming is done, or should we make a separate PR for every bin/lib?

kkotyk:
I think its because it retroactively crawls through every commit to modify the history of the filename and thus they show as changed commits on github. @mts299 probably knows more.

mts299:
@kkotyk is correct. It goes through the git history and replaces all fittex1.4 with fitttex so you can keep the commit history. Is this what we want?

kkotyk:
One thing I notice is a bunch of merge conflicts. I'm not sure how problematic those will be.

mts299:
Yea but they are all idl? This happens a lot when I change branches and I am not sure if these files change a lot or I have them stored somewhere.

mts299:
Fixed the conflicts but the travis-ci failed.

mts299:
Because this is sensitive to conflicts as the conflicts it had was minor to the renaming and being outdated from develop, I am going to close it for now until 4.2 is released and then I will whip through and remove all version numbers if we are okay with how this script does it. 

Also do we want to make a PR for each binary or do it all in one go?

Issue #163, opened 2018-06-15
Which version of RST?


mtwalach:
Is there an easy way to find out which version of/particular RST branch someone is running (something that works cross platform)? Obviously, if I pull directly from GitHub, I can look at my log, but it won't tell me if I actually re-installed it or just pulled it.. I was also wondering if there is something that works for people who don't git.

If this doesn't exist, can we create some kind of help keyword like "rst --help" that will tell you? 
I think this could be quite useful, especially if people reinstall a newer version or branch to test if it's been done properly or if they only use process data occasionally and can't remember what they installed or when they last updated it..

Thanks!

mtwalach:
> Is it easy to add the RST version to the output of RST programs?

That's exactly what I was thinking.. That way, even if you have several copies/versions downloaded, but only one installed you can easily find out which one it is.

egthomas:
@asreimer we don't manually rename the directory when we create a new release; that information is taken from the tag we assign and handled by github.  We also haven't been updating the library/binary version numbers because it makes the commit history difficult to follow.  I tried experimenting with the `git mv` command but wasn't terribly impressed with the result.  

I do like the idea of putting a version file in the top-level RST directory.  How would we handle the case of between-releases?  For example, would the current develop branch retain the 4.1 version number?  Or should it say something like 4.1-develop?  Or 4.2-develop?  Or 4.2-beta?  I like the output of git describe because it indicates which version the branch is based on and how many commits have been made since the release, e.g. `v4.1-67-gf72306a`.  But again that isn't very useful for non-git users.

egthomas:
Addressed in #173

PR #173, opened 2018-07-23
Implementation of new --version option to print current RST version number

egthomas:
This is a test demonstration of a new `--version` option which can be added to any command line binary that will print the current version number of the RST (stored in the newly added `rst.version` file in the home directory) and exit.  To test, try executing

`aacgmeval --version`

This was originally requested by @mtwalach in #163.


mtwalach:
Yes, this is what I had in mind. I tried it and it does what it says!

mts299:
Worked for me! I assume the version will change when it is merged in 4.2 as it currently reports 4.1?

egthomas:
Yeah, that's something we can update on the 4.2 release branch.  @ksterne can correct me if I'm wrong, but I believe the merging order goes something like

`develop` -> `release branch` (new branch)
`release branch` -> `master` (merge)
`master` -> `develop` (merge)

kkotyk:
Good to merge then?

egthomas:
I can add this option to some of the other binaries and include them with this pull request.  Should only take a few minutes.

mts299:
Depends, if it holds more than just RST version I think version.txt is okay but if it is only rst then RST_VERSION.txt would be more clear

mts299:
The version option works for me.

One thing I did notice was that savetodmap --help did not print anything?

egthomas:
I think that's because there is no doc.xml file for savetodmap which populates the help output. I may have added one in the option error handling pull request but it's been so long I don't remember.

mts299:
Okay, I was not sure because I saw the -help option in the code so figured it had one. Should we add one or make that another issue/ PR to make savetodmap doc? 

Otherwise, I think this PR is good to be merged.

egthomas:
I think once this pull request and #144 are merged then it will be easier for someone can go back and look for missing help documentation.

PR #182, opened 2018-08-24
RST 4.2 release back into develop

ksterne:
Following standard dev model, this is pull request of `master` branch back into `develop`.

Only thing that looks off here has to do with the CDF info in the top-level README file.  It's not showing up in the current `master` branch and wonder if those commits will propagate back to `develop` with this pull request.  Do we need a quick hotfix to `master` before this is merged back?


egthomas:
Well crap, I guess this is why we have to be careful about not committing directly to `master`.  Somehow when we reverted @mts299's changes to `master` that negated the subsequent changes to the `develop` branch when we did the 4.2 release.  I'll add her changes back to the develop branch - they'll just need to wait for the next release to become official.

PR #183, opened 2018-08-24
Re-adding README instructions lost in 4.2 release shuffle

egthomas:
This pull request adds the README instructions which were somehow lost in the merging of the 4.2 release branch into the master branch, and subsequent merge into develop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,19 +19,18 @@ netpbm | | netcdf-devel | |
  | | | zlib-devel | |
 
 You will also need the CDF (Common Data Format) library which can be downloaded from NASA.
-You can find the latest release at: http://cdf.gsfc.nasa.gov/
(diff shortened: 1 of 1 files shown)


ksterne:
Continuing the discussion from #182, this was probably caused by the reverting and second pull request process.  Somewhat surprised it didn't show up as a merge conflict since I think there should be two different commit hashes marking the same file.  Maybe there wasn't a merge conflict since the original commit was reverted.  I'm not sure.

Anyways, this is fairly straight forward and matches 9c7ef4c; so I'll merge this once the checks are completed (just for completeness).  The commit I pushed is just a copy of the `4.2-release` branch file that for whatever reason recognizes the newlines or carriage returns in lcurses to lncurses modification lines on the web-GUI.

PR #192, opened 2018-10-18
Update to hdw dat bks

ksterne:
This pull request is an update to the hdw.dat.bks file from measurements of tdiff.  These were pushed up to the hdw.dat. git repo earlier this month.

Not expecting much of a test here other than general compilation since these are edits to a text file.  I'm guessing the general travis-ci checks should be good enough, but pull it to your local machine if you don't trust it.


sshepherd:
Hi Kevin,

I have a few questions and concerns about your updated hdw.dat file for BKS. I was looking at MCM with Bill after he made the upgrade to USRPs and needed to change the value of tdiff. He arrived at a value of 39 ns. Looking at the data from MCM for 2018 you get something like this:

image

The blue dashed line is only for reference and corresponds to the Chisham virtual height model.
When I look at BKS data for 2018 I get the following for 11.0-12.0 MHz:

image

and the following for 10.5-11.0 MHz:

image

Neither look very good. I would recommend holding off on any reprocessing of data with these values until you can look more closely at the data. It could be that the value of tdiff is just not what you think it is, but you could also have a more serious problem. I don't know.

I would like to look at more BKS data but the last file I have is 20180505.0602.00.bks.rawacf.bz2. Perhaps the BAS mirror is not synched to Globus (I suspect that is a good thing) or you haven't been pushing your BKS files to the mirror.

A last point is that you are quoting your tdiff value to a tenth of a nanosecond! You might recall the presentations by the tdiff taskforce that showed daily fluctuations of tdiff at several radars on the order of a few nanoseconds. I would question whether the value is that accurate, particularly when factors such as temperature, operating frequency, beam direction, etc. are likely to introduce uncertainties or difference that are much larger than .1 ns.

egthomas:
@sshepherd do your plots for BKS (and MCM) include data from all available beams and months?  Also, what is the scale in the upper-right of each figure?

sshepherd:
No, this is for a single beam, but yes, all months available in 2018. The
other beams look similar. The distribution is scaled to a percentage of
returns at each range, just like Gareth did in his virtual height paper and
I have been showing at recent Workshops. The take-away is that for ranges
<2500 km most of the returns are between 50-60 degrees, which suggests that
tdiff is not correct. Beyond ~2500 km it could be low statistics causing
the large spread in elevation angles but it could also be something more
sinister. Perhaps more data would help.

egthomas:
> The distribution is scaled to a percentage of returns at each range, just like Gareth did in his virtual height paper and I have been showing at recent Workshops.

Oh I think maybe your colorbar has been cropped out of the figures you uploaded.  Not a big deal.

sshepherd:
@pasha-ponomarenko 's value of -320 ns looks better for the E-region scatter (which he uses to determine the value of tdiff from the data), however things don't look so good for half-hop F-region and 1-1/2 hop scatter.

11.5-12.0 MHz
image

10.5-11.0 MHz
image

So I don't think it is just a different value of tdiff.

egthomas:
Circling back to these hdw updates, I see there was also an update to the Hokkaido West hdw.dat file on the vtsuperdarn repository (https://github.com/vtsuperdarn/hdw.dat/commit/c2efd72b10b1002d438357626df4b094919cbbb9). Should that file be included in this pull request too?

ksterne:
Hey all, I will absolutely agree that the value for tdiff I've measured in hardware may not be completely accurate.  As I noted over in the hdw.dat repo (again, we've got to consolidate one day) the value that I have here is from best measurements of delays in the hardware systems.  It continues to baffle me why hardware measurements do not agree with data.  The intent of making this pull request was to get the first attempt value out there so that others like @sshepherd, @pasha-ponomarenko, and @ecbland could do analysis (this is the data analysis group) for a more accurate value.

My apologies on the backlog of bks data from getting out of Virginia Tech.  This has been corrected for sometime now and would be good to see updated values here with more data.  I suspect the scatter at farther ranges is due to low number of returns and/or low SNR (and lower xcfd SNR) from those returns.

@egthomas, as far as the hdw.dat.hkw update, I'm just now getting to this and since there are issues here that may not be resolved quickly, I think it might be best to make a separate pull request.  I'll do that shortly.

egthomas:
In the spirit of bringing the RST hdw.dat files in line with the VT repository, and including @ksterne's best guess at the measured BKS tdiff as a starting point for others to iterate on, should this pull request be approved?

ecbland:
@egthomas Those are good reasons to merge. If a more accurate value of TDIFF is proposed and approved by the Blackstone PI, then this can be addressed with a new PR.

egthomas:
Thanks @ecbland!  Not sure if there's anything else I'm allowed to test or approve at this point.

PR #193, opened 2018-10-24
Returning OS code

egthomas:
This pull request reintroduces the OS code and site tables which were removed prior to the 4.0 release.  The idea is to make it easier for newer versions of the RST to be installed at the radar site to take advantage of bug fixes and revised algorithms for real-time data processing.

Note that the `site.tst.1.8` library (which used to reside in the `usr/` directory) has not been included because it needs a few of the `codebase/superdarn/tk` libraries to properly compile.


egthomas:
@ksterne it's no problem if this pull request ends up being rejected, and I am more than open to suggestions about how best to handle the sometimes murky interaction between the RST and ROS (admittedly on the handful of radars you mentioned).  It would be great if the OSWG rose from the ashes to provide some direction.

pasha-ponomarenko:
I understand the necessity to be able to update FITACF for the real time display needs.  However, I am not sure that re-merging ROS and RST is the optimum solution, kepping in mind that we started RST refactoring from separating them for the resaons mentioned above by Kevin. At this stage I would restrict our involvement in ROS business by creating an issue on this topic, attract attention of ROS gurus and those formally in charge of OSWG to it  and see where it goes. 
Besides the issue of overstepping the DAWG mandate, I personally would feel better of staying away from mixing these things together and prefer to have some sort of interface instead.

egthomas:
@pasha-ponomarenko so whose arm do we need to twist to reactivate the OSWG?

kkotyk:
Instead of creating a PR to reintroduce the OS code, we could consider forking the RST repo and adding the ROS code to that.

egthomas:
I think it would be very useful to have a separate ROS repository, although as @ksterne and others have pointed out there are multiple generations of the ROS in use at the various radar sites.  Licensing concerns have also been raised in the past when I've suggested uploading older ROS distributions to github.

pasha-ponomarenko:
OK, I will email Mike and invite him to contribute to this discussion...

egthomas:
@pasha-ponomarenko any response from @mikeruo?

pasha-ponomarenko:
@egthomas, this is a rethoric question!
I repeat the request!

mikeruo:
Hello all, And sorry for the delay in responding. The OSWG is basically inactive except for the occassional assignment of radar operating parameters.  There are reasons for this besides the inattentiveness of the Chair (me).  I think you guys with your great work on the data analysis side are in a good position to drive some of the decisions on the OS side as this is arising organically, that is, from a real need.

As the other members of the OSWG are even less active than me, I will go ahead and offer an opinion on its behalf.  I agree with Pasha and Kevin that folding ROS back into RST seems less than optimal and even undoes some of the progress that was made in the refactoring. But I am sympathetic to Evan's request that ROS be able to tap into progress made with developing fitacf in order to generate high-quality real-time data. This is an important direction for us and we should put our best foot forward. Does Keith's suggestion of forking the RST repo seem like a solution? If we go this way, does it mean that someone would need to re-fork the RST repo each time there is an upgrade? Seems like a bit of a bother but perhaps not much more than would be involved with getting people to update their code at the radars.

pasha-ponomarenko:
I have no technical knowledge on the pecularities of different incarnations of OS, but it seems to be that they differ significanlty between groups and sites. Does it mean that there might be no common way of upgrading the on-site version of FITACF?

aburrell:
I agree with @egthomas that "it would be incredibly useful to have an archive of ROS versions available somewhere (similar to the rst-archive); even if it's not meant for active development" for both of his stated reasons.

pasha-ponomarenko:
@egthomas, @aburrell, what about creating some sort of a git ground for breeding crosses between the unique (holy?) RST and all sorts of ROS strains? It would contain the above-mentioned ROS archive  and provide an easy access to the RST so that the interested sides can pull out the latest version and excercise in mongrel production. 
Still, I am reluctant to being responsible for the outputs, especially for testing them, as I believe all this is Mike's domain.

aburrell:
@pasha-ponomarenko I think that would work.  I also think that it makes sense for the products that come out of this process to be outside the domain of the data analysis working group, since it involves the radar operations.

The most important thing, I think, is to make the ROS versions visible to everyone.

egthomas:
> The most important thing, I think, is to make the ROS versions visible to everyone.

thumb-up-sign_318-63754

ksterne:
It seems as though the thought is to keep the ROS part out of this repo and instead have the OSWG take up a similar style repo for radar operating code.  I'm in complete agreement with this as I don't think it's too often that plotting software (and some other types in this repo, but not all) are needed at a radar site or for operations.  I feel as though I could be poised to take up creating the ROS repo, but I don't quite have the bandwidth on this.  I'll defer to direction from @mikeruo.

egthomas:
I'll go ahead and close this pull request so the OSWG discussion does not continue to clutter the DAWG.  Please contact @ksterne, @mikeruo, or myself if you are interested in participating.

egthomas:
Following up on this, I've just created a `ros-archive` repository which contains the latest tar files and extracted source code I could find for the QNX4 ROS.1.25 software (which is still used at a majority of radar sites).  According to [this page](https://archive.is/20121211171209/http://superdarnx.jhuapl.edu/software/superdarn/ros/index.html) page, it looks like the last official releases out of JHU/APL were ROS 3.1 and 3.6 for QNX6/Linux - if anyone has a copy of the original install files I'd be happy to upload those as well (I know there are modified/development versions elsewhere on github but it can be useful to have the common starting point as a reference).

PR #196, opened 2018-11-07
Update map_addimf documentation

ksterne:
This doesn't change any functional code and just the documentation.  It would be good to get approval from @sshepherd and/or @egthomas that I've captured things accurately here.  This comes from some notes @egthomas noted to me while using the new mapping software.  The biggest thing is the need to have the solar wind speed in the IMF text file when using `-if`.  We also noticed that the default error code for data coming from the OMNI website uses a very large value whereas it should be zero.  So you can't just straight use the OMNI data and it needs to be modified some.

To see the changes, make sure to run `make.doc` and then look at:
`~/doc/superdarn/src.bin/tk/tool/map_addimf/index.html`
with your favorite html browser.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/map_addimf.1.18/doc/map_addimf.doc.xml
+++ b/codebase/superdarn/src.bin/tk/tool/map_addimf.1.18/doc/map_addimf.doc.xml
@@ -67,7 +67,7 @@
 
 <quote><em>year month day hour minute second bx by bz vx</em></quote>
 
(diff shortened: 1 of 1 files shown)


egthomas:
@ksterne I made a few other minor edits, otherwise everything looks fine to me.

ksterne:
@egthomas, that looks pretty good to me.  Maybe could be more explicit, but decent space science/engineer should be able to figure it out.  Feel free to go ahead.

PR #198, opened 2018-11-15
Fix for SMR output formatting

tdbabas:
Hi

When generating SMR files, I was noticing that in some cases values were running into each other, causing problems when trying to plot. For example, using a sample from the data from the `hal` radar on 2014-01-22:

```
time = 2014 1818804  4  8  12254 959 4  0 110  180  45 151   0
 10 59102103
    21.3     6.1     9.8    10.9
   235.8   -67.0    74.0   215.7
     17    151    150    230
```

With the fix to the output format, the data now looks like this:

```
time = 2014 1818804  4  8  12254 959 4  0 110  180  45 151   0
  10  59 102 103
    21.3     6.1     9.8    10.9
   235.8   -67.0    74.0   215.7
     17    151    150    230
```

### DIFF
--- a/codebase/superdarn/src.lib/tk/smr.1.7/src/smrwrite.c
+++ b/codebase/superdarn/src.lib/tk/smr.1.7/src/smrwrite.c
@@ -101,7 +101,7 @@ int SmrFwrite(FILE *fp,struct RadarParm *prm,struct FitData *fit,int pwr) {
    n = 0;
    for (i = 0; i < prm->nrang; ++i) if (rng[i] == 1) {
     
-     status = fprintf(fp, "%3d", i);
+     status = fprintf(fp, "%4d", i);
      if (status < 0 ){
         free(rng);
         return -1;


egthomas:
Hi @tdbabas - we generally make pull requests into the `develop` branch rather than the `master` branch of this repository.  I've gone ahead and changed that for you.

tdbabas:
HI @egthomas - thanks for changing that. Apologies for the mixup!

egthomas:
@tdbabas no problem, and thanks for making the pull request.  I admit I don't have much experience with SMR files, but I can see how the code was probably only expecting a maximum of 75 range gates and was never updated after some radars started using 100+ range gates.  This looks ok to me unless anyone else has any concerns.

ksterne:
@tdbabas, like @egthomas I'm not too familiar with a SMR file, could you provide some example code that you used to generate the results above?

tdbabas:
Hi @ksterne, @egthomas 

Here are the steps I used to replicate the data above:

1. Using a daily file created from the 2 hourly files, I created a FITACF file:
```
make_fit 20140122.hal.rawacf > 20140122.hal.fitacf
```

2. Using that FITACF file, I created a SMR file as follows:
```
make_smr 20140122.hal.fitacf > 20140122.hal.smr
```

When I tried running the SMR file through our plotting software, it complained of an error, and when I investigated, I noticed the values running into each other, so instead of finding three separate values, it found one large value which the plotting software didn't like.

The above steps were recreated with a 2 hourly file (`20140122.0000.03.hal.rawacf`) with the same effect.

I should also point out that I too am not overly familiar with SMR files, but it seemed to me that this was not expected behaviour.

Hope that helps!

ksterne:
Hey @tdbabas, thanks for the help here.  Probably would have been more apparent if I did a little more searching.  Do you or @egthomas know if there's documentation on the output format for the smr file?  It's not in the [current documentation](https://superdarn.github.io/rst/superdarn/src.bin/tk/tool/make_smr/index.html) for `make_smr`.  I don't think that's for this pull request, but something we should try to capture, if possible.  I'll bring it up as an "issue"("to-do") whenever this is closed out.

Onto the format though, will the additional space mess with current parsing/analysis code?  I'm not sure if anyone else uses these aside from USask, maybe @mts299 can speak to this?  Otherwise it does works as noted here.

Issue #208, opened 2019-01-12
fitACF 3.0 Caution


sshepherd:
Recent difficulties with the Christmas Valley radars reveal a potential issue with using fitACF 3.0. It was suggested to me that I blacklist some of the data files from CVW and CVE during a period when they were experiencing elevated noise levels. Below are RTI plots using v2.5 and v3.0, respectively.

![image](https://user-images.githubusercontent.com/2982719/51074066-46099400-1648-11e9-9269-400d7217c2c0.png)

![image](https://user-images.githubusercontent.com/2982719/51074069-5c175480-1648-11e9-8acb-9c0ba0056670.png)

Both plots show a high level of noise but v3.0 shows significantly more "data" values, which led to the blacklist request. The v2.5 plot is noisy but not nearly as bad as v3.0.

Scan plots for v2.5 and 3.0:

![image](https://user-images.githubusercontent.com/2982719/51074098-eeb7f380-1648-11e9-855e-0acb14f72034.png)

![image](https://user-images.githubusercontent.com/2982719/51074105-fb3c4c00-1648-11e9-9189-292851b241f8.png)

The real problem (which affected our realtime convection maps and anyone making grid files with v3.0) is the data when using standard gridding:

![image](https://user-images.githubusercontent.com/2982719/51074114-0a22fe80-1649-11e9-8984-f9df990d2fbd.png)

![image](https://user-images.githubusercontent.com/2982719/51074117-127b3980-1649-11e9-9a14-880adbe30544.png)

I should also say that processing these files with v3.0 took 2-4 times longer than when using v.2.5. I suspect that has to do with the increased amount of "data" or noisy values that are created.

pasha-ponomarenko:
This has been noticed and discussed before (https://github.com/SuperDARN/rst/pull/74): the old FITCAF seems to "sweep under the carpet" problems with interference (i.e. non-Gaussian noise) while FITACF3 flags them clearly. I believe, this was mentioned at one of our FITACF3 testing presentations in France. 

The processing time is inherently longer due to different ways of calculating power/phase variance.  This is the price for adequate fitting and realistic error estimates.

sshepherd:
Okay, but my cautionary point is that as a naive user v3.0 let’s through
noise in the gridding process. I anticipate you will suggest that the
gridding process should be modified, but it currently isn’t, and I suspect
those bogus vectors are likely in grid files that are being used.

pasha-ponomarenko:
Sure!
[...]

PR #209, opened 2019-01-13
Extending field_plot and time_plot functionality (errors, elevation angles, phi0, etc)

egthomas:
In response to #208, this PR adds functionality for plotting velocity errors from a fitacf-format file to `time_plot` and `field_plot`.  This PR also adds the ability to plot elevation angles using `field_plot` by adding a new field to the `RadarCell` structure (this was already implemented in `time_plot`).

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
@@ -182,9 +182,15 @@
 </option>
 <option><on>-w</on><od>plot spectral width.</od>
 </option>
+<option><on>-phi</on><od>plot phi0.</od>
+</option>
 <option><on>-e</on><od>plot elevation angle.</od>
 </option>
 <option><on>-ve</on><od>plot velocity error.</od>
+</option>
+<option><on>-we</on><od>plot spectral width error.</od>
+</option>
+<option><on>-p0</on><od>plot lag0 power.</od>
 </option>
 <option><on>-vec</on><od>plot velocity vectors.</od>
 </option>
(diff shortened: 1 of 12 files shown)


pasha-ponomarenko:
@egthomas, I just wonder what other parameters can be plotted? We might think about adding more...

egthomas:
@pasha-ponomarenko as of this pull request, `time_plot` and `field_plot` allow for plotting of 

- velocity
- velocity error
- power (p_l)
- spectral width (w_l)
- elevation angle

I suppose phi0 and/or lag0 may be two other parameters worth plotting?  I'm not sure if enough people use the sigma fit of the ACFs or the XCF-derived parameters to justify adding those.

pasha-ponomarenko:
Yep, pwr0 and phi0 would be useful and, probably, w_l_e...

mts299:
@egthomas @pasha-ponomarenko Do you want to do this for another PR?

pasha-ponomarenko:
Whatever is more practical... This is not that urgent.

egthomas:
Something is screwy with the phi0 color scaling, but it's too late in the week for me to look at this any longer.

mts299:
@egthomas can you let me know when I should test this PR.

egthomas:
I don't plan on adding the other plotting parameters to `field_plot` since it shares a data structure with several other binaries so you can go ahead and test the changes there.  Otherwise `time_plot` is good to go except for the `phi0` color scaling - any help fixing that would be greatly appreciated.

mts299:
@egthomas Is this PR ready to be tested again :)

egthomas:
@mts299 yes, this PR can be tested.  There is still some issue with the phi0 color scale but at this point I can't remember exactly what it was.

ecbland:
I think I've identified the problem with `phi0`:

Line 138 of `time_plot.1.8/fit.c`:

`if (fit->xrng !=NULL) tptr->phi0[i]=fit->rng[i].phi0;`
should be 
`if (fit->xrng !=NULL) tptr->phi0[i]=fit->xrng[i].phi0;`

egthomas:
@ecbland thanks, nice catch!  That fixed the `phi0` plotting error.  I went ahead and added the same functionality to `field_plot` to bring it in line with `time_plot`; now `phi0`, `w_l_e` (spectral width error), and lag0 power can be plotted with either routine.  Hopefully that is not considered to be too much scope creep.

ecbland:
Ok, I've tested this again and it's working well! This is a very nice addition to the plotting routines.

PR #216, opened 2019-01-29
Update HMB latitude limit

mtwalach:
This updates the HMB geomagnetic latitude boundary limit from 50 degrees to be at 40 degrees, whilst keeping to high-latitude limit the same (76 degrees).
It thus allows for a better fitting of the HMB when using data from the mid-latitude radars, which goes down to ~40 degrees.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c
+++ b/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c
@@ -36,9 +36,9 @@
 struct CnvMapData *map[3];
 struct GridData *grd[3];
 
-float bndarr[24][26];
-float lathmb[26];
-int latcnt[26];
+float bndarr[24][36];
+float lathmb[36];
+int latcnt[36];
 
 struct OptionData opt;
 
@@ -94,7 +94,7 @@ int main(int argc,char *argv[])
   float latdef=62;
 
   float latref=59;
-  int nlat=26;
+  int nlat=36;
 
   float bndstep=5;
   int bndnp;
--- a/codebase/superdarn/src.lib/tk/hmb.1.0/src/hmb.c
+++ b/codebase/superdarn/src.lib/tk/hmb.1.0/src/hmb.c
@@ -32,10 +32,10 @@
 #include "map_addhmb.h"
 
 float latref=59;
-int nlat=26;
+int nlat=36;
 
-float bndarr[24][26];
-float lathmb[26];
+float bndarr[24][36];
+float lathmb[36];
 
 #define HMBSTEP 50
 
@@ -183,7 +183,7 @@ void make_hmb()
 
       mlt=m;
 
-      lathmb[n]=n+50.0;
+      lathmb[n]=n+40.0;
       latmin=lathmb[n];
 
       bfac=(90-latmin)/(90-latref);


egthomas:
@mtwalach I haven't tested this yet but the changes look identical to those I made in my local IDL code for creating the TS18 model, so tentatively this looks good to me!

mtwalach:
@egthomas Ah, I figured you must have done something to it to build your model. Did you add the changes somewhere and I missed it?

ksterne:
@mtwalach, do you have a good date where there are lower latitude vectors that we could test with?  I'm not as familiar with the process of "added some vectors at lower latitudes".

mtwalach:
Of course - `2010 4 6 12 34 0 2010 4 6 12 36 0` for example. This 2 minute record should have a lower latitude cut-off of 48 degrees and the following few are also at 48 degrees.

egthomas:
@mtwalach I've tested this pull request and just wanted to confirm some of the HMB behavior I'm seeing.  Using this branch, if I have a grid file with only a single 2 min record containing 5 vectors with |v| > 150 m/s all located at:

1. 32.5 degrees MLAT, `map_addhmb` returns a latmin of 40 degrees (good)
2. 40.5 degrees MLAT -> latmin = 40 (good)
3. 41.5 degrees MLAT -> latmin = 40 (not good?)
4. 42.5 degrees MLAT -> latmin = 41 (not good?)
5. 43.5 degrees MLAT -> latmin = 42 (not good?)
...
6. 49.5 degrees MLAT -> latmin = 48 (not good?)
7. 50.5 degrees MLAT -> latmin = 49 (not good?)
8. 51.5 degrees MLAT -> latmin = 50 (not good?)

On the `develop` branch, I get:

1. 32.5 degrees MLAT -> latmin = 50 (expected)
2. 40.5 degrees MLAT -> latmin = 50 (expected)
...
6. 49.5 degrees MLAT -> latmin = 50 (expected)
7. 50.5 degrees MLAT -> latmin = 50 (good)
8. 51.5 degrees MLAT -> latmin = 50 (not good?)
9. 52.5 degrees MLAT -> latmin = 51 (not good?)

Shouldn't a set of 5 vectors located at, for example, 51.5 degrees MLAT return an HMB / latmin of 51 degrees?  To be clear, it seems like this behavior was not introduced by this pull request, but was present all along.  

Attached is a copy of the grid file I've used for testing `map_addhmb` (had to remove the other records to bypass the median filtering):  [20190315.hok.grid.zip](https://github.com/SuperDARN/rst/files/2992993/20190315.hok.grid.zip)

mtwalach:
Hi Evan,

Thanks for this.

Using your file, I also tried it on this branch and I agree that it's a bit weird as I got the following results.
Vectors:     Latmin:
31.5           40 (good)
32.5           40 (good)
40.5           40 (good)
41.5           40 (not good - I think this should be 41)
42.5           41 (not good - I think this should be 42)
43.5           42 (not good - I think this should be 43)
44.5           43 (not good - I think this should be 44)
45.5           44 (not good - I think this should be 45)
49.5           48 (not good - I think this should be 49)
50.5           49 (not good - I think this should be 50)
51.5           50 (not good - I think this should be 51)

As you said, I don't think I introduced this but I am also not sure what is causing it.. It looks like it's **consistently** rounding it.

egthomas:
@mtwalach thanks for looking into that.  I guess one simple answer could be the 5 test vectors are located on the compressed dayside portion of the HMB, rather than the circular nightside portion?  I can't remember where they are located in MLT at the UT of that particular grid record.

mtwalach:
That's a good point - Do you know if it uses the `grid_index` field for this at all? 
Because I tried changing `gmlong` for that record to what corresponds to MLTs around 2300-2400 (i.e. 265.733, 250.733, 260.733, 266.632, 270.667) and the result for setting `gmlat` to 51.5 degrees is still latmin=50.

egthomas:
@mtwalach I think this is the problem:

https://github.com/SuperDARN/rst/blob/7f5d33c8c7cf6490b4c73cd3234f1f10dfeee054/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c#L296-L299

If you remove the `-1` from the end of lines 297 and 299, the expected `latmin` values are returned by `map_addhmb`.

Looking through some older IDL code (ie the original Map Potential stuff from the 90s), that `-1` component is also present.  I'll try asking Mike and Simon today to see if they remember why that additional equatorward offset was included.

mtwalach:
I haven't tested it explicitly, but looking at the code, I think you are right - How strange!

mtwalach:
If you set the vectors at say 51.0000 instead of 51.5000 it sets latmin to 50 - could it be that the gridding has since changed?

egthomas:
So I just had the chance to ask Mike about this and he said it was intentionally included as kind of a "buffer" so that the H-M boundary was not too tight and the convection flows had some room to "close" above the zero-potential boundary.  If we decide to leave it as-is then we should probably document this, since it isn't described anywhere in the literature.

mtwalach:
Thanks Evan. 

Hm, I can see why this was put there but I find this a bit of a strange reason - If I understand this correctly - in a case where the data suggests a boundary should be at 39.5 deg. because that's where the flows go to zero, it is then actually placed at 38 deg., thus making the resulting potentials at 39.5 deg. not zero?

egthomas:
@mtwalach I guess another way to think of it is the HMB algorithm doesn't actually identify where the flows go to zero, only where the lowest-latitude flows can be found that exceed 100 m/s.  Whether the flows actually go to zero (or if propagation conditions / lack of irregularities prevent backscatter) equatorward of this point is often difficult to say.  Thus the additional 1 degree equatorward offset as a sort of fudge factor.

Not saying I agree with this procedure, just trying to understand Mike's reasoning.  Does that make more sense?  Does a 1 degree offset seem appropriate?  Or is the ~0.5 degree offset resulting from the layout of the equal-area grid sufficient?

mtwalach:
Thanks @egthomas - yeah, that makes sense!

Tricky to say if it is appropriate or not. Perhaps we should leave it for now and it could be a discussion point for the future! At least it's a consistent fudge factor..

ksterne:
Sorry I know I'm a little late on this, but is a follow on/up from this code being merged that any/all map files should be reprocessed?  Any/all here is from not taking the time to go through the process of finding where there are lower latitude data that would be included in.

ksterne:
Bueller?  Any feedback on reprocessing all files just to be safe @mtwalach, @egthomas?  I'm guessing yes is the answer here.

mtwalach:
Hi @ksterne!

Thanks for the nudge - I had to think about this.

So, I am working on reprocessing all of our archive and once I have done it, I will be able to provide a list with times the higher latitude HMB limit is an issue. 

Now, if everyone makes their map files in the same way I do, they can use that list, but as soon as you start using different options for the map fitting, this may not be true anymore (i.e. someone may decide to exclude radar X or use a different 'vel' limit for the HMB fitting etc).
Obviously, I am happy to provide the details of how I make the maps etc as well as the list, but it's not quite there yet!

The thing I am unsure about is which processing keywords exactly will affect it - obviously if you use a different gridding or different version of fitacf it may affect it and the `map_addhmb` keywords in particular will make a difference but I am not sure if the choice of model for example does...

In summary, if you use different processing keywords you may have to do the whole thing. So, probably yes.

egthomas:
@ksterne it may be worth waiting until a new version of the RST is released; if more changes are made to even a single piece of the fitting, gridding, or mapping software between now and then you'll need to reprocess everything all over again.

ksterne:
@mtwalach, that sounds good though I think a list of days isn't necessary for me as I'd rather fire of a large script to reprocess a bunch of files.

@egthomas, I certainly take the point that any change would likely cause reprocessing of lots of files.  But not knowing when that next change will happen and the current sluggish release schedule, I don't want data products to be out of date and include possibly faulty/buggy data from non-current processing routines.  It's a problem with how things are setup and trying to keep static data up-to-date with changing processing.  My question was more of being lazy and not digging into the changes here and whether or not they related to processing.  But I've got the answer for now.

PR #159, opened 2018-05-21
Statistical convection model solver (solve_model)

egthomas:
This pull request adds a new `solve_model` binary which, given solar wind / geomagnetic conditions as input via the command line, solves for the model potential and velocity values over either a uniform or equal-area grid and writes to stdout in a fixed-column ASCII-style output.  This supports all available SuperDARN convection models in the RST (RG96, PSR10, CS10, TS18, TS18-Kp).  Below is sample output:

```
Date:  2018-05-21 00:00
Model: TS18-Kp
Bin:   6<Kp<8, Bz-
Grid:  Equal-area (lat_step: 1.00 [deg])

MLAT [deg]   MLT [hr]   Pot [kV] Vazm [deg] Vmag [m/s]
---------- ---------- ---------- ---------- ----------
   46.5000  0.0483871  -0.138252   -90.2309    63.7825
   46.5000   0.145161  -0.136914   -90.2490    63.0686
   46.5000   0.241935  -0.135492   -90.2674    62.3145
   46.5000   0.338710  -0.133989   -90.2861    61.5213
   46.5000   0.435484  -0.132407   -90.3051    60.6902
   46.5000   0.532258  -0.130747   -90.3244    59.8223
...
```
And using a simple IDL routine to read and plot the model velocity vectors and potential contours:

image

Because this program relies on TS18 and TS18-Kp model functionality I am requesting to merge it into the `feature/ts18_model` branch first rather than directly to `develop`.

PR #218, opened 2019-02-06
Fixing bug in solve_model.c when determining longitude spacing for equal-area grid

egthomas:
This pull request fixes a bug when using the `-equal` option with `solve_model` to create an output grid with equal-area spacing but a latitudinal step size not equal to one.  For example, on the develop branch setting the `-equal` option and `-lat_step` to 2, eg

`solve_model -equal -lat_step 2 > develop_equal_2.txt`

gives you this:

develop_equal2

whereas executing the same command on this branch will produce this:

fix_equal2

Likewise for a `-lat_step` of 3, eg

`solve_model -equal -lat_step 3 > develop_equal_3.txt`

produces this output:

develop_equal3

and on this branch:

fix_equal3

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/solve_model.1.0/solve_model.c
+++ b/codebase/superdarn/src.bin/tk/tool/solve_model.1.0/solve_model.c
@@ -301,7 +301,7 @@ struct mdata *get_model_pos(float latmin,int hemi,int *num,
     for (i=0;i<nlat;i++) {
 
       grdlat=i*lat_step+latmin+lat_step/2.0;
-      lspc=((int) (360*cos(fabs(grdlat)*PI/180)+0.5))/(360.0);
+      lspc=((int) (360.0/lat_step*cos(fabs(grdlat)*PI/180)+0.5))/(360.0);
       nlon=lspc*360.0;
 
       for (j=0;j<nlon;j++) {


egthomas:
If it's any help for testing purposes, here is a link to a generic IDL routine I wrote for plotting the `solve_model` output for the original pull request last year:

https://github.com/SuperDARN/rst/pull/159#issuecomment-392333711

Note that I didn't include any code to plot the velocity vectors, but one should be able to add something like `oplot, polarx, polary, psym=4` near the end to draw symbols where the potential / velocity information is written by `solve_model`.

asreimer:
That's a pretty simple fix. I don't get the same plots as you do, but I do get the same behaviour. Side note: The velocity magnitudes are kind of crazy large (50 km/s).

This PR:

    solve_model -equal -lat_step 3 > develop_equal_3.txt

equal3

develop:

    solve_model -equal -lat_step 3 > develop_unequal_3.txt


unequal3


Because this is such a simple PR and I'm able to reproduce the expected behaviour trivially, I'm going to merge it.

egthomas:
@asreimer the default model pattern will be slightly different from my original plots from February because the dipole tilt angle has changed and thus the interpolated pattern will change, however the velocity magnitudes should not be that large.  Are you sure you haven't switched the potential values for the velocity magnitudes?  This is what I get when plotting the output of `solve_model -equal -lat_step 3 > test.txt` for today (velocity color scale 0-750 m/s):

solve_model_test

I've also attached the output text file and none of the velocities appear to exceed 400 m/s: [test.txt](https://github.com/SuperDARN/rst/files/3323185/test.txt)

asreimer:
Here's the [test.txt](https://github.com/SuperDARN/rst/files/3323247/test.txt) file that I get when I run:

    $ solve_model -equal -lat_step 3 > test.txt
    
    Date not set, using today's date
    
On a CentOS 7 server. I fixed my plotting code (see below). The example linked to for how to make the plots doesn't show how to add the quiver and properly use azimuth so I was just guessing. Boy oh boy did I guess wrong haha.

On my Fedora 29 laptop, here's the[test2.txt](https://github.com/SuperDARN/rst/files/3323264/test2.txt) file and resulting plot:

better


Comparing the test.txt and test2.txt, the azimuths are the same, just the magnitudes are messed up. I suspect this is due to some kind of typing/type casting problem or something like that. I can't debug this right now.

So in summary, there's a problem with running solve_model on a CentOS 7 system that causes velocity magnitudes to be too large, but everything is fine on Fedora 29:

asreimer:
No idea why, but the CentOS 7 system is working now. I logged out and back in. Maybe an environment variable needed to be set or something.

Sorry for the hassle.
index

PR #219, opened 2019-02-06
CDF fixes

egthomas:
This pull request is a partial attempt at addressing the issues raised in #211 by adding the `--version` option to several additional binaries, cleaning up spacing in the `dmaptocdf` code, and fixing a bug present in both `dmaptocdf` and `dmaptoncdf`.  The original problem with handling `ltab` arrays still remains to be fixed.

Issue #211, opened 2019-01-21
dmaptocdf not working (at least for fitacf files)


egthomas:
Following these steps (https://github.com/SuperDARN/rst/wiki/Conversion-Tools-(CDF)):

1. `dmaptoskeleton 20181129.cve.fitacf fitacf.skt fitacf.cdfmap`
2. [edit the `fitacf.skt` file for the correct variable sizes]
3. `/usr/local/cdf/bin/skeletoncdf -cdf 20181129.cve fitacf.skt`
4. `dmaptocdf 20181129.cve.fitacf fitacf.cdfmap 20181129.cve.cdf`

I'm seeing error codes for either `BAD_DIM_INDEX` or `BAD_DIM_COUNT` depending on which version of the CDF library I try using.  As far as I can tell, the error is occurring in `dmaptocdf` when trying to write the `ltab` array to the output CDF file.  This issue was originally identified by Lasse Clausen at UiO.

mts299:
Have we fixed this?

ksterne:
Status update: Still an outstanding issue, but seems like lower priority as no one has been using the code this way.

PR #186, opened 2018-09-26
rtgrid cleanup

egthomas:
This pull request fixes several issues with the `rtgrid` binary, including AACGM_v2 and Chisham virtual height model compatibility and typos in several command line options.  Updates are also made to the appropriate documentation pages.

PR #221, opened 2019-02-18
Fixing segmentation fault in dmap.c when dmap array dimension value is corrupted

egthomas:
As the title indicates, this pull request fixes an issue identified by @ksterne when trying to process `20180624.0601.00.hkw.rawacf` on the current `develop` branch.  An `xcfd` array near the end of this file appears to be corrupted with a negative dimension value.  The proposed fix tests for array dim values less than or equal to zero in `DataMapDecodeBuffer` and sets a few uninitialized pointers to NULL before breaking for the later calls to `DataMapFree` and `DataMapFreeArray`.

### DIFF
--- a/codebase/general/src.lib/dmap.1.25/src/dmap.c
+++ b/codebase/general/src.lib/dmap.1.25/src/dmap.c
@@ -1194,7 +1194,9 @@ struct DataMap *DataMapDecodeBuffer(unsigned char *buf,int size) {
     a->type=type;
     ptr->arr[c]=a;   
     ConvertToInt(buf+off,(int32 *) &(a->dim));
-    if (a->dim <= 0) {
+    /* check for possibly corrupted array dimension size */
+    if (a->dim <= 0 || a->dim > 999) {
+      fprintf(stderr,"Warning: %s array dimension possibly corrupted (%d)\n",a->name,a->dim);
       a->rng=NULL;
       a->data.vptr=NULL;
       break;
--- a/codebase/general/src.lib/dmap.1.25/src/dmap.c
+++ b/codebase/general/src.lib/dmap.1.25/src/dmap.c
@@ -1194,6 +1194,11 @@ struct DataMap *DataMapDecodeBuffer(unsigned char *buf,int size) {
     a->type=type;
     ptr->arr[c]=a;   
     ConvertToInt(buf+off,(int32 *) &(a->dim));
+    if (a->dim <= 0) {
+      a->rng=NULL;
+      a->data.vptr=NULL;
+      break;
+    }
     off+=sizeof(int32);
     a->rng=malloc(a->dim*sizeof(int32));
     if (a->rng==NULL) break;


ksterne:
This fixes the problem I'd found and the same that @egthomas found as well.  Is it worth adding a print to stderr warning about the bad value?  Goes back to making a user more aware of what's going on when things aren't so great here, but I don't have strong feelings on adding this or not.

Also hard to test or create the scenario, but wonder what happens if the record isn't at the end of the file.  I think this change will still allow subsequent records be read if they are good, right?

mtwalach:
Hi,
I don't know if this is helpful, but from a user point of view, I'd say it's always helpful to know why things are going wrong! There's nothing worse than having a nice processing pipeline that appears to work and having it randomly fall over and not knowing why..

ksterne:
@mtwalach, to be clear the code here keeps the dmap reading from falling over and failing.  It cleanly handles unexpected values while reading the dmap record.  My thought is more to not make this so quietly done so that the user is aware of the error in the dmap record.

mtwalach:
Ah sorry - yeah,  I guess that would be equally useful!

ksterne:
@egthomas, I'll take the inactivity to mean not wanting an error message here, which again is OK.  I understand this may be a one time thing on this file so it might not be worth the effort or additional code.  What do you think of adding a comment to the code here to at least describe what this if statement is checking for?  That way future devs can remember why this section was added to the code.

pasha-ponomarenko:
I think that having an adequate error message will never hurt.

codebase/general/src.lib/dmap.1.25/src/dmap.c
```diff
@@ -1194,6 +1194,11 @@ struct DataMap *DataMapDecodeBuffer(unsigned char *buf,int size) {
     a->type=type;
     ptr->arr[c]=a;   
     ConvertToInt(buf+off,(int32 *) &(a->dim));
+    if (a->dim <= 0) {
```
asreimer: A comment about what this new code is doing would be helpful.

asreimer:
I agree with @ksterne and @mtwalach. If the file is corrupted in some way, the user should know about it. Failing silently is not good behaviour.

And there should definitely be a comment in the code regardless of whether an error message is thrown.

Below is related, but not necessarily PR specific. Just some things to think about. TL:DR; failing silently is bad, users being given corrupted files is bad, is there any good reason to continue using dmap?

I think that there's definitely a balance that can be struck between verbosity and performance, like letting the user know that there's a problem with the file, so a particular record(s) didn't get read properly. If DAWG is expecting the community to use RST for scientific data analysis, this should be strongly considered. It's super helpful to have user feedback (thanks @mtwalach!).

I've mentioned this before, but these segmentation fault issues raise some questions:
1) Why are there corrupted files being served to users in the first place?
2) Is DMAP the best possible file format to be using for scientific data analysis? (What purpose does it serve that cdf or hdf5 don't already handle? There's a lot of time being spent here working on a proprietary file format that could be spent on other things).

For 2, I think we all know there are many other mature standard data file formats that already solve these problems and don't require users to install a custom software package just to work with SuperDARN data.

egthomas:
A warning message printed to stderr can certainly be included in this pull request to notify the user when a corrupted array dimension value has been encountered.  It should be noted however there are numerous other places throughout `DataMapDecodeBuffer` where memory allocation/freeing issues are correctly handled and no warning message is printed.  

**I cannot stress enough that I am not advocating for any one particular solution or another, just trying to provide some context that there are a lot of other cases where the code can cleanly fail without notifying the user.**

Regarding @asreimer's two questions, corrupted files are not intentionally served to users by the radar PI.  It often falls to the Data Distribution and Data Analysis Working Group members to determine whether a file is truly corrupted or if there is some deficiency in the software.  For example the rawacf file in question was causing a segmentation fault using some versions of `make_fit` with certain command line options but not others.  In this particular case I would guess the rawacf file will end up being blacklisted, however it is an open discussion within the DDWG as to whether we should blacklist entire files if only a single record is corrupted (the practice has generally been segmentation fault = blacklist).  Also, for real-time purposes we don't have the benefit of post-processing analysis by working group members - corrupted records often crash the real-time processing routines (eg #186) and require manual intervention.

A decision to abandon the dmap format is likely well above the authority of any working group and would need to be settled at the PI level.  There is already software within the RST to convert any dmap format file to CDF or netCDF format.  In fact, there are currently a few open issues and pull requests regarding these conversion routines (#211 and #219) - any assistance in helping resolve those issues or improving those utilities would be appreciated.

kkotyk:
I'm very happy to see this, but if we are okay with adding this, why can't we add full error checking? I recall there being a very heated debate about this topic last time it was discussed. I remember there being pretty strong opinions that all files should be somehow checked ahead of time and the reading routines should not have extra code that could affect speed. I disagreed with this opinion, but alas I don't think any resolution was made. I'm not trying to derail this PR or start an argument, I'm simply wondering if opinions have been changed. Any thoughts @sshepherd?

mts299:
@kkotyk @asreimer Thanks for voicing some opinions on hot-topics but I think these types of questions might serve well in issues on github such that we don't scope creep on the PR. 

If we decide to do full file checking then I think that can be a different PR all together when someone does it. 

@egthomas Let me know when you implement the error message in and I will test/review the code for you :)

asreimer:
@egthomas I understand that there are different needs (post-processing vs. realtime), but none of that precludes error messages saying that something unexpected has happened. My point is that failing silently is a terrible practice, regardless of how many other places in the code this is already happening. The relevant question is, do we continue this in this PR, or start changing it by adding an error message? Probably best left to another PR.

In regards to abandoning DMAP altogether, it was not my intent to suggest that, but I was a pretty vague wasn't I. Instead, I'm suggesting that the data that is distributed to the SuperDARN (and larger space physics) community for research purposes be vetted (no corruption possible) and in a different, more widely adopted file format. My stomach does backflips thinking about how much work it would be to completely abandom DMAP. Yeesh.

@kkotyk is right about error checking. This is approaching the Nth time that dmap has required a fix to handle another way that a dmap file was corrupted.

@mts299 you are right, this isn't necessarily the best place for these discussions. There aren't any guidelines on how to address these topics, but I think you are right that it's best done in a separate issue. Another reason guidelines would be nice to have :)

sshepherd:
Hi Keith,

I have muted most RST threads so I am unaware of any on-going discussion
related to this topic, or others. Perhaps this email was addressed to me or
the github muting isn’t perfect.

Since you asked, my position has not changed. Rather than post another
lengthy position piece, I might suggest that periodic telecons would be a
better format for discussing and resolving these sorts of issues. You
should give careful consideration to the make-up of the group in order make
progress.

Simon

ksterne:
To get back to this PR, @egthomas' last post had a good point in it:

> It should be noted however there are numerous other places throughout DataMapDecodeBuffer where memory allocation/freeing issues are correctly handled and no warning message is printed.

I certainly take this is a valid point, why should we treat this error checking different than any of the others?  I've thought about this on many commutes home.  My thoughts usually come back to not perpetuating that kind of practice or software style.  I'm not exactly sure who might have written the other parts, but we are now in a position to do better.  So I think we should make the code better.  Yes, I agree there could/should be a larger effort to update other error checks, but with this one it seems easy since we've figured out why this check needs to be in here.

Another way I've thought of it, is we could perpetuate the current style of little to no warning messages or we can set a good example and start adding some in.

egthomas:
After some additional testing with the real-time feeds via rtgrid and valgrind, the fix in this pull request does not address the issue when the array dimension (`a->dim`) is corrupted in the other direction, ie an absurdly large positive value:

```
a->dim: 1
a->dim: 1
a->dim: 1
a->dim: 1769209869
==56548== Warning: set address range perms: large range [0x395d8040, 0x1df2d8074) (undefined)
==56548== Invalid read of size 1
==56548==    at 0x4204C0: ConvertToInt (convert.c:26)
==56548==    by 0x41E180: DataMapDecodeBuffer (dmap.c:1207)
==56548==    by 0x403B43: FitCnxDecodeIPMsg (decodemsg.c:46)
==56548==    by 0x403AB6: FitCnxRead (fitcnx.c:76)
==56548==    by 0x402974: main (rtgrid.c:498)
```

The DataMap format can presumably handle any configuration of scalars and arrays, so we should be careful about assigning an upper limit on `a->dim`.  Maybe something like `a->dim > 999` would be sufficient to catch corrupted array dimension values while still allowing valid files?

egthomas:
Also, if anyone is interested in seeing this in action then now is a great time - the Canadian radars (SAS and PGR in particular) are crashing my rtgrid processes at least once per hour during this current Discretionary Time interval.

egthomas:
The commit I made a few weeks ago (0e623bc) should hopefully address everyone's concerns regarding printing a warning message when this particular corruption issue is encountered and is in line with the scope agreed to at the WG meeting.

asreimer:
This looks great and I thin it resolves all the concerns/issues people raised. I ran the following on the develop branch and on this PR with the following results:

    #develop
    $ make_fit 20180624.0601.00.hkw.rawacf > 20180624.0601.00.hkw.fitacf
    double free or corruption (!prev)
    Aborted (core dumped)
    
    #this PR
    $ make_fit 20180624.0601.00.hkw.rawacf > 20180624.0601.00.hkw.fitacf
    Warning: ED.Bgfn� array dimension possibly corrupted (-2004271360)

Then, I was able to plot the `20180624.0601.00.hkw.fitacf` file with `time_plot`:

    $ time_plot -et 23:59 -png -a -b 7 -et 08:00 20180624.0601.00.hkw.fitacf > plot.png

and obtained:

plot

sshepherd:
What is "ED.Bgfn�"? Is that the error message that a user sees?

Warning: ED.Bgfn� array dimension possibly corrupted (-2004271360)

egthomas:
@sshepherd `ED.Bgfn�` is the name of the dmap array with the possibly corrupted dimension value (as identified by the check `if (a->dim <= 0 || a->dim > 999)`).  Sometimes the array name is still meaningful when the dimension is corrupted, sometimes it isn't.

asreimer:
Yes, that is what is printed to the user. Since `a->dim` is corrupted (-2004271360), `a->name` can also sometimes be corrupted, resulting is a weird name like for this case: ED.Bgfn

asreimer:
@egthomas haha you beat me be a second :)

sshepherd:
I guess the question then is whether that is the _best_ message to print to
a user? Will a user that can't read documentation on how to change the axes
on a plot have any idea what this means?

Issue #215, opened 2019-01-28
Installation Guide


asreimer:
It's been awhile coming, but I've created a draft installation guide here: https://github.com/SuperDARN/rst/wiki/Installation-Guide. This was suggested in previous issues here.

The intention of this wiki page is to replace the OS specific notes in the "Installation Notes" in the RST README.md, as per the discussion in #164. I think we decided that the wiki is a better place for this because updating the installation documentation shouldn't need to be done with pull requests.

Once people are happy with the wiki page, the README.md can be updated to point to it. The README.md will still have some installation information, but will be general and not really OS specific.

From what I can see, here are the things left to add to the wiki Installation Guide:
- [x] Ubuntu 18.04 LTS installation instructions (should be similar to previous versions)
- [ ] macOS Conda installation instructions
- [x] macOS Homebrew installation instructions
- [ ] Windows 10 installation instructions
- [ ] macOS instructions for different versions?

It is possible to install RST without using macports or homebrew. This summer I installed RST on a student's macbook using only conda. I have an old macbook that friend gave me, so I'm hoping I can get it up and running and test the conda-only installation process. After that, I'll explore the windows installation process.

Finally, @aburrell, do you know if the macports installation instructions are dependent on precise macOS version? Are the instructions the same on High Sierra (10.13) compared to Sierra (10.12)? Sierra is the current version discussed in the RST installation notes.

Note: for windows, I've found this which might be helpful: https://github.com/esa/auromat/blob/master/README.rst

mts299:
This is an interesting choice on how to use the wiki. Typically the README will handle all installation information as it is the first place a user goes and reads as suggested here https://guides.github.com/features/wikis/

I would imagine the wiki containing release notes, choices made in the software (like why DMAP format files) and maybe some examples or some information on the software, like scipy does https://github.com/scipy/scipy/wiki/Dropping-support-for-Accelerate
¯\_(ツ)_/¯

asreimer:
@mts299 I have no strong opinions either way. I tried adding additional installation information to the README.md in #164 and it was suggested we make it a wiki page.

It would be really nice if we could agree to some standards so that people can do things "properly" the first time. Nice thing here is that the wiki page uses the same markdown as the README.md file.

mts299:
@asreimer I understand. I was just pointing out a guideline stated by github and followed by other repositories. But if this is the standard or preferred for us, I am okay with it. 

Just thought I would point it out if anyone was interested in what github standards are.

[...]

Issue #210, opened 2019-01-18
RST Documentation Revamp


mts299:
I have noticed a lot of "issues" #201 #197 #195 #194  revolving around documentation lately which I think is amazing :+1: 
However, the question might be where to put things and how to move forward in our documentation.
The following discussion points:

- [ ] RFC's are they something to keep maintaining or should we archive them and move to something different? 
- [x] More user-friendly documentation setup for RST functionalities? Read the Docs might be an option. 
- [x] RST for dummies - where should the link live?
- [x] RST for dummies - what platform should it be shared on Overleaf vs. google doc? 

I raise these questions to give us more of a direction on documentation, also I find the current sifting through RST documentation annoying and I will be starting to document pydarn with Read the Docs and I am willing to start an RST version as well to kill 2 birds with one stone. 

Also, RST for Dummies can get discussed here to help figure out its home and other details. 

Let's discuss!

aburrell:
I think Read the Docs is a good place for the RST code documentation.  Even if we decide to move it in 10 years or so, we'll have all the documentation written and in an easy-to-read format.

mts299:
I have no problem with starting to transfer information over from RFC's to read the docs as long as there is potential future for it :)

ecbland:
I don't have any experience with Read the Docs, but would it be an option for hosting RST for Dummies?

aburrell:
Yes, though perhaps not the best option.  It has some good features for coding examples and document strings, but (as far as I know) is not as good for creating a parallel pdf that can be downloaded and digested offline.  But other people may be more experienced in its use and have other opinions.

mts299:
@aburrell, you beat me to the punch. Read the Docs would be good for code documentation. 

Would markdown or latex be of interest for RST for Dummies? Markdown we can host in GitHub but also make it pdf available.

pasha-ponomarenko:
A note in passing: I am not sure about _markdown_, but I like the old good _LaTeX_: it is easy to use, and, more importantly, **I** know how to use it!
;-)

egthomas:
It's also worth remembering that while the current .xml documentation is a pain to sift through in its html format, it is also used to build the command-line help which some people use quite frequently.

mts299:
@egthomas can you elaborate on the process of using XML to build the command-line help?

asreimer:
+1 to moving towards a documentation format that is both inline with the source code and automatically updated using tools like readthedocs.

ecbland:
[...]

PR #227, opened 2019-04-19
Update README.md for CentOS

aburrell:
Updated README with package instructions for the CentOS operating system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,18 +5,18 @@ Radar Software Toolkit
 
 In addition to this code, you will need the following packages:
 
-Debian 8.7 | Mint 18.1 | OpenSUSE 42.2 | Ubuntu 16.04 | macOS 10.12.4
----------- | --------- | ------------- | ------------ | --------------
-gcc | libc6-dev | gcc | libhdf5-serial-dev | libhdf5
-libhdf5-serial-dev | libncurses5-dev | hdf5-devel | libncurses-dev | libnetcdf
-libnetcdf-dev | libnetcdf-dev | libpng16-devel | libnetcdf-dev | libcurses
-libncurses | libpng16-dev | libX11-devel | libpng12-dev | libpng16
-libpng12-dev | libx11-dev | libXext6 | libx11-dev | libx11
-libx11-dev | libxext-dev | libXext-devel | libxext-dev | cdf
-libxext-dev | | netcdf | netpbm | netpbm (10.77.03_2+x11)
-netpbm | | netcdf-devel | |
- | | | ncurses-devel | |
- | | | zlib-devel | |
+Debian 8.7 | Mint 18.1 | OpenSUSE 42.2 | Ubuntu 16.04 | macOS 10.12.4 | CentOS 6.10
+---------- | --------- | ------------- | ------------ | ------------- | -----------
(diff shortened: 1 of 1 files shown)


asreimer:
See discussion in #215. I think the direction being taken here is to put all installation instructions in to an `INSTALLATION.md` file or something like that in the root directory of the repo. There's also been some recent discussion about using readthedocs.

In the mean time, there are instructions (here)[https://github.com/SuperDARN/rst/wiki/Installation-Guide] for CentOS 7, but nothing yet for CentOS 6, so they could be added there.

aburrell:
@asreimer until the INSTALLATION.md file is added, it looks like the README.md is still the right place to record this information.  I added a line to the wiki, though I can't confirm the yum statement, since I don't have the appropriate permissions on the system I'm using.

asreimer:
Note #164.

aburrell:
Noted.  I just don't see the point in making things hard to find.

asreimer:
That's a valid concern, but not the problem. The issue is that if we keep adding installation instructions into the README in the current format, that table will be pages wide, which is not helpful. Also, it's nearly impossible to read that table with a text editor.

Point is we need a better solution. I think I'm hearing that you like the INSTALLATION.md idea, which is what the discussion in #215 was sort of also coalescing on too. I'm waiting for some more feedback in there and then I'll make a PR to introduce an `INSTALLATION.md` file.

aburrell:
If there is an INSTALLATION.md file or some other update to the README.md that replaces the table, then I will retract this pull request.

ecbland:
The only proposed change here is to add the list of packages for another OS (i.e. a new column). I can see that we are running out of space for more columns, so a simple solution for now would be to reformat the table into a horizontal list for each OS, i.e. 

---
**Debian 8.7:** gcc  libhdf5-serial-dev  libnetcdf-dev  libncurses  libpng12-dev  libx11-dev  libxext-dev  netpbm
**Mint 18.1:** libc6-dev libncurses5-dev libnetcdf-dev libpng16-dev libx11-dev libxext-dev
...
...
**CentOS 6.10:** [wow, it's really slow to copy & paste from the table!]

---
Not as pretty as the table but it is much easier to copy & paste into a terminal :) Would this be acceptable @aburrell @asreimer ? 

We have several decisions to make regarding the documentation, but I think this PR can be dealt with in the scope of the current README file.

asreimer:
@ecbland ya that's a good compromise IMO. I'll work on that now.

asreimer:
Hmmm. Given discussion in #210 leading to the `doc/read_the_docs` branch, I have another question actually. I'm going to ask this question in #215.

ecbland:
@asreimer Now that this information is in readthedocs, I guess we can close this PR (without merging)?

ksterne:
@ecbland, good catch on the readthedocs only having CentOS7.  I've updated the `doc/read_the_docs` branch.  Can @asreimer or @mts299 check things out here?  I think that should close this PR if so.  I should note, I've just moved the notes from @aburrell and haven't tested this myself.

mts299:
I checked things out and did one fix from what you added @ksterne. Thank you for updating this! 

Closing PR and hoping to make a new PR with these changes and more within in the week :)

PR #228, opened 2019-04-30
Adding shortcuts to commonly used binaries on documentation front page

egthomas:
One of the issues raised during the DAWG telecon a few weeks ago is that it's hard for users to find the documentation for specific binaries like `make_fit`.  To help address this problem, this pull request adds direct links to several of the most commonly used binaries to the front page of the html documentation, eg:

http://superdarn.thayer.dartmouth.edu/WG-sched/assets/doc_enhancement/index.html

Hopefully this change coupled with the RST for Dummies guide will make our users' lives easier.


mts299:
@egthomas In the telecon, we also talked about readthedocs also in issue #210, do we not want to start focusing our energies there? 

This is what I have gotten so far: 
https://radar-software-toolkit-rst.readthedocs.io/en/latest/

@ecbland has already done some work to convert the RST to dummies into markdown so I can add it to the readthedocs. 
When `install.md` gets created, I can also reference that file from the readthedocs platform and include it.

egthomas:
@mts299 given the pace at which changes have been approved and implemented since the 4.2 release, this seemed like an easy way to address a few of the concerns raised during the telecon for at least the short term.  If we hold off on another release until an entirely new documentation system is ready then we will likely violate our release guidelines by at least 6-12 months.

ksterne:
I can understand wanting to focus on more updated documentation style, but also agree here that we should correct/update the content of our current documentation while we're getting the new documentation style squared away.

I can test this once I get access to a good VM, but if someone else beats me to the punch I won't be sad.  I'll mark myself as a reviewer for now.

mts299:
I looked at this last week with a new co-worker and summer student discussing documentation styles how they felt about the documentation. 

All the links work so it can be merged.
My only two comments would be:
- `time_plot` plots rang-time parameter plots as well so maybe add that into the description as it is a more commonly known term
- Do we want to add `combine_grd` as it is needed to generate map files? We have all the other binaries needed :p 

Otherwise, everything else looks good for now.

ksterne:
Looks like there's a piggy back commit onto this with bd53a67; @egthomas, is this commit OK here?  Looks like it's pretty simple addition to documentation here.  Maybe this is something I missed at the workshop.

mtwalach:
Hi @ksterne and @egthomas ! 
Yes, I thought it (https://github.com/SuperDARN/rst/commit/bd53a67a4b018184c38b1cc365dac3cd1daac168) fitted the theme of the PR and only adds a tad more information on the fitting algorithm and probably doesn't warrant its own PR as a result. Is it in the wrong place?

ksterne:
@mtwalach, it doesn't change anything operationally/functionally.  But @ecbland and I are trying to keep pull requests more on point and not scope creeping as much.  It's really OK here as long as it's not a problem with @egthomas who originated the PR.

egthomas:
I don't have any problem with the comment about the HMB latmin behavior being added to the `map_addhmb` documentation in this pull request, but this portion could possibly be clarified:

> ... the median filtered boundary is moved 1 degree equatorward of the boundary found by the algorithm when the options "<code>-vel</code>" and "<code>-cnt</code>" are used.

I belive this 1 degree equatorward offset is **always** applied unless a user-specified `latmin` is supplied - not only when the `-vel` and -`cnt` options are used.  Is that correct @mtwalach?

mtwalach:
Hmm, that's a good question! I will have to check that because I have only ever used `-vel` and `-cnt`...

ksterne:
@mtwalach, any update here?  If not then I'd say lets revert that commit as I don't want this PR to be waiting on that.

asreimer:
From what I can tell reading [map_addhmb.c](https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c) @egthomas is correct. `-vel` and `-cnt` always have a -1 latitude applied to the boundary.

`-lat` sets `hmblat`, and [only affects the determination of the boundary with this half of this if statement](https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c#L208-L252). Addtionally, in this part of the if statement, `-lf` sets `lname` which tries to load a file with latitude limits in it. [Whereas `-vel` and `-cnt` only affect the `else` portion of the same if statement](https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c#L252-L372). Digression: Trying to follow which brackets are open/closed makes you go cross-eyed..

[And focusing on this part of the `else`](https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c#L308-L311), the boundary is moved -1 degree always, as long as neither `-lf` nor `-lat` are used. [Notice that this particular if statement is always false](https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c#L310).

In summary, the documentation should state that the median filtering boundary determination is only done when `-lf` or `-lat` command line options are not used (kind of does right now if https://github.com/SuperDARN/rst/commit/bd53a67a4b018184c38b1cc365dac3cd1daac168 is reverted). So I think we need to revert https://github.com/SuperDARN/rst/commit/bd53a67a4b018184c38b1cc365dac3cd1daac168 (sorry @mtwalach). But, a separate PR could easily be opened that cleans up the `map_addhmb.c` docs and code in the following ways:
1) clarify that `-vel` and `-cnt`, do nothing if either `-lf` or `-lat` are used
2) clarify that `-lf` takes precedence over `-lat` (if `-lf` is specified, `-lat` does nothing)
3) refactor `map_addhmb.c` so that it is readable (consistent brackets and spacing/nesting)
4) remove [this if statement](https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c#L310)

ecbland:
Can we please revert @mtwalach 's commit and merge this?

ksterne:
@mtwalach, I've reverted the commit for now until it can be cleared up what the proper content should be for updating the documentation.  This can be a different PR as we've noted here.  Merging this as things look good.

PR #229, opened 2019-05-30
radar.dat update

egthomas:
This pull request returns the header information previously included at the top of the `radar.dat` file in RST versions [3.3](https://github.com/SuperDARN/rst-archive/blob/rst.3.3/tables/superdarn/radar.dat) to [3.5](https://github.com/SuperDARN/rst-archive/blob/rst.3.5/tables/superdarn/radar.dat) but removed at some point in [VTRST3.5](https://github.com/SuperDARN/rst-archive/blob/vtrst.3.5/tables/superdarn/radar.dat).  No other changes are made to the contents of the `radar.dat` file.

### DIFF
--- a/tables/superdarn/radar.dat
+++ b/tables/superdarn/radar.dat
@@ -1,3 +1,35 @@
+# radar.dat
+# =========
+#
+# This file contains status information for each radar in the SuperDARN network.
+# Radar information is stored, one radar per line, as a series of white space separated fields.
+# Text fields are quoted (") to allow spaces. The definiton of those fields are as follows:
+#
+# 01  Station ID Number  Unique numerical radar identifier.
+#                        Zero (0) is reserved for a "test" radar. Other numbers are
+#                        allocated on request.
+#
+# 02  Status             Current status of the radar. Possible values are:
+#                        (0)     Offline, or under test
+#                        (1)     Operational
+#                        (-1)    Decommissioned
+#
+# 03  Start Date         First date (YYYYMMDD) that data is available from the radar.
+#
+# 04  End Date           Last date (YYYYMMDD) that data is available from the radar. For
+#                        operational radars, this is set to some future date.
+#
+# 05  Station Name       Full name of the radar site.
+#
+# 06  Operator           Operator, or PI instituition of the radar.
+#
+# 07  Hardware Filename  Filename of the text file containing radar hardware configuration.
+#
+# 08+ Station ID Codes   Station identifiers. Radars are usually identified by a three letter
+#                        code. Older radars also have a single letter code letter. Multiple
+#                        identifiers can be defined and they can be of any length to allow
+#                        short nicknames for a given radar.
+
 0   0 20080203 25000101 "Test Radar"              "United Nations of SuperDARN"         "hdw.dat.tst" "tst" "x"
 1   1 19831001 25000101 "Goose Bay"               "Virginia Tech"                       "hdw.dat.gbr" "gbr" "g"
 2  -1 19861001 19950930 "Schefferville"           "CNRS/LPC2E"                          "hdw.dat.sch" "sch" "s"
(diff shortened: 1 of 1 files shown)


ksterne:
Is it worth it to do some additional documentation info here?  Reviewing things, the only thing I could see is adding to the Station ID codes on a note about how the newer radars use "i" and "z" as placeholders for the single radar code to be backwards compatible.  I get a lot of questions from new users on why this is.  I can add in a commit if needed/desired (though not likely this week).

egthomas:
@ksterne should we just change the single letter code for all of the newer radars (ie those without dat files) to `"z"` to maintain consistency?  I think WAL was intentionally set to `"i"` but that was later incorrectly applied to a few of the newer radars.

egthomas:
Here's another question (which I never received a reply from Mark about after several years of trying) - should we remove the EKB radar info and hdw.dat file?  It's not officially part of the SD collaboration.

mts299:
I heard from Kathryn that Russia is trying to work something out to be part of SuperDARN. Not sure if it is a really a big deal to delete the file or not, but I don't see any harm in just it in there for now.

ecbland:
@ksterne @egthomas : a brief explanation about the use of `z` and `i` for the newer radars without a single-letter station code is a good idea. Otherwise this is ready to merge. I think we should deal with the EKB radar question separately (I agree that it's harmless to leave it there).

egthomas:
Is a single letter code even necessary for the newer radars?  Maybe Wallops, McMurdo, Zhongshan, and Falkland Islands can keep "i", "m", "z", and "q", but the other newer radars seem to have just copied "i" or "z".  We may be able to just drop them entirely and the various functions accessing the `radar.dat` file should still work.

ecbland:
@egthomas 

> Is a single letter code even necessary for the newer radars? ...We may be able to just drop them entirely and the various functions accessing the radar.dat file should still work.

After some detective work, it looks like the single-letter codes for the newer radars are actually detrimental to the functionality of some RST routines. The routine `RadarGetCodeNum` counts the number of identifier codes for a particular radar (e.g. `sas` and `t` for Saskatoon), and then any of these identifier codes can be used to specify that radar in RST routines such as `fov_plot`. 

Because the 1-letter codes for the newer radars are not unique, we get some interesting behaviour for specific RST routines. Here are two examples: 

**(1) `radar_id` -- prints basic information about a radar, read from `radar.dat` and/or the hardware file**
```
$ radar_id -vb -n i
name:"Buckland Park"
```
_In this case, `radar_id` returns the FIRST radar listed in `radar.dat` with identifier code `i`. Same behaviour for `z`, which returns info for Zhongshan_

**(2) `fov_plot` -- plots the radar field of view**
```
fov_plot -x -stereo -coast -fov -lat -90 -latmin 20 -st i
```
_This plots the FOV for the Buckland Park radar._ 

This cannot be the intended behaviour. Removing the 1-letter codes for the new radars would therefore improve the functionality of RST, in addition to removing unnecessary and confusing information. 

Perhaps we should check that other software packages which use `radar.dat` are not going to crash if there is no 1-letter code for some radars.

ksterne:
I think the question of whether or not the single letter radar code is needed here gets out of the scope of the original intent of this pull request.  I'd say add in the description for the single letter radar code and this is ready to be merged.  @egthomas, would you like to do this, or can I add something in here?

egthomas:
If the single letter codes for the new radars are removed then it doesn't seem like the header information for that field needs to be changed:

> Station identifiers. Radars are usually identified by a three letter code. Older radars also have a single letter code letter.

What more would need to be said?

sshepherd:
Probably too verbose, but "Early SuperDARN radars were originally
identified by a single letter code. However, when it became evident that
the number of radars would surpass the number of unique possibilities,
radars were given three letter codes. All radars now have unique three
letter codes and single letter codes are only retained for backward
compatibility."

Issue #231, opened 2019-06-17
Possible bug in rst idl library: superdarn/main.1.25/rprm.pro


agrocott:
I just tried to load in a random fitacf file: 20020202.0201.00.han.fitacf and the code crashed:

IDL: Conflicting data structures: structure tag,<INT       Array[24, 2]>.
IDL: Execution halted at: RADARDECODERADARPRM  275 /Applications/superdarn/rst-develop/idl/lib/superdarn/main.1.25/rprm.pro

Looking at the line in question:
```

  if (prm.mplgs gt 0) && (arrid[1] ne -1) then $
     prm.lag[0:prm.mplgs,*]=(*(arrvec[arrid[1]].ptr))[*,*]

```

it seemed as though the data range being specified is too long, i.e. it ought to be:

```

 if (prm.mplgs gt 0) && (arrid[1] ne -1) then $
     prm.lag[0:prm.mplgs-1,*]=(*(arrvec[arrid[1]].ptr))[*,*]

```
This fixed the issue for this file, however, testing this fix on another file (20120202.0201.00.sas.fitacf) caused the opposite problem, of the data range being too short. So, apparently not a fix, but also a problem with consistency!

I note this is related to pull request #121 - I am running 4.2 and seem to have the new version of the code detailed in this request.

Thanks,
Ade

egthomas:
@agrocott thanks for the message.  It seems like while the native IDL code does crash on that file, the C version of the code used by the DLMs appears to handle things correctly.

Examining that particular file more closely with `dmapdump`, the number of lags listed in the parameter block (`mplgs`) does not match the dimensions of the lag table array (`ltab`):

```
	short	"mppul" = 7
	short	"mplgs" = 18
        ...
arrays:
	short	"ptab" [7]
	short	"ltab" [19][2]
```

egthomas:
Actually, the error I encountered was because I tried calling the DLM and IDL versions consecutively and forgot to free the `fit` structure in between.  Just using the native IDL code to read the `fitacf` file we created for that day, eg

```
fitfp = FitOpenIDL('20020202.han.fitacf', /read)
while FitReadIDL(fitfp, prm, fit) ne -1 do begin
;    print, prm.time.hr, prm.time.mt, prm.time.sc
endwhile
s = FitCloseIDL(fitfp)

end
```

seems to work fine (where I've renamed each function in `fit.pro` to ensure the native IDL version is being called).  Likewise, reading the original dat file also seems to work for me:

```
[...]

PR #232, opened 2019-06-17
IDL and DLM bug fixes

egthomas:
This was identified while testing #231 but is unrelated to the actual issue raised by @agrocott.  The `FitClose` function was already written in `fitdlm.c`, however it was not included in the list of available functions in `fitdlm.dlm`.  Various other `...Close` functions were already included in all of the other superdarn DLM code.


egthomas:
While testing to make sure the removal of the single letter station codes in #229 did not break any of the native IDL or DLM functionality, I noticed a couple of other bugs.  First, the `cnum` value in the radar structure was not being set in the native IDL code, making it impossible to track the number of station codes loaded into the radar structure.  Second, the DLM versions of the various `Radar` functions were not setting the `atten` value from the `hdw.dat` files (while the native IDL code handled it correctly).

ecbland:
Testing on this branch:
1. The correct `atten` value is loaded when using the DLM
2.  The correct `cnum` value is loaded when using the native IDL code
3. `FitClose()` now works when using the DLM

Does this need any other testing?

egthomas:
@ecbland no that should cover everything. Thanks!

Issue #214, opened 2019-01-25
time plot auto-adjusting time domain 


mts299:
While testing #209 I noticed the time domain does not auto adjust for a given file. So when I input a 2-hour file it gives me a 24 hr plot: 
![time_3_0_sas](https://user-images.githubusercontent.com/6520530/51768606-61dc5380-20a6-11e9-80fd-d9e4d74b9803.png)

A possible enhancement is for it to adjust the time scale based on the start time and end time of the file.

asreimer:
Addressed by #235

Issue #233, opened 2019-06-20
Comparisons between pointers and integers introduced by #165


egthomas:
Pull request #165 introduced 11 instances of comparisons between a pointer and integer with code similar to 

```
} else if (rawfp->rawread==-2) {
```

These can be found in the following binaries:

- dattorawacf
- test_raw
- cat_raw
- make_fit
- make_fitex
- make_fitex2
- make_info
- make_lmfit
- raw_cp
- rawstat
- trim_raw

These were noted in [Dec 2018](https://github.com/SuperDARN/rst/pull/165#issuecomment-444288336) and [Mar 2019](https://github.com/SuperDARN/rst/pull/165#issuecomment-475436174).  It would be nice if these were addressed, particularly since the compiler warning for `trim_raw` is the last thing someone sees when compiling the RST without IDL.

egthomas:
So since it has been removed from the project, is this not going to be addressed before the 4.3 release? It doesn't seem ideal to distribute code with known problems that we as developers introduced.

ecbland:
This looks like something that can be fixed fairly easily in time for the v4.3 release, and I agree that we should avoid distributing code with known problems. Do you have time to implement the required changes @egthomas ?

ksterne:
@egthomas, @ecbland, I initially removed this from the project since there wasn't an open pull request here.  I think until I see a pull request on it, this could potentially slow the project down.  I understand the hesitation here since the warning is the last thing that could be seen while compiling.  However, the warnings do not cause compilation to fail and do not seem to have an effect on data quality.  So, I evaluated this as something that could take a while to resolve and doesn't seem to be critical.

As we noted in #165, there's a bigger issue here of exit/return codes for error handling that doesn't seem to be well handled.  From what I could see, the solution to this issue (233) will need some significant rewriting of the code in order to properly set the rawread variable and return an error code that handles things properly.  Diving into this isn't something I can handle here while trying to get months old code into a new release.  Maybe there's a different solution I'm not seeing that could be easier though.

So, I think there's two solutions here:
1. We wait for a pull request to be generated from this issue that is tested and accepted/merged before starting the 4.3 release process.
2. We continue on with the 4.3 release while this issue is still being worked on and once it is resolved, we can have a 4.3.1 release that can include this.

Sorry, I know some of this is out of scope for the issue and should be more so in #237, but there's some detail on potential solutions as well.  

Let me know thoughts on which would be preferred; I'll go with the group even though I recommend the later solution.

[...]

PR #188, opened 2018-10-09
Codacy test

egthomas:
This pull request addresses a mix of memory leaks, resource leaks, and other potential errors identified when I was testing Codacy a while back (https://app.codacy.com/project/egthomas/rst/dashboard).  The number of issues listed on their website is buggy (2-3x the actual number) but the individual issues found under the various filters on the "Issues" page should still be accurate when comparing this branch to the develop branch.  

Moving forward I don't plan to do much more with this but am happy to add any future repositories created under the SD organization to Codacy if requested.

PR #235, opened 2019-06-20
Auto-adjust time domain in time_plot

ecbland:
By default, `time_plot` produces plots with a 24-hour range, regardless of the actual time range of the input file (#214). This pull request allows the time range to be determined automatically from the input file, unless the user specifies the end time (`-et`) or extent (`-ex`) of the plot. I have implemented this only for the FIT/FITACF format files. CFIT and SMR are a mystery to me ;)

I hope my implementation is not too much of a hack. I used `FitFread()` to loop through to the last record in the file, and then calculate the end time from the last value of `prm->time`. It is then necessary to use `FitSeek()` to jump back to the start time before the data are read in "for real". 

`time_plot -png -a -b 7 20170505.0601.00.sas.fitacf` now produces this: 

tplot

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
@@ -1086,7 +1086,8 @@ int main(int argc,char *argv[]) {
   if (etime !=-1) {
     if (edate==-1) etime+=stime - ( (int) stime % (24*3600));
     else etime+=edate;
-  } else {
+  } else if (extime !=0) etime=stime+extime;
+  else {
    /* determine end time from the last record in the file
       if the total time range is less than 10min then set it to 10min.
       this has been implemented only for fit/fitacf format files */
@@ -1108,7 +1109,7 @@ int main(int argc,char *argv[]) {
   if (etime-stime<10*60) etime=stime+10*60;
   }
 
-  if (extime !=0) etime=stime+extime;
+  
 
   if (name==NULL) name=dname;
 
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
@@ -1082,11 +1082,35 @@ int main(int argc,char *argv[]) {
     stime=cfit_find(cfitfp,cfit,sdate,stime);
   }
 
+  
+
 
   if (etime !=-1) {
     if (edate==-1) etime+=stime - ( (int) stime % (24*3600));
     else etime+=edate;
-  } else etime=stime+24*3600;
+  } else {
+   /* determine end time from the last record in the file
+      (then navigate back to the start time) 
+      this has been implemented only for fit/fitacf format files */
+    int status=0;
+    if (fitflg && old) {
+       while (status!=-1) status=OldFitRead(oldfitfp,prm,fit);
+       etime=TimeYMDHMSToEpoch(prm->time.yr,prm->time.mo,prm->time.dy,
+                               prm->time.hr,prm->time.mt,
+                               prm->time.sc+prm->time.us/1.0e6);
+       TimeEpochToYMDHMS(stime,&yr,&mo,&dy,&hr,&mt,&sc);
+       status=OldFitSeek(oldfitfp,yr,mo,dy,hr,mt,0,NULL);
+     } else if (fitflg) {
+       while (status!=-1) status=FitFread(fitfp,prm,fit);
+       etime=TimeYMDHMSToEpoch(prm->time.yr,prm->time.mo,prm->time.dy,
+                               prm->time.hr,prm->time.mt,
+                               prm->time.sc+prm->time.us/1.0e6);
+       TimeEpochToYMDHMS(stime,&yr,&mo,&dy,&hr,&mt,&sc);
+       status=FitFseek(fitfp,yr,mo,dy,hr,mt,0,NULL,inx);
+
+     } else etime=stime+24*3600; /* cfit or smr format: default 24 hour */
+  }
+
   if (extime !=0) etime=stime+extime;
 
   if (name==NULL) name=dname;
@@ -1509,6 +1533,7 @@ int main(int argc,char *argv[]) {
     }
 
   } while (atime<etime);
+
 
   if (fitflg){
     if (old) OldFitClose(oldfitfp);
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
@@ -1083,14 +1083,13 @@ int main(int argc,char *argv[]) {
   }
 
   
-
-
   if (etime !=-1) {
     if (edate==-1) etime+=stime - ( (int) stime % (24*3600));
     else etime+=edate;
   } else {
    /* determine end time from the last record in the file
-      (then navigate back to the start time) 
+      (then navigate back to the start time)
+      if the total time range is less than 10min then set it to 10min.
       this has been implemented only for fit/fitacf format files */
     int status=0;
     if (fitflg && old) {
@@ -1109,6 +1108,7 @@ int main(int argc,char *argv[]) {
        status=FitFseek(fitfp,yr,mo,dy,hr,mt,0,NULL,inx);
 
      } else etime=stime+24*3600; /* cfit or smr format: default 24 hour */
+  if (etime-stime<10*60) etime=stime+10*60;
   }
 
   if (extime !=0) etime=stime+extime;
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
@@ -1088,7 +1088,6 @@ int main(int argc,char *argv[]) {
     else etime+=edate;
   } else {
    /* determine end time from the last record in the file
-      (then navigate back to the start time)
       if the total time range is less than 10min then set it to 10min.
       this has been implemented only for fit/fitacf format files */
     int status=0;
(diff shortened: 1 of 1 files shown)


egthomas:
@ecbland is it worth implementing a "minimum" auto-adjusted duration (eg, 10 minutes)?  I can imagine a scenario where someone tries plotting data from a file with only one / a few records and the x-axis becomes extremely distorted.

egthomas:
I haven't tested the end time functionality yet, but I did notice the documentation for `time_plot` is messed up for the min/max options, eg

```
-pmin pmin               set the minimum value of the power scale to <a>pmin.
-pmax pmax               set the maximum value of the power scale to <a>pmax.
-vmin vmin               set the minimum value of the velocity scale to <a>vmin.
-vmax vmax               set the maximum value of the velocity scale to <a>vmax.
```

It looks like there was a typo in the original `time_plot` documentation (pre-RST4.0) which I've continued propagating as more parameters were added.  Lines like this:

```
<option> ... power scale to <a>pmin</ar>.</od>
```

should be this:

```
<option> ... power scale to <ar>pmin</ar>.</od>
```

ie leading the input option with `<ar>` instead of `<a>`

ecbland:
@egthomas I think the documentation errors should be addressed in a separate pull request to prevent scope creep.

I found a small bug in my code so please wait before you test it (won't be long). 

I actually have some concerns about implementing this auto-adjust functionality because there is a speed penalty. On my machine it takes ~5 seconds to produce a 24 hour plot instead of ~3.5s previously. As an RST user I am nearly always using 24-hour files as input to `time_plot`...if most other users are like me then the benefit of the auto-adjust functionality is questionable. Comments on this are welcome.

ecbland:
ok, this is now ready for testing

sshepherd:
I do not use time_plot but was going to raise the concern that reading
through the file twice must exact some sort of performance penalty. Is the
auto adjust invoked by default?
There are command line options to adjust the plot, correct? I, personally,
would think that is sufficient. If you get a plot you don't like then
adjust it.

ecbland:
@sshepherd  The auto-adjust will kick in unless the user provides either `-et` (end time) or `-ex` (extent). I suppose I could introduce an optional `-autoscale` input parameter and keep the 24-hour timespan as the default, but then I wonder whether anyone would use it. Normally I know the time extent of my file before I plot it :)

This functionality seemed like a good idea before I implemented it, but now I'm not so sure. I can't see a way to avoid the time penalty without a major restructuring of the code.

mts299:
I think this is a better use case for users who do not know our data file structure. I was one of these user's and wasn't as aware of using `-et` `-st` options; thus, why I asked for the enhancement.  

Is the time performance a huge concern when these plots are only supposed to be used for "quick look " plots? 

Also, how much of a hit is it?

ecbland:
@mts299 That's a good point re: non-expert users. We now have a tutorial for `time_plot` (https://radar-software-toolkit-rst.readthedocs.io/en/latest/user_guide/time_plot/) in addition to the usual documentation. The default time axis range is one of the first things that's mentioned under "getting started", so perhaps that is sufficient to help the non-expert?

People will rarely plot more than 24 hours of data at a time. On my machine this takes about 3.5s. With the new code it takes 5s. As a percentage it's quite a big increase. Since `time_plot` is designed for the impatient user wanting a quick look at the data, this is a lot of extra waiting time. 

I really thought this was a good idea until I saw the speed penalty. If there is a way to avoid it then that would be great.

asreimer:
Testing this, I grabbed a full 24 hr file from maxwell at the U of S, specifically: `20170505.C0.sas.fitacf`
Here are the results:

    master takes ~2.9 seconds
    this PR takes ~3.8 seconds

So the performance hit is less than 1 second for a 24 hour file or ~30% increase in time required to plot. This is similar to what @ecbland get in their testing. In my opinion, this is well worth it for the convenience. The speed hit is really not a big deal at all if you think about the use cases for this tool. Honestly, if the tool isn't convenient, no one is going to use it.

If people need the speed, they can simply specify the end time with the `-et` option. I think this should be merged in.


This is the plot that was made in both the `master` and `timeplot_time_domain` branches.
plot

I used the command:

    $ time time_plot -png -a -b 7 20170505.C0.sas.fitacf > plot.png
    real	0m3.252s
    user	0m2.883s
    sys	0m0.347s

and ran it several times. Note that the "user" time is what matters most.

ecbland:
@asreimer That's how I tested the speed too. 

I actually don't object to this being merged...I just wanted to highlight the potential issue. It took me by surprise that `FitSeek()` is actually pretty slow :)

asreimer:
Consulting with @mts299 and because there is no performance hit if the `-et` option is used, I'm merging this.

We just need to note the default behaviour change in the release notes. 

Again, no one is going to use these tools if they are not easy to use.

egthomas:
> I think this is a better use case for users who do not know our data file structure. I was one of these user's and wasn't as aware of using -et -st options; thus, why I asked for the enhancement.

As @ecbland mentioned, there is now an RST for Dummies guide which does / should cover this, there are several examples in the [documentation](https://superdarn.github.io/rst/superdarn/src.bin/tk/plot/time_plot/index.html), and there is command-line help which explains all of the available options.  So I don't think the time penalty should be so easily discarded in favor of "usability" if the users ignore the myriad of documentation.

asreimer:
@egthomas that's not the issue here at all. Sure, there's documentation, but the **expected** behaviour for someone using `time_plot` is that it will plot the whole file and have the time axis cover the contents in the file.

This PR fixes the violation of the [Principle of Least Astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and does so without crippling the performance for those who might want it for making batch plots with a script.

sshepherd:
I think that speed should always be considered when making changes to the
software. 30% is a big number, particularly for something that can be
accomplished without making this modification.
In terms of the "user", if someone tries to make a plot for <24 hours and
sees the 24 hour scale, they should look at the documentation.

mts299:
Should we make an issue on FitSeek performance? 

I feel scope creep happening :p 

Isn't inconvenience a performance hit to the user?

asreimer:
Last thing I'll say is that there's a principle in computer science that says "Don't prematurely optimize".

Speed for speed sake isn't necessarily worth it. As @ecbland mentioned, `FitFSeek` is surprisingly slow. I suggest we focus energy investigating optimizing `FitFSeek` instead.

egthomas:
I'm not going to drag this out any further, and just close by thanking @ecbland for implementing this requested feature in the code and hope that in the future we allow more than 15 minutes from a raised concern to approved / deleted branch, particularly given how the DAWG membership spans multiple time zones.

sshepherd:
The limitation is that you are looking through the file to find the last
record so that you can determine the time range of the data. You must then
read through the data a second time and do the plotting.
You can't get around the fact that this will be slower and negatively
impact some users. The documentation should clearly state how to change the
plotting axes (which I think it does) and users should
be expected to be able to read documentation if they don't like what they
see.

sshepherd:
Agreed. It does seem like this was approved in the middle of a discussion
and there may have been opinions that extended even further.

ecbland:
Yes, I think this was merged prematurely (but I love the enthusiasm). Let's leave it merged for now, and I will reconsider my position on this over the weekend.

asreimer:
@ecbland I'm genuinely confused that people are upset this was merged.

1) This PR doesn't break performance since `-et HH:MM` retains the same functionality/performance as before the change, so power users who read the documentation still get their performance
2) It implements a feature that was requested
3) It brings the code inline with what users intuitively expect
4) It is a small change that doesn't break anything upstream (because there is nothing upstream in RST
5) Performance change isn't crippling and is therefore arguably out of scope (and we are supposed to be keeping scope narrow)

So what concerns are there that require delaying this in lieu of discussion? I'm sorry that I pre-maturely merged this, but I genuinely don't understand what the problem is.

If it really is a big deal, can we please publish some guidelines for contributing? And the nice thing about git is we can always revert.

ecbland:
@asreimer My gut feeling is that this was a good addition to RST; it just happened before the discussion had reached an equilibrium. 

Let's not turn this into a big deal. We've had a really productive week with RST and I think this is a reflection of some renewed enthusiasm.

asreimer:
Ok, sounds good to me :-)

egthomas:
I was just fixing the merge conflict on the `codacy_test` branch in #188, and noticed there were a couple of warnings/errors introduced in `time_plot.c` with this pull request.  First is an implicit declaration of `FitFseek`, which can be addressed by including `fitseek.h` near the beginning of `time_plot.c`.  The other is that `yr`, `mo`, `dy`, `hr`, `mt`, and `sc` are used uninitialized at line 1104 in the call to FitFseek:

```
     } else if (fitflg) {
       double atme;
       status=FitFseek(fitfp,yr+1,mo,dy,hr,mt,0,&atme,inx);
       etime=atme;
       TimeEpochToYMDHMS(stime,&yr,&mo,&dy,&hr,&mt,&sc);
       status=FitFseek(fitfp,yr,mo,dy,hr,mt,0,NULL,inx);
     } else etime=stime+24*3600; /* cfit or smr format: default 24 hour */
```

I fear this could cause unexpected behavior when calculating the `tval` epoch time for which to search for in the fitacf file.  @ecbland, what are your thoughts here?

ecbland:
@egthomas This should be easy to fix. Should we make a separate PR, or add another commit to #188 ?

egthomas:
@ecbland you can add the changes to the `bug_fixes` branch I recently created (which adds some command line options I forgot to include in the `time_plot` parameter updates).  @ksterne could also include his fixes for #233 / #165 on that branch.

ecbland:
@egthomas 

> First is an implicit declaration of FitFseek, which can be addressed by including fitseek.h near the beginning of time_plot.c. 

Looks like you have already fixed this on `bug fixes`

> The other is that yr, mo, dy, hr, mt, and sc are used uninitialized at line 1104 in the call to FitFseek

Have I missed something here? These were initialised on lines 467-468 (on `develop`)

egthomas:
Yes, they are declared as ints / doubles there but have not been assigned any values before the call to FitFseek. So you may just need to change the arguments of FitFseek to something like `prm->time.yr` etc, similar to the earlier OldFitSeek call for the old file format case.

ecbland:
@egthomas Ahhh, I see what you mean now...oops! I can fix that.

PR #239, opened 2019-06-24
Comissioning data error

mts299:
While processing all of the SupderDARN rawacf data to Fitacf 3.0 for our local server machine, I found Buckland Park (bpk) keeps throwing fitacf errors like this:
```
make_fit 20140426.0002.00.bpk.rawacf > 20140426.0002.00.bpk.fitacf
Failed to get site information.
make_fit 20140426.0202.00.bpk.rawacf > 20140426.0202.00.bpk.fitacf 
Failed to get site information.
```
This problem was similar to when I was processing some Rankin Inlet data for the year 2006 and soon realized it is an issue to do with commissioning dates in [radar.dat](https://github.com/SuperDARN/rst/blob/a3858d423ef370cda686471bada1d95f5319040b/tables/superdarn/radar.dat)
However, with further investigation, I found that **20140426.0002.00.bpk.rawacf** was empty and the other one was not. So the errors were a bit miss leading to what was wrong with the files. Also, @kevinkrieger told me the DDWG policy for the commission data is based on if the PI's wanting it in the data set or not (blacklisted). I will be contacting Australia's PI about the bpk files; however, this does not address how DAWG/RST handles the issue. 

This PR's scope focuses on introducing 2 new error checks:
1. Error: File may be empty 
2. Warning: this file is commissioning data  

I left 2. as a warning because users may still want to use the data. 

Test:
```
make_fit 20140426.0002.00.bpk.rawacf > 20140426.0002.00.bpk.fitacf
Error: Could not read the contents of the file, may be empty
make_fit 20140426.0202.00.bpk.rawacf > 20140426.0202.00.bpk.fitacf 
Warning: Data is before commissioning date set in the radar.dat file
```

Tested the fitacf file by quickly plotting it:
`time_plot -png -a 20140426.0202.00.bpk.fitacf > 20140426.0202.00.bpk.time_plot.png`
20140426 0202 00 bpk time_plot

This PR is to address the commissioning dates issue and empty files issue that is unclear from the "Failed to get site information," which made me think it was a hardware file problem. 

This PR is not to address why bpk is in superDARN Canada set and whether these files should be blacklisted or not that is under the DDWG  control.


mts299:
Status update: I emailed John Devlin about the bpk data.

codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
```diff
@@ -205,13 +205,20 @@ int main(int argc,char *argv[]) {
     }
     status=RawFread(fp,prm,raw);
   }
+  
+  if (status == -1)
+  {
+      fprintf(stderr, "Error: Could not read the contents of the file, may be empty\n");
+      /* Using -2 to simbolize reading error of the file*/
```
asreimer: Spelling: symbolize

asreimer:
This looks pretty good. There's a subtle thing included here that might need some discussion down the road, but I don't think it should hold this PR up. I've made a new issue instead (#240). If it becomes a scope issue, I would recommend changing the `exit(-2)` line from this commit to be `exit(-1)` like the rest of the exit codes.

mts299:
@asreimer switched the exit value but made a comment on which error code it should be. 
Also, added my plotting line to make things easier to test.

egthomas:
I get this message when trying to unzip the first BPK rawacf file:

```
bunzip2 20140426.0002.00.bpk.rawacf.bz2 

bunzip2: Compressed file ends unexpectedly;
	perhaps it is corrupted?  *Possible* reason follows.
bunzip2: No such file or directory
	Input file = 20140426.0002.00.bpk.rawacf.bz2, output file = 20140426.0002.00.bpk.rawacf

It is possible that the compressed file(s) have become corrupted.
You can use the -tvv option to test integrity of such files.

You can use the `bzip2recover' program to attempt to recover
data from undamaged sections of corrupted files.

bunzip2: Deleting output file 20140426.0002.00.bpk.rawacf, if it exists.
```

so that file seems like it may be corrupted and needs to be blacklisted.

Printing a warning message that a file is from before the official radar start date will be easily missed by users ripping through years of `dat` or `rawacf` files at once.  Maybe an option could allow the user to process the file anyways, but I don't think the default behavior should be changed to allow such files to be processed.

Also, this pull request has removed several of the updates to `make_fit` introduced by the `feature/elevation` branch (specifically the `-old_elev` option) - please make sure that is fixed before anyone merges this.

codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
```diff
@@ -71,7 +71,6 @@ int main(int argc,char *argv[]) {
   unsigned char version=0;
 
   unsigned char vb=0;
-  unsigned char old_elev=0;
```
asreimer: This and some other lines were not intended to be removed by the latest commit right? Just need to revert them.

asreimer:
@egthomas I've requested changes to hold the PR. Looks like @mts299 like the commit to change exit(-2) to exit(-1) came with some unintended side effects.

The problem with the `bpk` file is clearly stated in the bzip2 error message: `Compressed file ends unexpectedly;` This means is was truncated, meaning the file was never successfully and completely copied at some point. Did this occur during download or upload? In any case, whether or not this file should be blacklisted is entirely not the point of the PR and is a matter to be referred to DDWG.

Back to the issue that this PR is addressing. I think we all agree that RST should be able to process commissioning data. The question is now whether we should require a command line flag to do so.

I can be persuaded, but requiring a command line option to force process the data here isn't necessarily consistent with previous PRs handling issues with data files, such as #221? If DDWG is handling blacklisting of files, do we need to worry about adding a command line option to protect the user? If so, why aren't we protecting the user when there's a problem with a file like in #221?

codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
mts299: Yes I remerged the make_fit file back in but I need to make sure other files were not over written

egthomas:
I downloaded the `20140426.0002.00.bpk.rawacf.bz2` file from the BAS mirror rather than Globus and lo and behold it correctly unzipped and processed into a `fitacf` file - granted it only contained about 10 minutes of data, but still far from an empty file.  So maybe the integrity of this file on the USask mirror needs to be confirmed.

My feeling is that the start date in the `radar.dat` file should mean something.  In this case, the software failed to load the radar site information because there is no site information for that date - where is the inconsistency?  If the start date is incorrect, then it's up to the radar PI to update that information (just like with the hardware files).

Furthermore, if we change the behavior of `RadarEpochGetSite` to ignore the radar start date, then we're changing the behavior of a lot of other RST functionality beyond just `make_fit`.  Extensive testing is going to be required to examine the impact this will have on the dozens of other plotting and processing binaries which rely on this function.

If you want to allow the commissioning data to be processed, then sure, let's add that capability.  But it's dangerous to assume all users will want to use those files when previously the software rejected it for them, hence why I think it should be "opt-in" and not default behavior.

ecbland:
So far this discussion has only addressed _retrospective_ processing of commissioning phase data. How is the data from a newly-constructed radar supposed to be processed during the commissioning phase itself? In the past I've done this by manually entering an earlier start date in `radar.dat`. I realise this is only an issue for the PI/institution which owns the new radar, but thought it was worth mentioning. 

I agree with @egthomas that we need to be careful about how this change might affect other RST functions. But I would also point out that the PI should only be releasing commissioning phase data if they have determined that these data are of reasonable scientific quality, in which case _why does the user care if these data were recorded during the commissioning phase?_

I'm not advocating for any particular position here...just asking questions.

asreimer:
I'm only going to address concerns about how "dangerous" this kind of a change is. I think the rest of the discussion about how should maintain what hdw.dat files etc, is a blurry messy undefined out of scope beast. Until offline discussions happen and some kind of standards are written up and published it does not good to discuss that here. It would be helpful for the DDWG and DAWG to communicate about areas of overlap such as this (probably this is happening as we speak).

The change does not result in `RadarEpochGetSite` ignoring the start date in the hdw.dat.xxx file, it changes the behaviour from error to warning when the time in the data file is earlier than the start date in the hdw.dat.xxx file. 

`RadarEpochGetSite` is not really used anywhere except `RadarYMDHMSGetSite`, so if you want to look, you need to search for that. The direct consequence of the proposed change to `radar.c` seems to be:
1) [this if statement](https://github.com/SuperDARN/rst/pull/239/files#diff-e3c5af479929369d291c80b9d47fe266R46) no longer returns NULL, but prints a warning instead,
2) `RadarYMDHMSGetSite` will now return the first entry in the radar's hdw.dat (hdw.dat.xxx) if the `tval` is less than any `tval` in hdw.dat.xxx. So for @mts299's usage example, the first entry in hdw.dat.bpk is used when the time in a data file is less than that in hdw.dat.xxx, and
3) this means that if the commissioning needs different parameters than are in the first entry of the hdw.dat.xxx file, the resulting data could have problems.

Digging through all the files that use `RadarYMDHMSGetSite` there are common themes on how the returned site info is used (see more detailed listing below): some functions don't actually use the site information despite calling `RadarYMDHMSGetSite` (whoops), others use the site info for plotting purposes, and the rest use it for processing iqdat -> rawacf and rawacf -> some fitted data file.

None of these consequences are dangerous.

## Now that we know the consequences, how does this change the above listed code?
Here's a summary of the files that use `RadarYMDHMSGetSite` and the changes for them:

1) 'make_raw.c' - assume that the first entry in the hdw.dat.xxx file is valid for the iqdat file being processed
2) 'make_fitex2.c' - assume that the first entry in the hdw.dat.xxx file is valid for the rawacf file being processed (same as `make_fit`)
3) 'map_grd.c' - only uses `RadarYMDHMSGetSite` to know which hemisphere the radar is in
4) 'make_grid.c' - site is passed down to `GridTableAddBeam` which uses location and recrise, but recrise isn't used anywhere...
5) 'lagfr_fix.c' - site information isn't used even though it is queried...
6) 'make_fit.c' - assume that the first entry in the hdw.dat.xxx file is valid for the rawacf file being processed
7) 'radar_id.c' - this prints out site information, so it would have a warning printed and then print out the first entry in the hdw.dat.xxx file
8) 'make_lmfit.c' - assume that the first entry in the hdw.dat.xxx file is valid for the rawacf file being processed (same as `make_fit`)
9) 'rtgrid.c' - same thing as `make_grid.c`
10) 'dattorawacf.c' - site info is used to calculate `rprm->bmazm` which is written to rawacf file
11) 'fittoascii.c' - site information is read but not used, doesn't care if site=NULL or not. Wow.
12) 'fittofitacf.c' - site info is used to calculate `rprm->bmazm` which is written to rawacf file
13) 'fov_plot.1.10/make_fov.c' - site info is used for plotting
14) 'grid_plot.1.12/make_fov.c' - site info is used for plotting
15) 'field_plot.c' - site info used to detect hemisphere
16) 'fov_plot.c' - site info is used for plotting
17) 'fov_plot.1.10/make_fov.c' - site info is used for plotting
18) 'time_plot.c' - site info is used for plotting
19) 'make_fov.c' - for each radar, get site info then plot the radar FOV
20) 'rposdlm.c' - this is basically an IDL version of radar.c, so change needed to be consistent with radar.c change, then consequences for IDL that uses it will be the same as above

If you read to here, at least now that we are all on the same page and understand what exactly the affect of this PR would be on the rest of the code.

asreimer:
Given everything discussed so far:
1) do we close this PR because it is touching on a bigger issue than initially thought and we need to figure out what to do?, or
2) improve the current PR so that it makes a change to `rposdlm.c` so that the IDL stuff is consistent with changes in `radar.c` and document the resulting change in behaviour?

ecbland:
@asreimer - thanks for providing the rundown on which routines are affected by this change! 

I think this should be implemented, along with the equivalent changes to the IDL code (both native and DLM). If there are any unforeseen consequences then these should come out with proper testing of this PR. 

@ksterne any opinion about this PR from a data distribution point of view?

I have already identified one issue though: the warning message appears for any call to `fov_plot` if the chosen date is earlier than the commissioning date of the newest radar (LYR). This applies regardless of which radar FOVs are actually plotted. 
``` fov_plot -x -stereo -coast  -ffov -st sas -d 20161018``` - produces the warning
``` fov_plot -x -stereo -coast  -ffov -st sas -d 20161019``` - does not produce the warning

The warning text is also a bit weird here ("**Data** is before commissioning date") because `fov_plot` doesn't take any data as input.

**Comment by mts299** [comment]
@ecbland thank you for taking a look at this :) 

I can try to add the same changes to IDL, may need some help. 

The Warning/Error message can be changed too. I was just guessing the terminology but I feel having commissioning phase  or reference to radar.dat would be good to have. 

And I will look into the fov_plot warning problem tomorrow.

**Comment by egthomas** [comment]
> The change does not result in RadarEpochGetSite ignoring the start date in the hdw.dat.xxx file, it changes the behaviour from error to warning when the time in the data file is earlier than the start date in the hdw.dat.xxx file.

You're right, it ignores the start date in the `radar.dat` file, not the `hdw.dat` file.  And this is getting pedantic, but if the code previously returned a NULL at some point and stopped because the requested date was earlier than the official radar start date, but now the code proceeds beyond that point, how is the start date not being ignored?  Practically speaking?

> 'fov_plot.1.10/make_fov.c' - site info is used for plotting

```
fov_plot -png -ofov -stereo -mag -coast -fcoast -d 19930101 -grd -latmin 40 -fofov > fov_plot.png
```

`-d 19930101`

this_is_why_we_test

> If you read to here, at least now we are all on the same page and understand what exactly the affect of this PR would be on the rest of the code.

I tend to disagree that the larger effect on the rest of the codebase has been fully appreciated.

**Comment by asreimer** [comment]
Ah, yes, it's in the radar.dat file! My mistake. The time isn't ignored though, it's specifically checked, which is all I was trying to point out.

>I tend to disagree that the larger effect on the rest of the codebase has been fully appreciated.

What else is missing? @ecbland mentioned the problem with `fov_plot` and you provided another illustration of it. I mentioned that `fov_plot` is affected by this PR. What else are is specifically not being appreciated? General statements about the boogeyman looming aren't that helpful to resolving the problem.

File: codebase/superdarn/src.lib/tk/radar.1.22/src/radar.c [hunk 3]
```diff
@@ -43,7 +43,9 @@ struct RadarSite *RadarEpochGetSite(struct Radar *ptr,double tval) {
 
   /* If tval is before the radar.dat start time of the radar
      then return NULL */
-  if ((ptr->st_time !=-1) && (tval<ptr->st_time)) return NULL;
+  if ((ptr->st_time !=-1) && (tval<ptr->st_time)) {
+      fprintf(stderr,"Warning: Data is before commissioning date set in the radar.dat file\n");
```
asreimer: As @ecbland mentioned, this doesn't make much sense for things like `fov_plot`. This if is checking `tval`, so the time that was input, so a better message might reference time.

egthomas:
@ecbland pointed out there was a warning message regardless of which radars were being plotted, NOT that every radars' FOV will be plotted for any date regardless of their individual start dates.  Those are two very different problems.  I know this proposed change will also break my existing IDL code for plotting FOVs, which is borrowed from VT's Davit package (running their website), which was the original framework for Davitpy and was also based on Steve Milan's Go package used across our community.

One of the most impactful ways of illustrating the international, collaborative nature of SuperDARN to an external audience is to show a plot of the radar FOVs - particularly as a function of time as the network has evolved over both hemispheres.  Unintentionally breaking the software's ability to generate plots of radar FOVs which were active on a particular date should not be hand-waved away as a boogeyman.

I repeat that I am not opposed to allowing commissioning-phase data to be processed, which as @ecbland stated can potentially contain scientifically useful data.  My point is that the current proposed implementation has unintended consequences which should be addressed.

asreimer:
The Boogeyman comment was an attempt to solicit more explicit details. I
don't think we want to unintentionally or intentionally break anything. If
we could be clearer about concerns from the outset that would be great. It
sometimes feels like those in the know are withholding details that help
others understand their position, yet voice opposition without fully
explaining why. It's not the most constructive. Maybe we can figure out a
better way to communicate than on here because this really feels like a lot
of work for no gain.

Probably the simplest solution is to just update the radars.dat and hdw.dat
files in the end. Now that we have a clearer picture of the consequences we
can all more easily understand.

mts299:
So I am just going to close this PR, it is getting a bit overwhelming with the opinions and scope changes from blacklisting files to consistency amongst other error handling with no clear discussion or understanding what is going on. I guess this is one way to discourage development and productivity :( 

Here is a nice reading on code reviews and feedback, has a nice chart on comments and how to give feedback:  
https://www.alexandra-hill.com/2018/06/25/the-art-of-giving-and-receiving-code-reviews/

Thank you @ecbland for the suggestions and comments on `fov_plot`, when I attack this issue again I will look for a more elegant way to handle this error/warning so it is not "dangerous" to the RST existence. Maybe a review of the radar.dat file is needed and what it means to RST as @asreimer mentioned; however that is not what this PR was fully intending to address. And I have many other important things I can be working on, so I will just take a break from RST :) 

If you want my opinion just let me know :+1:

ecbland:
Reading through the comments here, I just wanted to offer the following:
1. Thank you @mts299 for pointing out a very real problem with how RST handles commissioning phase data, and for making a first attempt at fixing it. If this had been posted as an issue instead of a PR, I suspect we would have had a much longer discussion which may not have identified the specific problems relating to `fov_plot` and the existing IDL code. So @mts299 's work in implementing a possible solution probably saved everyone else a lot of time here. Again, thank you. 
2. Yes it is frustrating if someone's comment doesn't provide the information you need the first time. It often takes some time for everyone to understand a particular issue and its scope, so we should be careful to recognise these comments as such. If someone has not provided the information you need then ask again (and that appears to be what happened here). 
3. The link that @mts299 posted about giving feedback has lots of good advice.

kevinkrieger:
Sorry I'm late to the party. I agree that the link @mts299 posted contains a lot of good information, and should be seriously read by everyone here.

@ecbland to comment on your question about the DDWG policy on pre-commission period data: We will add it to the distribution if the PI asks us to do so. We will remove it from the distribution if the PI asks us to do so (example: lyr data before the commissioning date was blacklisted/removed from the distribution after we received word from the folks at UNIS :D). If, at some point in the future Dag decides he would like to have that data out in the wild, we will unblacklist it, and add it back into the distribution. 

Naturally, the radar.dat file should be updated based on the PI's wishes.

So I guess my opinion would be to have at the very least a warning if commissioning data is being used, I don't think my opinion holds much weight since I don't work on RST, but that's what I would do as a reasonable person.

PR #241, opened 2019-07-02
Fixing pointer comparisons

mts299:
As addressed in Issue #233 coming from PR #165, I created a hotfix to remove the warnings. 

This is not a complete solution nor it is best practices but good enough for now until we determine the error codes standard for RST.

 develop branch compilation:
`make.code |& grep "warning" | wc -l 
127`

`make_fit -old 2001102410w.dat w_fit
WARNING : rawopen : OldRawOpenFd : num_byte < 0 in record header, potentially corrupted file.`

fix/pointer_comparison 
`make.code |& grep "warning" | wc -l 
114`

`make_fit -old 2001102410w.dat w_fit
WARNING : rawopen : OldRawOpenFd : num_byte < 0 in record header, potentially corrupted file.`

### DIFF
--- a/codebase/superdarn/src.bin/tk/reformat/dattorawacf.1.8/dattorawacf.c
+++ b/codebase/superdarn/src.bin/tk/reformat/dattorawacf.1.8/dattorawacf.c
@@ -162,7 +162,7 @@ int main(int argc,char *argv[]) {
   if (rawfp==NULL) {
     fprintf(stderr,"File not found.\n");
     exit(-1);
-  } else if (rawfp->rawread==-2) {
+  } else if (rawfp->error==-2) {
     /* Error case where num_bytes < 0 */
     free(rawfp);
     exit(-1);
--- a/codebase/superdarn/src.bin/tk/testing/test_raw.1.12/test_raw.c
+++ b/codebase/superdarn/src.bin/tk/testing/test_raw.1.12/test_raw.c
@@ -116,7 +116,7 @@ int main (int argc,char *argv[]) {
       if (rawfp==NULL) {
         fprintf(stderr,"Could not open file %s.\n",argv[c]);
         continue;
-      } else if (rawfp->rawread==-2) {
+      } else if (rawfp->error==-2) {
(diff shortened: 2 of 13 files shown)


asreimer:
I see 185 warnings in develop and 173 warnings in this PR, but I'm using GCC 8.2.1, which is much more strict than older versions about stuff like indentation and string formatting for print statements. **But I explicitly checked the build log** and don't see the warnings for the oldraw stuff like I did in develop.

So I am confirming that this PR definitely fixes the relevant warnings introduced by #165. Noice.

I grabbed the `2001102410w.dat` file from globus, but it doesn't give me an error when executing:

    make_fit -old 2001102410w.dat w_fit

Here it is bzipped, remove the .txt from the filename. [2001102410w.dat.bz2.txt](https://github.com/SuperDARN/rst/files/3351806/2001102410w.dat.bz2.txt)

It seems the SuperDARN Canada globus mirror has replaced the corrupted file with a good version then?

Relevant minor comment pertaining to #165: the most relevant warning in the build logs was this one:

    rawopen.c:95:21: warning: assignment to ‘int (*)(struct OldRawFp *, struct RadarParm *, struct RawData *)’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
         ptr->rawread=-2;

which is telling us exactly what the problem was in the first place :)

ecbland:
Thanks for implementing this @mts299 !
I also get a different total number of warnings, but the specific warnings addressed by this PRs are gone :) 

@asreimer : This is an aside to this PR, but I still get the error with the file you attached.

asreimer:
@ecbland oh that's very interesting! I only get the warning if I do:

    make_fit -old 2001102410w.dat.bz2 w_fit

instead of:

    make_fit -old 2001102410w.dat w_fit

Not related to the PR since you can reproduce the error. Likely related to some difference in versions of GCC and such on my system.

asreimer:
Since both @ecbland and I verified this works, I think we can merge this. All checks also passed, so we're good to go right?

PR #247, opened 2019-07-17
Bug fixes in time_plot

ecbland:
This PR fixes a few small bugs in `time_plot` introduced in #235  and #209 

To test: 

1. Run `time_plot -x -a -old [oldfitfile]` and check that the data displays correctly
2. Check that you are able to specify color keys for the lag zero power and phase parameters
e.g. `time_plot -x -p0 -p0key grey.key -phi -phikey green.key [fitacffile]`


ksterne:
@ecbland, could you fix the codacy check issues?

ecbland:
The codacy problem here is that `fitseek` and `fitread` are called on consecutive lines so that the first value of `status` is never used:
```
status=OldFitSeek(oldfitfp,yr,mo,dy,hr,mt,0,NULL);
status=OldFitRead(oldfitfp,prm,fit);
```
A simple fix would be to check the value of `status` before `OldFitRead` is called, and give an error if `status==-1`. I'm not sure if this error could ever occur, but would this be an acceptable fix? i.e.

```
status=OldFitSeek(oldfitfp,yr,mo,dy,hr,mt,0,NULL);
if (status==-1) {
  fprintf(stderr,"Error determining start/end time of file. Please specify using -st and -et options",\n);
  exit(-1);
}
status=OldFitRead(oldfitfp,prm,fit);
```

ksterne:
@ecbland, oh yes, I see it now in https://github.com/SuperDARN/rst/blob/423ac4f4d43b9b55e7ae408a77e6a503f9a27ac8/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c#L1106-L1107
I think your solution of doing a (error) check of the status variable would be good.  Though we've had feedback that we should only do error checking if it does not significantly impact the speed of the code.  Can you check the speed before/after adding in the if statement to make sure the performance is not impacted too much for normal/good operations?

asreimer:
There are 2 solutions here that will work to satisfy the codacy check:
1) use a different status variable for each function call: status1 for OldFitSeek and status2 for OldFitRead
2) actually check the status for an error

2 is what has been implemented and is what @ecbland is proposing and is IMO the proper solution. If the speed increase from the if statement is, then 1 could be a hack around it, but if it matters, I really really really don't like it.

codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
```diff
@@ -1099,9 +1104,11 @@ int main(int argc,char *argv[]) {
                                prm->time.sc+prm->time.us/1.0e6);
        TimeEpochToYMDHMS(stime,&yr,&mo,&dy,&hr,&mt,&sc);
        status=OldFitSeek(oldfitfp,yr,mo,dy,hr,mt,0,NULL);
+       status=OldFitRead(oldfitfp,prm,fit);
```
asreimer: Why is this needed here?

ecbland: `FitFSeek` and `OldFitSeek` work in different ways. Without the additional `OldFitRead` line, the following code will produce a blank plot (no data and no radar information):
```time_plot -x -a -old [file]```

This problem only manifests itself when plotting all beams, so ```time_plot -x -a -old -b 7 [file]``` will work just fine. I'm not sure exactly why this is, so there might be a more elegant solution. But the OldFitRead line is not redundant.

asreimer: Thanks! This code is really not easy to read, so I thought I would ask.

Sounds like a different issue that can be addressed by a different pull request if needed.

ecbland:
> Though we've had feedback that we should only do error checking if it does not significantly impact the speed of the code. 

I ran the code several times on this branch and develop, and the error check introduced in my last commit does not have any noticeable impact the speed.

aburrell:
Great, then let's use the error check!

ecbland:
The error check has been implemented in commit 59f136a
If this PR has been tested then can someone please merge it?

aburrell:
I tested the two versions of the command with fitacf 2.5 and 3.0 files, creating the plots shown here.  Not sure how to create a file that the `-old` flag will work with.  Elucidation or alternative tester needed.

time_plot_bugfix_jul2019_standard_rti

time_plot_bugfix_jul2019

ecbland:
@aburrell You can download old format `fit` files from the VT server. Choose a year before 2006 and look in the `fit` directory

ksterne:
@ecbland, by VT server, do you mean the BAS or Globus servers?  Otherwise, I'd be curious on how you're downloading from VT (email me off github).

aburrell:
I only have access to the BAS server, and as far as I know they don't have the fit files.

ecbland:
@aburrell Ok, download any `dat` file from BAS and then run `make_fit -old [file].dat [file].fit`

ksterne:
@aburrell, certainly no rush here as I want things done right, but wanted to check to see if you've been able to convert the dat to fit.  Otherwise, for testing this branch I could host/post some fit files to use.  If so, just let me know a particular date/time that you want me to post.

aburrell:
@ksterne if you could host a fit file somewhere that would be great.  I don't have any preference for date or radar.  I will likely reuse that file for all sorts of testing, so a place and time with both ionospheric and groundscatter would be useful.

ksterne:
Hopefully this is something the data standards group can take up in the future...or I know we've talked about, but that's for another time/conversation/issue.  For now I've got some fit files posted at:
[vt.superdarn.org/share/rst_github_files/pr247/](vt.superdarn.org/share/rst_github_files/pr247/)
for testing purposes.  Let me know if these don't seem to work so well.

ksterne:
@aburrell, I know you and others are up against the AGU abstract deadline today, but any chance you'll be able to take a look at the fit files I posted to see how they work with this PR soon?

mts299:
I tested it with the fit file: 2004051404d.fit
command `time_plot -png -a -old 2004051404d.fit > 2004051404d_summary.png`

result 
2004051404d_summary

Does this look good? If so I can merge the branch.

ecbland:
@mts299  Yes, that's working!

mts299:
Merging!

PR #248, opened 2019-07-22
Fixed the typo in rtgrid

mts299:
I found a typo in `rtgrid` documentation that I pointed it out in PR #186 but then merged that PR. So this PR is now focused on fixing that one typo. 

One question I do have is, do we still want to use the address in the example:  gbayqnx.jhuapl.edu?
It didn't work and APL doesn't run GoosBay anymore.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/rtgrid.1.22/doc/rtgrid.doc.xml
+++ b/codebase/superdarn/src.bin/tk/tool/rtgrid.1.22/doc/rtgrid.doc.xml
@@ -95,15 +95,15 @@
 </description>
 
 <example>
-<command>rtgrid -L log -tl 120 -i 120 -if pid.id kapqnx.jhuapl.edu 1024</command>
-<description>Generate a grdmap file using the real-time data stream from the host "<code>kapqnx.jhuapl.edu</code>", served at port 1024. Ignore the scan flag and use a scan length of two minutes. Generate a record every two minutes.  The process identifier is recorded in the file "<code>pid.id</code>", logs of all transactions are recorded in the file "<code>log.<em>yyyymmdd</em></code>" where <em>yyyy</em> is the year, <em>mm</em> is the month, and <em>dd</em> is the day.
+    <command>rtgrid -L log -tl 120 -i 120 -if pid.id <ip address/hostname> <port number></command>
+            <description>Generate a grdmap file using the real-time data stream from the specified <ip address/hostname>, served at the provided <port number>. Ignore the scan flag and use a scan length of two minutes. Generate a record every two minutes.  The process identifier is recorded in the file "<code>pid.id</code>", logs of all transactions are recorded in the file "<code>log.<em>yyyymmdd</em></code>" where <em>yyyy</em> is the year, <em>mm</em> is the month, and <em>dd</em> is the day.
 </description>
 </example>
 
 <example>
-<command>rtgrid -rpf -L log -if pid.id -p /data/rt/grd gbayqnx.jhuapl.edu port.gbr</command>
+    <command>rtgrid -rpf -L log -if pid.id -p /data/rt/grd <ip address/hostname> port.<radar identifier></command>
 <description>
-Generate a grdmap file using the real-time data stream from the host "<code>gbayqnx.jhuapl.edu</code>", served at port contained in the file "<code>port.gbr</code>". The process identifier is recorded in the file "<code>pid.id</code>", and logs of all transactions are recorded in the file "<code>log.<em>yyyymmdd</em></code>" where <em>yyyy</em> is the year, <em>mm</em> is the month, and <em>dd</em> is the day.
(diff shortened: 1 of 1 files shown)


ecbland:
> One question I do have is, do we still want to use the address in the example: gbayqnx.jhuapl.edu? It didn't work and APL doesn't run GoosBay anymore.

Is there an address that works that can be used here?

mts299:
There are addresses that work, but I think the concern is security risks. 
We could have a line like this:
`rtgrid -L log -tl 120 -i 120 -if pid.id <ip address/hostname> <port number>`

`rtgrid -rpf port.<radar identifier> -L log -if pid.id -p /data/rt/grd <ip address/hostname>`

Not sure how explicit we need to be in our examples, especially ones that produce any visuals.

ecbland:
@mts299 Since security is an issue here, it makes sense for the example to be generic as you have suggested. And I think that's better than a specific example that doesn't work.

mts299:
Done! 
You can view the documentation by:
`make.doc`
then on linux: `xdg-open doc/html/index.html`
For other systems, you can also double click the index.html file or place it in a web browser. 
And navigate to the rtgrid documentation: superdarn-> src.bin->tk->tool-> rtgrid

ksterne:
@mts299, can you elaborate on what the security concern might be?  The ip address and port number being public would just be an instance where a data stream is present.  I guess you could make a bunch of connections to that stream in as a DoS attack?

mts299:
Pretty much DoS attacks. Since they know the IP address, they could spam a lot of ports or the given one that is also provided in the command line. I know @egthomas had some concerns about this when some of our code on GitHub had the Dartmouth radar IP addresses and port numbers open to the public.

Plus, if we use one radar's IP address, it may change down the road, and then we would have to remember to come back and change the documentation. This change may happen a lot depending on the network the radar is on and who the providers are. So having these place holder labels may also help in documentation maintainability. 

However, if you want to volunteer a radar IP address and port then I can put it in the documentation :p

ksterne:
Just want to make sure (again, we should be in good habit of this), the test here is:

- [ ] Does the code still compile? (Run `make.code` is successful)
- [ ] And does it update the documentation? (Run `make.doc` to update the documentation on rtgrid)

ksterne:
So, I _thought_ the compiling should have been an easy box to check.  But `make.code` is failing as the 'hlpstr' and 'errstr' are undeclared at:
https://github.com/SuperDARN/rst/blob/a6e22b75524a866101d6801bc2072bd1741ae569/codebase/superdarn/src.bin/tk/tool/rtgrid.1.22/rtgrid.c#L357-L373
 which means they aren't being included correctly.  But this seems to be that the 'errstr.h' and 'hlpstr.h' files are zero size when running a `make.code`.  Also happens in `rtgrid.1.22/` when running `make clean` and `make`.  I don't _think_ my environment is messed up, but I can't be sure.

@mts299, can you try doing a `make clean` and `make` in `rtgrid.1.22/` and see if you get the same thing?

mts299:
@ksterne I will look into that. Didn't think changing some sentences around would cause problems in the code compilation :/

mts299:
The mystery gets more interesting
```bash
make clean
make: *** No rule to make target 'clean'.  Stop.
```

mts299:
Soooo I am pretty sure this is not me changing the documentation but rather the documentation finally changing to make `make.code` try to compile the command-line options. 

I might make another PR for this because this is kind of out scope in the sense this may actually be an issue with other documentation. Just let me investigate some more.

ksterne:
I'm not sure I totally follow you, but I've gotten the same result on a similar but different OS.  It sounds like you're seeing the same, though I'm also now seeing the Travis-CI build is failing (wasn't it passing before?).  

Let us know what you find, but I think it may be good to correct it here as I can confirm the current `develop` branch is compiling.  I don't think it'd be good to merge this with the travis-ci check failing, even if there was a follow on PR.  Maybe one solution is to make a PR into the rtgrid_documentation branch?

ksterne:
Taking a look at this, I think the issue comes in with the use of the <> characters here and xml trying to make an odd tag with these...or just failing, probably a reason the `errstr.h` and `hlpstr.h` files aren't generated.  It seems as though quotes (") or square brackets ([) seem to work in place of what you've used to mark the ip address/hostname and port and such.

mts299:
@ksterne I did try "" quotes after you pointed out the problem but it still was failing on me. So I am not sure what is going on. And yes before Travis CI was passing before... Sorry I got busy with other things and forgot to look into it.

mts299:
RIP:
```
xmldoc -ilf ~/superdarn/rst/build/help/ ~/superdarn/rst/build/help/cfg/ ~/superdarn/rst/codebase/superdarn/src.bin/tk/tool/rtgrid.1.22/doc/rtgrid.doc.xml > hlpstr.h
Segmentation fault (core dumped)
```

This is the command to make the `errstr.h` and `hlpstr.h` files

mts299:
NVM there was a typo in that command and I missed a pair of < > so, yes those were the reason it was failing. 

Fixed the issue now and about to push once I test `make.code`

PR #261, opened 2019-12-31
AACGM_v2 update

egthomas:
This pull request updates the AACGM_v2 coefficients and source code based on @sshepherd's recently released v2.6 of his AACGM_v2 software.  The main improvement is the utilization of the new IGRF13 model coefficients which extends the temporal coverage by 5 years from 2020 to 2025, as well as a few other bug fixes made by @sshepherd.  Note that without these changes, AACGM_v2 coordinate transformations for dates on or after 1st January 2020 (ie tomorrow) will be impossible.

I've fixed a few other typos and environment variables necessary for compilation, and also generated older-style igrf/dgrf coefficient files needed by the IGRFCall function used by the gridding software.  

Any questions or complaints can be directed to @sshepherd.  Happy new year!

PR #265, opened 2020-01-28
4.3.1 release

ksterne:
This patch release is to get the AACGM v2.6 update (#261) out as a general release since it's been determined that there is a more urgent need to have this released and it cannot wait for 4.4 release.  This should be a quick check/test here so I'll merge it in on Feb. 3 if there's no objections on how this looks/functions.  This should give some time for people to review and comment.

We will still proceed with release 4.4 as scheduled about 3 weeks after this release.


egthomas:
@ksterne could you clarify why this needs to be treated as a patch rather than a minor release?  That would simplify the merging procedure by allowing you to include the other already-approved changes on the develop branch rather than cherry picking.  And I don't recall any extra testing time being required for minor vs patch releases:

```
- No more than two weeks of testing should be performed on each new MINOR release branch before merging into the master branch
- No more than one week of testing should be performed on each new PATCH release branch before merging into the master branch
```

ksterne:
This is being treated as a patch as it's one feature that is critical to being released rather than other parts.  Minor release would require longer testing period and is already in the works that was agreed upon during the last Data Analysis WG meeting.

sshepherd:
Good news, Kevin. I think it best that someone else test the installation and whether the test functions work after recompiling. Evan set things up here so it might be redundant for me to test. I don't expect any different behavior from the previous version since the DLMs shouldn't change.

egthomas:
@ksterne ok thanks.  In that case the text of the release guidelines should probably be changed to reflect the longer testing periods now being required (rather than only an upper limit being defined).

ksterne:
Looks like we're good to go here.  I'll send out a note to the darn-users list.  Someone (maybe Simon or Evan) could you also send out an e-mail to the SPA newsletter as a follow up if it's appropriate?

egthomas:
@ksterne just noticed that the `rst.version` file was not updated to reflect this patch release.

ksterne:
Ah crud.  I'd say it's not worth fussing over as we'll have a 4.4 out soon enough.  Others thoughts?

PR #266, opened 2020-02-05
hdw.dat.sps update from hdw.dat repo

ksterne:
Hello all, I can't remember where we landed with this and if this pull request is overkill.  But I thought I'd make it anyways.  Simple enough to check the vtsuperdarn/hdw.dat repo and this change to see they are the same.  And make sure it doesn't break compilation, though the travis-ci should do that for us. Easy enough to knock out!


tables/superdarn/hdw/hdw.dat.sps
```diff
@@ -98,5 +98,6 @@
 #  1   2      3        4         5       6      7      8    9   10   11   12  13(1)  13(2)  13(3)  14   15  16 17
 #
 # UNTIL 01/Jan/3000 05:45:36.0
-   22 2999 31556736  -89.995  118.291  3000.0   75.7  3.24  1   10  0.025 -1    0.0  -97.5    0.0   0.0  2  75 16
-# EOF 
+   22 2020 10000  -89.995  118.291  3000.0   75.7  3.24  1   10  0.025 -1    0.0  -97.5    0.0   0.0  2  75 16
```
ecbland: I know it's not technically the DAWG's responsibility to check hardware files, but column 3 (`10000`) implies that this line is valid until 1 January 2020 at 02:47UT. This is suspicious because it's the same value as the next line.

There is also no comment line for this entry (`# UNTIL 01/Jan/2020 ...`)

ksterne: Very true.  The short is it's up to the PIs/operators on how they would like their hdw.dat files to look like.  For example, the hdw.dat.cly file in the hdw.dat repo has a much shorter header.  And while there is a requirement that the last line be the year 2999 (yes, it really does) there isn't a requirement for the ending time, so I could see its easy enough to keep the same time.  While I would like to keep things looking nicely here, it sounds like something for the Data Standards WG to address?

ecbland: @ksterne My main concern was whether the exact time of the hardware change was correct. I see now that column 3 on both lines were changed to 10000 so yes it's probably just to keep them the same. 
Comments in the hardware files can be addressed by DSWG.

PR #267, opened 2020-02-05
Updating iqdat analysis tools

egthomas:
This pull request updates various routines for reading and plotting `iqdat` files.  Previously, both the native IDL and DLM versions of IQRead were unable to initialize the `samples` array to the correct size when reading a record from an `iqdat` file - this has been fixed by doubling the size of the array if the XCF flag is set in the parameter block.

Probably more significant are updates to the `iqplot` binary.  New command line options have been added which allow the user to control the delay time between frames, select start and/or end times, plot samples from the interferometer array instead of the main array, plot values from only a single beam, etc.  The documentation has also undergone significant revisions to include these new options as well as to describe previously existing options.

Unfortunately I don't have many different `iqdat` files available to me, so some of these changes may not work with `iqdat` files recorded by radars with analog receivers or other designs.  @asreimer may be able to comment on that, as it looks like he did a lot of work on `make_raw` several years ago on his fork of the RSTLite repository.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/iqplot.1.6/doc/iqplot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/iqplot.1.6/doc/iqplot.doc.xml
@@ -16,6 +16,26 @@
 </option>
 <option><on>-h <ar>height</ar></on><od>set the height of the plot to <ar>height</ar>.</od>
 </option>
+<option><on>-r</on><od>plot the real (I) component of the I&Q samples.</od>
+</option>
+<option><on>-i</on><od>plot the imaginary (Q) component of the I&Q samples.</od>
+</option>
+<option><on>-p</on><od>plot the power of the I&Q samples on a logarithmic scale.</od>
+</option>
+<option><on>-m <ar>ymax</ar></on><od>set the extent of the Y-axis to +/-<ar>ymax</ar> when plotting I&Q components.</od>
+</option>
+<option><on>-s <ar>xmin</ar></on><od>set the first I&Q sample plotted to <ar>xmin</ar>.</od>
+</option>
+<option><on>-n <ar>xnum</ar></on><od>set the number of I&Q samples plotted to <ar>xnum</ar>.</od>
+</option>
+<option><on>-xmaj <ar>xmajor</ar></on><od>set the number of X-axis major tick marks to <ar>xmajor</ar>.</od>
+</option>
+<option><on>-xmin <ar>xminor</ar></on><od>set the number of X-axis minor tick marks to <ar>xminor</ar>.</od>
+</option>
+<option><on>-ymaj <ar>ymajor</ar></on><od>set the number of Y-axis major tick marks to <ar>ymajor</ar>.</od>
+</option>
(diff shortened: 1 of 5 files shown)


ecbland:
Thanks @egthomas! The new options in `iqplot` work for the LYR radar data. I noticed that the date & time do not display correctly on the plot (see image below). This also occurs on the develop branch so it's not a specific problem with this PR.

I'm not able to read an IQDAT file in IDL:
```
IDL> unit=iqopen('20180612.1001.00.lyr.iqdat')
% Loaded DLM: IQDLM.
IDL> s=iqread(unit,prm,iq,badtr,samples)
Segmentation fault
```
This happens on this branch and also on `develop`. I'm sure I've able to read these files in the past, but am I doing something wrong? I also tried calling `iqseek()` before `iqread()`, in which case `iqread()` does not segfault but returns error -1. 

If anyone else would like to test this PR with the LYR I&Q data, just send me an email. LYR is a Leicester-style stereo radar. 

20180612 1001 00 lyr iqdat

egthomas:
@ecbland `iqplot` uses `tval` in the `IQ` structure to generate the date and time strings for each pulse sequence at the top of the plotting area.  The LYR `iqdat` files you shared with me do not have any values written for the sequence times, attenuation, noise, offset, or size fields (all zeros).  The lack of sequence times are causing the `01 Jan 1970 00:00...` date/time strings in `iqplot`, and the lack of `offset` values in the `IQ` structure are likely causing the seg fault in the `IQRead` routines.

egthomas:
@ecbland after looking more closely at the LYR `iqdat` files, the following fields appear to be missing:

- `btnum` scalar in the iqdat file / [tbadtr](https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/iq.1.7/include/iq.h#L42) scalar in the C IQ structure
- `tbadtr` array in the iqdat file / [badtr](https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/iq.1.7/include/iq.h#L48) array in the C IQ structure
- `badtr` array in the iqdat file / [badtr](https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.lib/tk/iq.1.7/src/iqwrite.c#L98) array in the C code

Let's all pause here and take a moment to appreciate just how intuitive those variable names are.

So the IDL DLM throws a seg fault [here](https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.dlm/iqdlm.1.6/iqidl.c#L84) when it tries to copy the missing `iq->badtr` array from the IQ structure into a new array.  The native IDL code without using any DLMs is able to handle these missing fields however.

The latest commit should fix the DLM version of IQRead, however my removal of the chnnum parameter from the samples array initialization may break things for older analog-style radars.

ecbland:
Tested again using IQ data from LYR, SAS and BPK which I had lying around. The IDL DLMs are now working properly and the plots are displaying correctly. 

Would anyone else like to test this with IQ data from other radars?

egthomas:
@ecbland if you could share an iqdat file from SAS and BPK that would be really helpful (particularly given BPK's unique interferometer antenna arrangement).

PR #269, opened 2020-02-11
Added instructions on how to contribute to RST

ecbland:
This pull request adds some general information to the `README.md` file to help users find out what the DAWG is working on and how to contribute to RST. 

This is in response to feedback from the SuperDARN community obtained during the AGU Fall Meeting in 2019, and also an attempt to recruit some new developers/testers. 

@ksterne We could list our names & email addresses on line 27 (if you're happy to do that). Or is there a better way to do this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The DAWG welcomes new testers and developers to join the team. Here are some way
   - Test pull requests: to determine which [pull requests](https://github.com/SuperDARN/rst/pulls) need to be tested right away, filter them by their milestones
   - Get involved in projects: the [projects](https://github.com/SuperDARN/rst/projects) page is used to organize our priority areas. Each project board shows items which are in progress, ready for testing/review, and completed.
  - Discuss [issues](https://github.com/SuperDARN/rst/issues) and answer questions
- - Become a developer: if you would like to contribute code to RST, please contact the DAWG chairs
+ - Become a developer: if you would like to contribute code to RST, please contact the DAWG chairs: Kevin Sterne (kevintyler *at* vt *dot* edu) and Emma Bland (emmab *at* unis *dot* no)
 
 ## Historical Version Log
 
--- a/README.md
+++ b/README.md
@@ -2,19 +2,29 @@
 
 Radar Software Toolkit
 ========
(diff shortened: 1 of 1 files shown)


mts299:
That is a nice section you got their @ecbland  ;)

ecbland:
> That is a nice section you got their @ecbland ;)

Yes, shamelessly adapted from the pyDARN readme...thanks @mts299 😉

egthomas:
It may not need to go in the readme or be decided with this pull request, but the procedure for generating the Zenodo/DOI citation (ie author list and order) for each release should also be formalized.  Are all past authors always included in the author list?  Only those who contributed since the previous release?  Some middle ground between those two options?  That's why I haven't updated the Zenodo page for the most recent patch release - unlike previous minor releases the approach was not quite as clear.

mts299:
Good question @egthomas, however, to keep this PR within the scope and to avoid bloating of documentation I moved your comment to issue #272.

ksterne:
@ecbland, what are your thoughts on including e-mail contact info for us?  I think it will increase my amount of spam, but I'm OK with adding one of my e-mail addresses (the kevintyler at vt edu one) to this.

ksterne:
Are the permissions setup so that non-members of a list can post to that list though?

ecbland:
> Are the permissions setup so that non-members of a list can post to that list though?

@ksterne I doubt it, although I can't see anything on the mailing list webpage related to this so I can't verify. 

As an alternative we could replace line 27 with something like "If you would like to contribute code to RST, please open a new issue on the RST github page (https://github.com/SuperDARN/rst/issues) and we will get in touch". This is future-proof, it prevents spamming both the mailing list and the DAWG chairs, and any potential developer would need a github account anyway. Thoughts?

ksterne:
Seems as though non-members can e-mail the list...since I just did.  I don't have a problem either way.  I could see an advantage to keeping personal e-mails so that everyone doesn't get spammed in the future.  I've heard in the past the creating a github account can deter some people away.  Though if they are wanting to contribute specifically to this repo (and not just be apart of the DAWG), then they'll more than likely need to make a github account.

ecbland:
@ksterne Ok, let's keep it with our personal email addresses then. In which case perhaps this is ready to merge if there are no other issues?

PR #275, opened 2020-02-20
More iqdat fixes

egthomas:
This pull request makes a few more bug fixes to the `iqplot` and `make_raw` code.  Using the changes adopted from @asreimer's fork of the RSTLite repository, `make_raw` on this branch is capable of reproducing a rawacf file from an iqdat file for the Christmas Valley East and West radars with the `-d` option.  This can be tested by following a procedure such as

```
make_raw -d yyyymmdd.cve.iqdat > yyyymmdd.cve.rawacf
make_fit yyyymmdd.cve.rawacf > yyyymmdd.cve.fitacf
time_plot -x -b 11 -a -e yyyymmdd.cve.fitacf
```

and compare with the output from the corresponding rawacf file distributed via the data mirrors.

Unfortunately, there is an issue with the other set of iqdat files most readily available to me (Fort Hays East / West) where the sequence offset and size values in the IQ structure appear to be incorrect based on the number of samples per sequence, so `make_raw` does not correctly reproduce the rawacf files for those sites.  I think there may be a similar issue with reproducing Stereo rawacf files from the LYR iqdat files @ecbland shared with me (although the LOS data was somewhat noisy that day so it's hard to say).

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/iqplot.1.6/iqplot.c
+++ b/codebase/superdarn/src.bin/tk/plot/iqplot.1.6/iqplot.c
@@ -160,7 +160,7 @@ void plot_time(struct Plot *plot,
 
 
 void plot_ephem(struct Plot *plot,
-               float xoff,float yoff,float wdt,float hgt,
+               float xoff,float yoff,float wdt,float hgt,int interfer,
                int bmnum,int channel,int tfreq,float noise,int nave,int ave,
                unsigned int color,unsigned char mask,
                char *fontname,float fontsize,
@@ -168,6 +168,7 @@ void plot_ephem(struct Plot *plot,
   int i;
   char txt[256];
   float txbox[3]={0,0,0};
+  char antenna[256];
 
   char *dig="0";
 
@@ -177,9 +178,11 @@ void plot_ephem(struct Plot *plot,
   txtbox(fontname,fontsize,strlen(dig),dig,txbox,txtdata);
   cwdt=txbox[0];
 
-  sprintf(txt,"bmnum=%.2d channel=%.1d tfreq=%.5d noise=%4.2f smp=%.3d/%.3d",
-          bmnum,channel,
-          tfreq,noise,ave,nave);
+  if (interfer) sprintf(antenna,"Interferometer");
+  else          sprintf(antenna,"Main Array");
+
+  sprintf(txt,"bmnum=%.2d channel=%.1d tfreq=%.5d noise=%4.2f smp=%.3d/%.3d (%s)",
+          bmnum,channel,tfreq,noise,ave,nave,antenna);
   txtbox(fontname,fontsize,strlen(txt),txt,txbox,txtdata);
   x=xoff;
   y=yoff+txbox[2];
@@ -616,7 +619,8 @@ int main(int argc,char *argv[]) {
 
       plot_time(plot,2,2,wdt-4,hgt-4,tval,fgcol,0x0f,"Helvetica",12.0,fontdb);
 
-      plot_ephem(plot,100+2,2,wdt-4,hgt-4,prm->bmnum,prm->channel,prm->tfreq,
+      plot_ephem(plot,100+2,2,wdt-4,hgt-4,interfer,
+                 prm->bmnum,prm->channel,prm->tfreq,
                  prm->noise.search,
                  prm->nave,n,fgcol,0x0f,"Helvetica",12.0,fontdb);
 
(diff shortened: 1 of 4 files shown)


egthomas:
Ok so `make_raw` is able to almost reproduce the original rawacf files from the iqdat files for LYR but I think there may still be a range offset not being accounted for due to the stereo organization.  For example, creating a fitacf file from the iqdat file, eg

```
make_raw -d -chnnum 1 20180612.0825.15.lyr.iqdat > 20180612.0825.15.lyr.rawacf
make_fit 20180612.0825.15.lyr.rawacf > 20180612.0825.15.lyr.fitacf
time_plot -x -c a -b 7 -x -a -e 20180612.0825.15.lyr.fitacf
```

Shows the following:

lyr a

whereas the original rawacf file produces the following:

lyr orig a

Likewise for channel B (iqdat version):

lyr b

and original rawacf version:

lyr orig b

If you toggle between the two versions for each channel the subtle differences become more obvious.

ecbland:
@egthomas Are you able to share an iqdat file from CVE/CVW so that I can test this? Which other radars would require the `-d` option (apart from FHE/FHW)?

egthomas:
@ecbland I've uploaded one rawacf and iqdat file each for CVE and CVW here:

http://sdnet.thayer.dartmouth.edu/web_data/iqdat_test/

I'm not entirely sure which radars will require the `-d` option.  It has to do with how the I&Q samples in the buffer / sample array are organized - my understanding is that radars with digital receivers alternate between all of the I&Q samples for the main array, followed by all of the samples for the secondary (interferometer) array.  Older systems multiplexed the samples from each antenna array, so it went `I_main, Q_main, I_int, Q_int, I_main, ...`.  There are some further complications depending on whether or not the XCF flag is set, as we've found that radars using the MSI-style QNX6 ROS3.x software will still record I&Q samples from the interferometer array even if the XCF flag is not set, while the QNX4-era radars will not.

I don't know if any radars using the older hardware are running the most recent QNX4 software which allows writing of iqdat files.  If so, and if any iqdat files were ever produced at those sites, they are likely in the minority.  (That's partly why I asked the DSWG about performing a new radar operations audit).  It might therefore be reasonable to swap the `-d` option around to make it the default and instead require the user to set a `-a` option to specify if a file originated from a site with analog receivers.

egthomas:
@ecbland I've uploaded one rawacf and iqdat file each for CVE and CVW here:

http://sdnet.thayer.dartmouth.edu/web_data/iqdat_test/

Short answer is I think most radars will require the `-d` option, but I'm not exactly sure which ones.  It has to do with how the I&Q samples in the buffer / sample array are organized - my understanding is that radars with digital receivers alternate between all of the I&Q samples for the main array, followed by all of the samples for the secondary (interferometer) array.  Older systems multiplexed the samples from each antenna array, so it went `I_main, Q_main, I_int, Q_int, I_main, ...`.  There are some further complications depending on whether or not the XCF flag is set, as we've found that radars using the MSI-style QNX6 ROS3.x software will still record I&Q samples from the interferometer array even if the XCF flag is not set, while the QNX4-era radars will not.

I don't know if any radars using the older hardware are running the most recent QNX4 software which allows writing of iqdat files.  If so, and if any iqdat files were ever produced at those sites, they are likely in the minority.  (That's partly why I asked the DSWG about performing a new radar operations audit).  It might therefore be reasonable to swap the `-d` option around to make it the default and instead require the user to set a `-a` option to specify if a file originated from a site with analog receivers.

ecbland:
@egthomas I've tested this with the CVE/CVW files you provided. The output of `make_raw -d` is the same as the supplied rawacf files :) 

Regarding making `-d` the default, perhaps we should wait until we know which radars might be affected? There could be some confusion if someone out there is indeed running `make_raw` for an older radar. I noticed that if I omit the `-d` option for CVE/CVW then the code still runs but produces different elevation angle values.

The LYR data seems to be a separate and complicated problem. It would help to have some more IQdata from that radar, but that will have to wait until its antennas grow back ;)

egthomas:
@ecbland that's great!  I've also tested this with a few files from GBR, KAP, and BKS (all running the older QNX4 software with a mix of available interferometer data), although I'll leave that up to @ksterne whether any of those files should be made more widely available for testing purposes.

> Regarding making -d the default, perhaps we should wait until we know which radars might be affected? There could be some confusion if someone out there is indeed running make_raw for an older radar.

Agreed, just thought I'd suggest it for future consideration.

> I noticed that if I omit the -d option for CVE/CVW then the code still runs but produces different elevation angle values.

Yes, I think the `-d` option should only affect the handling of XCF-derived parameters.

> The LYR data seems to be a separate and complicated problem. It would help to have some more IQdata from that radar, but that will have to wait until its antennas grow back ;)

Agreed!  Do you know how often iqdat files were recorded by LYR?  They could eventually be useful for the sea ice project I am working on (although that's unrelated to this pull request).

ecbland:
> @egthomas Do you know how often iqdat files were recorded by LYR? 

Just that day, unfortunately (20180612). We were testing whether we could save the IQ data, but we never started recording it routinely.

PR #276, opened 2020-02-20
JME info for radar.dat file

ksterne:
Hello all,

I got info from @mysunshinezjj on info for the radar.dat file over on the Data Distribution WG e-mail list.  The most basic of tests here for those that can is to make sure I copied things over correctly.  If you don't know, then ask someone that's on the ddwg to forward you the e-mail.  Otherwise, you could check basic compilation test, but travis-ci should do that for us.  Sorry this was slow to get in.

### DIFF
--- a/tables/superdarn/hdw/hdw.dat.jme
+++ b/tables/superdarn/hdw/hdw.dat.jme
@@ -1,0 +1,102 @@
+#  Hardware parameters for Jiamusi East radar.
+#
+# Each radar has a distinct set of hardware parameters that are used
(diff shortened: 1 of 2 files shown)


egthomas:
@ksterne should the `hdw.dat` file for JME be added as well?

ksterne:
Great catch @egthomas!  I'd lost track of whether or not the file was in this repo.  Another to get these linked up....one day. But not today.  I've added the file just a moment ago with 367d20a.

Issue #268, opened 2020-02-11
What's the radius of the Earth?!


ecbland:
During the DAWG telecon on 21 January we discussed how mathematical and physical constants are handled within RST (#238). We agreed that mathematical constants (e.g. pi) would be defined using `math.h`, and physical constants (e.g. speed of light, earth radius) would be defined in a separate file in RST.

This issue is specifically about the value of the **Earth radius**. So far I've found two definitions of the Earth radius in RST:

1. Earth's mean spherical radius, RE=6371.2km. This value is used in the IGRF, so RST uses it for anything related to IGRF/AACGM. 

2. Earth's polar radius, RE=6356.779km, corresponding to the Earth radius at 88°N geodetic latitude. This value is used by the field of view mapping routines (`RPos*`/`RadarPos*`), which convert range/beam coordinates to geographic coordinates. 

I suspect that (2) is used in the `RPos` routines because the SuperDARN radar fields of view cover higher latitudes where the earth radius is slightly smaller. But this will always be an approximation and it might not really matter which value is used.

I've created this issue to highlight the fact that multiple definitions exist and to give people an opportunity to offer their expertise. @mts299 is planning to create a file of physical constants in RST so that they don't need to be defined separately in each routine (see #238), and has asked for advice on what values should be used. I suggest that we create two separate definitions of RE for now, and then take the opportunity during testing to see if there is any practical difference between the two values.

mts299:
@DannoPeters

aburrell:
> I suggest that we create two separate definitions of RE for now, and then take the opportunity during testing to see if there is any practical difference between the two values.

I agree with the above statement made by @ecbland

egthomas:
[rmath.h](https://github.com/SuperDARN/rst/blob/master/codebase/base/src.lib/math/rmath.1.8/include/rmath.h) has historically been the source of mathematical constants for the RST.  That seems like a logical place to consolidate the constants already defined there like C or pi for a minimal impact on the existing software.

I would advocate for some caution when changing the constants defined in externally developed libraries (AACGM, IGRF, AstAlg, etc).  For example, by changing the AstAlg constants you might be making them more consistent with those used by other RST libraries while also making the results inconsistent with those obtained by anyone else using that published library.

[...]

PR #281, opened 2020-03-05
Feature/constants

mts299:
So I am a bit swamped with other work regarding map file generation for Kathryn. However, I did create the constants branch and got it to compile. I just haven't tested it. I expect FITACF 3.0 values will be different as that was the one algorithm that defined it's own constants the most. 

Avoided outside algorithms, and not changing too much of the current code. Mainly just added "rmath.h" to all the code I could find using its defined constants. 

Other constants up for debate was `LN_TO_LOG`, again FITACF 3.0 defined it's own but now used "rmath.h" value. 

And "rmath.h" has its complex data structure. This may get out of scope but maybe considering if we should use <complex.h> instead. Anyways I got a workaround by rewriting the 
```
ifndef _COMPLEX_H
```
a line such that if a user wants to use complex.h, rmath.h won't get in the way. 

I can make fixes/changes as people test it. 

Side note: I didn't touch IDL code... sorry not as familiar with that.


mts299:
curious did I do something wrong the commits stream looks different than usual?

mts299:
I see that `develop` is missing our LICENSE file.

ecbland:
@mts299 this might be related to the most recent patch release - we had to cherry-pick the commits relating to IGRF/AACGM. I'm not sure how to explain the missing licence file though. @ksterne?

egthomas:
@mts299 @ecbland It looks like the license file was added directly to `master` by #255 

Regarding this pull request, something strange appears to have happened to `cnvtcoord.c` - maybe a find/replace algorithm gone wrong?

ksterne:
Yes, I suspect the 10 commits behind (even though it's 85 commits ahead as well) is from the latest patch.  I didn't pull `master` back into develop since I did cherry-pick the critical commits for the patch.  I did this at the time as it didn't seem like there would be much more added to the next minor release...though that's certainly changed now.

mts299:
Thanks @egthomas on the catch with "find and replace" I went through the file and fixed it up correctly now. I also need to go in and mention my modifications to files.

codebase/base/src.lib/math/rmath.1.8/include/rmath.h
```diff
@@ -34,23 +34,28 @@
 #ifndef _RMATH_H
 #define _RMATH_H
 
-#ifndef _COMPLEX
+/*Need to use _COMPLEX_H to not
+ * override the comple.h definition*/
```
ecbland:
```suggestion
 * override the complex.h definition*/
```

codebase/base/src.lib/math/rmath.1.8/include/rmath.h
```diff
@@ -34,23 +34,28 @@
...
 #ifndef HUGE_VAL
 #define HUGE_VAL 1.0e36                /* default value of HUGE */
 #endif
  
-#ifndef PI
+/*PI constant*/
 #define PI 3.14159265358979323846
-#endif
+
+/*Earth's radius in km*/
```
ecbland:
```suggestion
/*Earth's mean spherical radius in km*/
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -5,36 +5,37 @@
...
  but WITHOUT ANY WARRANTY; without even the implied warranty of
- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- GNU Lesser General Public License for more details.
+ MCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Less General Public License for more details.
 
- You should have received a copy of the GNU Lesser General Public License
+ You should have received a copy of the GNU Less General Public License
  along with RST.  If not, see <http://www.gnu.org/licenses/>.
 */
```
ecbland: Looks like some typos have been introduced here - what's going on?

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -61,14 +62,14 @@ double slant_range(int frang, int rsep,
 /**
  * Converts from geodetic coordinates (gdlat,gdlon) to geocentric spherical
  * coordinates (glat,glon). The radius of the Earth (gdrho) and the deviation
- * off the vertical (del) are calculated. The WGS 84 oblate spheroid model
+ * off the vtical (del) are calculated. The WGS 84 oblate spheroid model
```
ecbland:
```suggestion
 * off the vertical (del) are calculated. The WGS 84 oblate spheroid model
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -61,14 +62,14 @@ double slant_range(int frang, int rsep,
...
  * of the Earth is adopted.
  **/
 void geodtgc(int iopt, double *gdlat, double *gdlon,
              double *grho, double *glat,
              double *glon, double *del) {
 
-    /* WGS 84 oblate spheroid defining parameters */
+    /* WGS 84 oblate sphoid defining parameters */
```
ecbland:
```suggestion
    /* WGS 84 oblate spheroid defining parameters */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -99,7 +100,7 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
     /* Calculate the geocentric Earth radius at the geodetic latitue [km] */
     *grho=a/sqrt(1.0+e2*sind(*glat)*sind(*glat));
 
-    /* Calculate the deviation of the vertical [deg] */
+    /* Calculate the deviation of the vtical [deg] */
```
ecbland:
```suggestion
    /* Calculate the deviation of the vertical [deg] */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
ecbland: Ok, I see there are many instances of this. Perhaps something related to the find & replace mishap?

ecbland: @mts299 I think there is still a problem here. Lots of "er" letter combinations have been erroneously removed from the comments. Can you please check the entire `cnvtcoord.c` file?

mts299: Yea sorry meant to respond to this yesterday. However, my furnace blew a fuse so the repair guy came just as I was about to commit the code. Check it out now.

mts299:
Hey, sorry if I am slow to react to comments/suggestions. Currently battling some chronic laryngitis and several other things going on. Thank you for your patience!

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -13,7 +17,7 @@
 
  RST is free software: you can redistribute it and/or modify
  it under the terms of the GNU Lesser General Public License as published by
- the Free Software Foundation, either version 3 of the License, or
+ the Free Software Foundation, eitherer version 3 of the License, or
```
ecbland:
```suggestion
 the Free Software Foundation, either version 3 of the License, or
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -59,7 +64,7 @@ double slant_range(int frang, int rsep,
 
 
 /**
- * Converts from geodetic coordinates (gdlat,gdlon) to geocentric spherical
+ * Converts from geodetic coordinates (gdlat,gdlon) to geocenterric spherical
```
ecbland:
```suggestion
 * Converts from geodetic coordinates (gdlat,gdlon) to geocentric spherical
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -76,27 +81,27 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
     /* Ellipsoid semi-minor axis */
     b=a*(1.0-f);
 
-    /* Second eccentricity squared */
+    /* Second eccenterricity squared */
```
ecbland:
```suggestion
    /* Second eccentricity squared */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -76,27 +81,27 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
...
     e2=(a*a)/(b*b)-1;
 
     if (iopt>0) {
 
-        /* Convert geodetic latitude/longitude to geocentric [deg] */
+        /* Convert geodetic latitude/longitude to geocenterric [deg] */
```
ecbland:
```suggestion
        /* Convert geodetic latitude/longitude to geocentric [deg] */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -76,27 +81,27 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
...
         *glat=atand( (b*b)/(a*a)*tand(*gdlat));
         *glon=*gdlon;
 
-        /* Convert geocentric longitude to +/- 180 degrees */
+        /* Convert geocenterric longitude to +/- 180 degrees */
```
ecbland:
```suggestion
        /* Convert geocentric longitude to +/- 180 degrees */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -76,27 +81,27 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
...
         if (*glon > 180) *glon=*glon-360;
 
     } else {
 
-        /* Convert geocentric latitude/longitude to geodetic [deg] */
+        /* Convert geocenterric latitude/longitude to geodetic [deg] */
```
ecbland:
```suggestion
        /* Convert geocentric latitude/longitude to geodetic [deg] */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -76,27 +81,27 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
...
         *gdlat=atand( (a*a)/(b*b)*tand(*glat));
         *gdlon=*glon;
 
     }
 
-    /* Calculate the geocentric Earth radius at the geodetic latitue [km] */
+    /* Calculate the geocenterric Earth radius at the geodetic latitue [km] */
```
ecbland:
```suggestion
    /* Calculate the geocentric Earth radius at the geodetic latitude [km] */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -107,8 +112,8 @@ void geodtgc(int iopt, double *gdlat, double *gdlon,
 
 
 /**
- * Calculates the geocentric coordinates (frho,flat,flon) of a field point given
- * the angular geocentric coordinates (rrho,rlat,rlon) of the point of origin,
+ * Calculates the geocenterric coordinates (frho,flat,flon) of a field point given
+ * the angular geocenterric coordinates (rrho,rlat,rlon) of the point of origin,
```
ecbland:
```suggestion
 * Calculates the geocentric coordinates (frho,flat,flon) of a field point given
 * the angular geocentric coordinates (rrho,rlat,rlon) of the point of origin,
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -179,7 +184,7 @@ void geocnvrt(double gdlat, double gdlon,
 
 
 /**
- * Calculates the geocentric coordinates (frho,flat,flon) of a radar field point,
+ * Calculates the geocenterric coordinates (frho,flat,flon) of a radar field point,
```
ecbland:
```suggestion
 * Calculates the geocentric coordinates (frho,flat,flon) of a radar field point,
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -324,7 +329,7 @@ void fldpnth_gs(double gdlat,double gdlon,double psi,double bore,
  * if this value is less than 90 then it is assumed to be the
  * elevation angle from the radar. If center is not equal to zero, then
  * the calculation is assumed to be for the center of the cell, not the
- * edge. The calculated values are returned in geocentric spherical
+ * edge. The calculated values are returned in geocenterric spherical
```
ecbland:
```suggestion
 * edge. The calculated values are returned in geocentric spherical
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -361,9 +365,9 @@ void RPosGeo(int center, int bcrd, int rcrd,
 
     /* If the input height is less than 90 then it is actually an input
      * elevation angle [deg], so we calculat the field point height */
-    if (height < 90) height=-re+sqrt((re*re)+2*d*re*sind(height)+(d*d));
+    if (height < 90) height=-RE+sqrt((RE*RE)+2*d*RE*sind(height)+(d*d));
 
-    /* Calculate the geocentric coordinates of the field point */
+    /* Calculate the geocenterric coordinates of the field point */
```
ecbland:
```suggestion
    /* Calculate the geocentric coordinates of the field point */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
```diff
@@ -546,15 +545,15 @@ void RPosCubicGS(int center,int bcrd,int rcrd,
     psi=pos->bmsep*(bcrd-offset)+bm_edge;
 
     d=slant_range(frang,rsep,rx,range_edge,rcrd+1)/2;
-    if (height < 90) height=-re+sqrt((re*re)+2*d*re*sind(height)+(d*d));
+    if (height < 90) height=-RE+sqrt((RE*RE)+2*d*RE*sind(height)+(d*d));
     fldpnth_gs(pos->geolat,pos->geolon,psi,pos->boresite,
                height,d,&rho,&lat,&lng);
 
-    /* convert to x,y,z (normalized to the unit sphere) */
+    /* conervt to x,y,z (normalized to the unit sphere) */
```
ecbland:
```suggestion
    /* convert to x,y,z (normalized to the unit sphere) */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c
ecbland: @mts299 There were still a few problems remaining in this file so I've added change suggestions to fix them. Once you commit them I think we're good to go.

mts299: Sweet! sorry, learned my lesson on find and replace :p

mts299:
@ecbland committed all your suggestions!

codebase/base/src.lib/math/rmath.1.8/include/rmath.h
```diff
@@ -34,23 +38,28 @@
...
-#define C 2.997e+8
+/*Speed of Light in m/s*/
+#define C 2.99792458e+8
 #define LN_TO_LOG 4.342944819
 
 #ifndef HUGE_VAL
 #define HUGE_VAL 1.0e36                /* default value of HUGE */
 #endif
  
-#ifndef PI
```
asreimer: Maybe I'm being daft, but if one were to include both `math.h` and `rmath.h` (as is done in codebase/base/src.lib/graphic/fbuffer.1.19/src/matrix.c) then removing the `ifndef` here overwrites the existing `PI` definition? Maybe that's ok, just pointing it out.

codebase/base/src.lib/math/rmath.1.8/include/rmath.h
```diff
@@ -34,23 +38,28 @@
...
+/*PI constant*/
 #define PI 3.14159265358979323846
-#endif
+
+/*Earth's mean spherical radius in km*/
```
asreimer: Let's not rehash #268, but in it there was discussion that we should clarify which particular definition of Earth radius we use. The 6371.2km value comes from: https://en.wikipedia.org/wiki/Earth_radius#Average_distance_from_center_to_surface. If you search for mean radius of the Earth, you'll find there are a ton of definitions and this one is the first most common one (most common is R_1=(2*a+b)/3 = 6371.0km for the WGS84 reference ellipsoid).
```suggestion
/*Earth's mean spherical radius (average distance from center to the actual surface everywhere) in km*/
```

codebase/base/src.lib/math/rmath.1.8/include/rmath.h
mts299: So math.h actually reference Pi as `M_PI` so this library wouldn't affect rmath.h `PI` constant. I hope this would prevent people from using math.h `PI` for RST's consistency sake.
 
https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html
https://www.tutorialspoint.com/c_standard_library/math_h.htm

Granted I left `ifndef` on `HUGE_VAL` which I think is also defined in math.h. 

I had to make `ifndef` around `complex` just because that is a complicated other issues too with RST defining predefined data structures.

I was thinking we want `rmath.h` to overwrite to ensure consistency in other predefined libraries this may help avoid weird behaviors too. However, the external libraries we include in RST is hard to say, but I think that is the authors of those libraries problem ... or we need to come up with a better strategy but that is out of scope for this PR.

asreimer: Ha, totally called it, I was being daft. I got `PI` and `M_PI` confused...

egthomas:
One stray thought here (sorry, typing from my phone) - is there a hardcoded value for Re in the IDL cnvtcoord code? If so, should that be made consistent with the value in rmath.h if it isn't already? I know @mts299 probably doesn't want to mess around with the IDL stuff but some people may still use the native IDL routines, so maybe @ecbland (or someone else) could check?

mts299:
I found some more constants differences and I noticed that shf has it's own constants header file:
codebase/superdarn/src.lib/tk/shf.1.10/include/shfconst.h
Is this an external library too? 

Map functions tend to use shfconst.h file for constants. 
@ecbland What would you like to do?

mts299:
I also made an issue for IDL constants #282 as that is out of scope on this PR and could be its own PR. Especially since I need this PR for getting Simon's elevation angle into Fitacf 3.0.

ecbland:
> Map functions tend to use shfconst.h file for constants.
> @ecbland What would you like to do?

@mts299 Hmmm this is getting way outside my knowledge of RST. Since the primary purpose of this PR is to allow us to add the elevation angle code to FITACF3, I wonder if it's ok to merge this as-is and deal with the map functions separately?

mts299:
Whoops, my bad you are right we can look at this later. 
I will make an issue and hopefully, we can get it figured out before the release.

ecbland:
It's not entirely obvious how to test this, but I have done the following comparisons between this branch and `develop`:

1. `make_fit -fitacf-version 2.5` produces slightly different velocities, spectral widths, elevation angles (and their respective errors)
2. `make_fit -fitacf-version 3.0` produces identical output files
3. `make_grd` produces identical output files

mts299:
I find that odd make_fit 2.5 is different but not make_fit 3.0 since make_fit 3.0 had to use different constant values :/

ecbland:
@mts299 I agree this is strange (and should have made that clear in my earlier comment, sorry). Do you get the same behaviour as me?

codebase/superdarn/src.lib/tk/gtable.2.0/src/gtable.c
```diff
@@ -286,7 +286,7 @@ int GridTableAddBeam(struct GridTable *ptr,
     if (bm==NULL) return -1;
 
     /* Inertial velocity correction as a function of radar geodetic latitude [m/s] */
-    velco=(2*PI/86400.0)*6356.779*1000*cos(PI*pos->geolat/180.0);
+    velco=(2*PI/86400.0)*RE*1000*cos(PI*pos->geolat/180.0);
```
egthomas: I'm not sure this change is appropriate - here the velocity correction factor is a function of the equatorial Earth radius and latitude, and therefore should not use a mean Earth radius value.  Although it isn't clear why a polar radius was originally used here.

aburrell: @egthomas, theoretically, shouldn't the radius at the geographic latitude be used?  If so, I believe the mean earth radius is the most appropriate (given our degree of exactness required).

egthomas: @aburrell the `cos(PI*pos->geolat/180.0)` term on the right accounts for the grid cell's actual geographic latitude.  If you scale a mean Earth radius by the cosine of the latitude instead of an equatorial value the calculated velocity will be incorrect.  Here's an example I found:  https://www.nasa.gov/pdf/142825main_Bobsled_Launch.pdf

aburrell: Thanks for the example @egthomas!  In this case, I'd recommend adding the equatorial radius to the constants (`RE_EQ`?).

asreimer: @aburrell No, I think you are correct. Unless I'm misunderstanding something, the point is to calculate the linear velocity at the surface of the Earth at the latitude `geolat` right @egthomas? If this is the case, then to be precise, one needs to use the "radius from the axis of rotation" at `geolat`, which requires knowledge of the radius of the Earth at that latitude (one must correct for the ellipsoidal shape of the Earth). The PDF linked to assumes (implicitly) a spherical Earth:

> This equation quantifies the effect of latitude on initial velocities for any location on Earth, and it works well, **despite the slight oblateness** (being compressed or flattened at the poles) of Earth.

So that means it will "work well" using a mean radius as well. I don't see an elliptical correction being performed anywhere in the source code for `gtable.c`, so picking the polar, equitorial, or mean radius does not, in a meaningful way, make anything more or less precise.

If the point is to be consistent, using the mean radius that is defined is fine. If the point is to be more precise, then we really need to worry about the ellipsoidal correction that is needed.

So, I suggest we shelvethis and defer it to another PR if the we decide we need to be more precise.

asreimer: Just for perspective on how different the velocities would be using the equatorial vs. mean radius, we're talking `(6378.2-6371.2)/6378.2 = 0.1%`. We already had this discussion earlier in #268 regarding the `RPos*` and `RadarPos*` functions and agreed that we would use a common radius.

egthomas: my apologies @mts299 @aburrell @asreimer , I haven't been at my sharpest these last few weeks with the lack of sleep. This inertial velocity term was something I vaguely remember intending to update a few years ago to use the geodetic latitude at the grid cell location rather than a fixed Earth radius value but never got around to doing so. @asreimer is correct that the practical impact on the calculated velocity is very small, so the proposed changes in this pull request are fine as-is.

asreimer: No worries. Hopefully you can get some needed sleep!

mts299:
@ecbland sorry for the delay actually been working on the elevation_library branch trying to get that too work. I will at it this morning.

mts299:
@ksterne I didn't that morning I had to deal with some data transfer issues. 
I looked at this morning but then I got distracted by running it through memory profiler in hopes to figure out my memory issues in the elevation_library branch.

mts299:
Okay, looked into this. A couple of things:
1. fitacf 2.5 difference never vary over 1.0 and relatively average 10^-03 - 10^-4
which may be expected as PI was changed for more precision and the same with C. 

2. fitacf 3.0 is not different because
- RE change happened only in the refractive_index function which is not called instead 1.0 is used for the refractive index 
- Speed of light actually didn't vary originally
- It is still using the old version of PI  

My expectation we should see at least the same amount of changes in both algorithms not including elevation angles. 

To test I made a pyDARN script let me know if you want to see/use it.

mts299:
Actually, I stand corrected:
old rmath.h C value:
`2.997e+8`

old fitacf 3.0 C value:
`299792458`

New rmath.h C value:
`#define C 2.99792458e+8`

old rmath.h PI value:
`3.14159265358979323846`

old fitacf 3.0 PI value:
`3.14159265358979323846`
New rmath.h PI value:

`#define PI 3.14159265358979323846`

So you can actually see we adopted fitacf 3.0 constant values and fitacf 2.5 got updated in a sense because C became more precise. Since fitacf 3.0 doesn't actually call `refractive_index` function we see no change with radius of the earth change.

ecbland:
@mts299 Ok, that makes sense regarding the behaviour of FITACF2.5 and 3.0. 

Is this ready to merge now, or is there some more testing we should do?

mts299:
I think it should be good to merge. 
Already working on the next phase of this release.

PR #287, opened 2020-04-24
4.3.2 release

ksterne:
This patch release includes updates to hdw.dat files and the radar.dat file for new (DCN, JME) and updated radar information (DCE, SPS).


egthomas:
I guess version numbers in the `rst.version` file aren't getting updated anymore?

ksterne:
Great catch *after* I merged.  Phew, so many things to keep track of.  Let me see if I can right that ship.  I'll commit direct to master.

ecbland:
Sounds like we need to make a checklist for all the tasks required to make a release (maybe on the wiki).

egthomas:
Or bring back the "no self merging" rule?

The main problem is that once a release is created on Github, Zenodo makes a snapshot and assigns a DOI, at which point the master branch is permanently archived (ie you can't edit the files anymore).  So if you want to include the correct `rst.version` file in a DOI'd Zenodo archive, you now have to create a new release which is assigned a new DOI.

ecbland:
I agree that we should avoid self-merging but that doesn't guarantee that mistakes will be caught. I think a checklist would be useful.

ksterne:
Agree with @ecbland where self-merging could have led to this error.  Luckily it's not a big error and something that I think a checklist would help.  For me, it's just a habit/routine that I haven't gotten into.

Ah, that tricky Zenodo thing.  Well, to make it more confusing, I've redone the release...I'll try to avoid this in the future as I know it's not desirable.  Hopefully this is all mute when we get out a 4.4 release (soon).

PR #291, opened 2020-05-07
Fixing bug in RawEncode for missing acfd array

egthomas:
This pull request fixes a bug in `RawEncode` of `rawwrite.c` where a segmentation fault occurs when trying to write a rawacf file (eg with `trim_raw`) if the `acfd` array is missing (and `slist` is empty).  This can be tested by using `trim_raw` on the following two files (many other examples are available):

`20080103.2201.00.inv.rawacf` (23:10:03 UT)
`20080110.2033.57.inv.rawacf` (20:51:43 UT)

On the develop branch there will be a segmentation fault, while on this branch a warning message will be printed to stderr and the `pwr0` array will be written before proceeding to the next record.


ecbland:
Thanks @egthomas! This works as described.

Issue #289, opened 2020-04-29
C90 vs. C99? 


mts299:
As I am tirelessly working on the elevation code, I am noticing we are still using C90 for compiling our code. Is there any reason? 

C90 is causing more compiler warnings because people are using C99 syntax/types that are not supported in C90. Also, C99 as fixed several bugs with C90. 

So can we move to C99 or C11? 

Here are some links supporting to moving to C99: 
https://stackoverflow.com/questions/41535927/compatibility-of-c89-c90-c99-and-c11
http://www.simplyembedded.org/archives/battle-of-the-standards-round1/

ksterne:
This is a great question.  I can't say I know enough about the quirks, differences, advantages, and disadvantages of each.  So my response is not very much from a technical side of things.  My instinct is that we should upgrade to a newer syntax/standard.  

I don't know why we are still using C90 for compiling aside from some of the *initial* building blocks of this code might go back to the mid to late 90s and just nothing was updated.  It might be worth investigating if there's a dependency (that we treat more like it's part of the RST) that needs C90 compiling.

This sounds like a major overhaul to a lot of our files as they might need updating for C99 or C11.  I'd say that we should try for C11, but maybe that is too big of a jump.  Maybe this isn't too bad as I'm now reading #293.

Great job on starting up a project related to this and other things we need to update.

egthomas:
At least a dozen radars in the network are still using QNX4 (circa 1997) for their radar control software.  If we ever want to be able to port these updates to the dmap, rawacf, fitacf, or other libraries to those radars then we need to make sure we retain backwards compatibility with those sites.  That is after all one of the strengths of SuperDARN - common analysis software.

mts299:
Keeping the discussion here so that people can test and report potential issues with the code on the PR. 
Compiler changes do not change how the code analysis the data that is not how compilers work they focus on the updates to the programming language. As some changes to the language can break backward compatibility of compiling that is not the case with C99 - C11 as they were updates to the C90 code having bugs and tight restrictions that people did not enjoy programming in. So no backward compatibility will be harmed. If you have concerns @egthomas then please test it and get back to me. 

The only issue I can actually see is if radars actually pull RST down onto their computers? My thought is that this is not the case since they use ROS and do no produce fitacf files at the site. Plus many radars moving are moving to newer systems since we cannot keep supporting QNX4/other hardware. why not move the software along too :p

egthomas:
[...]

PR #293, opened 2020-05-08
removed ansi C90 and added pedantic

mts299:
## Scope
To further aid in elevation code I raised the issue #289 to switch to C99. 
Thus this PR is removing the `ansi` option for compile and adding `pedantic`

## Reason
More flexibility including some the RST code was actually coded for C99 and `pedantic` is good practice for enforcing more restrictions. 

## Test 
Compile on Linux or MacOS. 
I did see how many warnings I got via: 
```bash
make.build; make.code 2>&1 | grep -i "warning" | wc -l 
```
Got 128 on `develop` and 200 on this branch.

```bash
gcc --version
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/build/make/makebin.darwin
+++ b/build/make/makebin.darwin
@@ -7,7 +7,7 @@
 # Darwin (GCC)
 #
 
-CFLAGS = -Wall -O3 -ansi -D_GNU_SOURCE -D_DARWIN -I/opt/local/include
+CFLAGS = -Wall -pedantic -O3 -D_GNU_SOURCE -D_DARWIN -I/opt/local/include
 
 ifdef XPATH
   ifdef XLIB
--- a/build/make/makebin.linux
+++ b/build/make/makebin.linux
@@ -7,7 +7,7 @@
 # Linux (GCC)
 #
 
-CFLAGS+=-fPIC -Wall -O3 -ansi -D_GNU_SOURCE -D_LINUX
+CFLAGS+=-fPIC -Wall -pedantic -O3 -D_GNU_SOURCE -D_LINUX
 
 ifdef XPATH
   ifdef XLIB
--- a/build/make/makedlm.darwin
+++ b/build/make/makedlm.darwin
@@ -7,7 +7,7 @@
 # Linux (GCC)
 #
 
-CFLAGS = -fPIC -Wall -O3 -ansi \
+CFLAGS = -fPIC -Wall -pedantic -O3 \
          -D_GNU_SOURCE -D_LINUX $(INCLUDE)
  
 SUF=so
--- a/build/make/makedlm.linux
+++ b/build/make/makedlm.linux
@@ -7,7 +7,7 @@
 # Linux (GCC)
 #
 
-CFLAGS = -fPIC -Wall -O3 -ansi \
+CFLAGS = -fPIC -Wall -pedantic -O3 \
          -D_GNU_SOURCE -D_LINUX $(INCLUDE)
 
 LFLAGS = -shared -lm -lz 
--- a/build/make/makelib.darwin
+++ b/build/make/makelib.darwin
@@ -7,7 +7,7 @@
 # Linux (GCC)
 #
 
-CFLAGS = -Wall -O3 -ansi -D_GNU_SOURCE -D_DARWIN -I/opt/local/include $(INCLUDE)  
+CFLAGS = -Wall -pedantic -O3 -D_GNU_SOURCE -D_DARWIN -I/opt/local/include $(INCLUDE)  
 SUF=so
 
 VSTR=$(shell get.version ${CURDIR}/..)
--- a/build/make/makelib.linux
+++ b/build/make/makelib.linux
@@ -7,7 +7,7 @@
 # Linux (GCC)
 #
 
-CFLAGS += -fPIC -Wall -O3 -ansi \
+CFLAGS += -fPIC -Wall -pedantic -O3 \
          -D_GNU_SOURCE -D_LINUX $(INCLUDE)
 
 LFLAGS += -shared -lm -lz 


ksterne:
Sorry I was slow to comment on #289.  I've been slammed with tasks to do.  Thanks for putting this in a good format/template including test code!  One question before I'm testing it out, I'm guessing the reason the number of warnings increases here is that there aren't any updates to the C90 syntax code which is causing additional warnings?

Another question, does this move from `ansi` to `pedantic` break compilation on any reasonably old OS's?  Ubuntu 20 has just been released, but maybe some systems are still running Ubuntu 12 and wonder if it would break compilation there.  I know of at least one system that is ( ;) ) and I know it's unreasonably old (one day I'm working on it).  But maybe a little more reasonably old is Ubuntu 14 or 16?  I'm also biased to the Ubuntu line, but can be expanded to other linux lines.  I'm not advocating a thorough test on each of these OSs, but just if you know off the top of your head what minimum OS requirement this might bring us up to.  If not, then it's not a huge deal.

mts299:
@ksterne No worries, I am finally finding some time for RST and other tasks as well. Thanks! trying to get PR's test faster so hopefully, the template and test code helps :) 

Some warnings like using `double double` would appear with `ansi` C90 but without it, it removes this warning as it is allowed in C99 and up. However, I put `pedantic` in as a stricter warning checker. However, maybe this is a bit too pedantic ;) we could go to Wextra which also good. Here is a linked that explains the flags better: https://embeddedartistry.com/blog/2017/06/07/warnings-weverything-and-the-kitchen-sink/

It shouldn't break since C99 and higher are subsets of C90 and should still compile. If you use ancient GCC compilers and OS's like Solaris then it might break. GCC 4.8.1 and higher support C11 so I am assuming C99 is still safe with older compilers. I guess if you are running a compiler/OS that was written before 1999 then you might be limited but I don't know what specific versions those are.

egthomas:
Copying this here since I originally thought I was commenting on this pull request:

At least a dozen radars in the network are still using QNX4 (circa 1997) for their radar control software. If we ever want to be able to port these updates to the dmap, rawacf, fitacf, or other libraries to those radars then we need to make sure we retain backwards compatibility with those sites. That is after all one of the strengths of SuperDARN - common analysis software.

mts299:
Never mind, it would be nice if issues were addressed sooner so we don't waste people's time.

asreimer:
I don't think it is time just yet to close this issue.

mts299:
# Time results

Here are my result for compiling (`make.code`) and testing with `make_fit` for **C90** vs. **C99**. To ensure no further discussion around adding `-pedantic`, I have also timed that with and without. 

Please note I am using the `time` command. It will print out 3 times:
- **real**: wall time so when I hit `enter` to when the program finishes 
- **user**: CPU time spend on running the program 
- **sys**: CPU time spent in kernel mode

So the time we are looking at is **real** time as that is the execution time of the program. 

## setup
OS - Ubuntu 18.04 
GCC version: 
```
gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
I am running this on my laptop that has 8 GB of RAM and Intel i5 processor, so nothing too powerful. 

## C90

### Compiling
`develop` branch 
``` bash
real	1m27.960s
user	1m12.417s
sys	0m15.235s
```

### FITACF generation
**fitacf 2.5**

``` bash
time make_fit 20170410.1801.00.sas.rawacf > 20170410.1801.00.sas.fitacf
real	0m0.698s
user	0m0.645s
sys	0m0.044s
```

**fitacf 3.0**

```bash
time make_fit -fitacf-version 3.0 20170410.1801.00.sas.rawacf > 20170410.1801.00.sas.fitacf
real	0m2.416s
user	0m2.336s
sys	0m0.064s
```

## C99 without Pedantic flag
### Compiling
``` bash
real	1m18.073s
user	1m5.758s
sys	0m12.554s
```
### FITACF generation
**fitacf 2.5**
``` bash
time make_fit 20170410.1801.00.sas.rawacf > 20170410.1801.00.sas.fitacf
real	0m0.721s
user	0m0.650s
sys	0m0.064s
```
**fitacf 3.0**
``` bash
time make_fit -fitacf-version 3.0 20170410.1801.00.sas.rawacf > 20170410.1801.00.sas.fitacf
real	0m2.368s
user	0m2.321s
sys	0m0.032s

```

## C99 with Pedantic flag

``` bash
real	1m18.244s
user	1m6.747s
sys	0m12.067s
```
### FITACF generation

**FITACF 2.5**
```bash
time make_fit 20170410.1801.00.sas.rawacf > 20170410.1801.00.sas.fitacf
real	0m0.675s
user	0m0.647s
sys	0m0.012s
```

**FITACF 3.0**
```bash
time make_fit -fitacf-version 3.0 20170410.1801.00.sas.rawacf > 20170410.1801.00.sas.fitacf
real	0m2.042s
user	0m2.013s
sys	0m0.020s
```

## Conclusions

1. Is there no is conclusive show in a change of `make_fit` performance with the compiler change. 
2. The overall `mak.code` showed a ~9-second decrease in using **C99**. (yes the times vary but there was always a decrease in **C99**) 
- Please note a scientist taught in **C99** or even **C++** will run into errors when coding as **C90** will error even on inline comment syntax `//` this will cause more time and frustration for the scientist trying to implement code. 
- Someone who codes in **C90** will not be errored out by **C99** compiler. 
- To add some other information on the compiler differences:
https://www.electronicdesign.com/technologies/dev-tools/article/21798314/whats-the-difference-between-c-now-and-then
- If we ever want to run algorithms with GPU's to increase performance as algorithms will start to take more time, we are going to need to be compiling in C11. Food for thought :p 

## Final Remark
Regards to any more **time** on this PR, I am done (unless we are going forward and you need to fix/document something up). It is very demotivating when one needs to spend more **time proving themselves** then the **time it took to code** this and provided links. I think we are at the stage of testing porting and/or push it forward, and see what happens or close another PR. At this point, I am done proving my point and wasting my time doing so when there are tons of blogs, and publications were done by Computer scientists showing this fact. 

If you have a general question on compilers, I have no problem answering them on the issue I made as I have taken a grad course on compilers and built a compiler.

mts299:
Where are we on this? As @ecbland and 1 write developer guidelines and re-design fitacf 3.0 I would like to know if I should add in the new syntax the C99 allows like inline comments:+1:

ecbland:
@mts299 This might take some time to resolve because people are away on leave. Is it ok to wait a few weeks?

mts299:
Sure I don't mind delaying my work on Fitacf 3.0 some more. 
Again I have said enough on this PR it gets pushed or it doesn't. 
I just need to know more about future directions.

ksterne:
I ran the same tests on Ubuntu 20 (gcc version 9.3.0) VM and didn't see much of a significant hit in performance.  For Ubuntu 20 stats:  

- when compiling on `develop` see 191 warnings; on this branch see 263 warnings
- sample size of one for compilation time was a little over 3 seconds faster on this branch
- sample size of a couple noticed a `make_fit` command took about the same time (sometimes slower, sometimes faster) between the branches.  There might be a slight trend of being faster with this branch.

This PR looks good to go, but I'll leave it for a few days for any additional tests or feedback here.

ecbland:
I've tested this on Linux Mint 18 & 20. Compile time was about this same on this branch and `master` (within 1s), and also no difference to run `make_fit`.

Issue #302, opened 2020-06-08
FOV bug in grid_plot


ninaer:
When trying to plot fov in a grid plot using:
`grid_plot -fov somedate.north.grd`
All radar field of views are plotted, and not just the ones that contains data. This error message also occurs in the terminal:

> ERROR: coefficients are not valid for altitudes above 2000 km: 2016.378259.
       You must either use field-line tracing (TRACE or ALLOWTRACE) or
       indicate that you know this is a very bad idea (BADIDEA)

I use rst version 4.3.2 and ubuntu version 18.04.4 LTS

mts299:
@aburrell @billetd @ecbland @egthomas anyone know about this?

egthomas:
Sorry, this functionality was never added to `grid_plot` - it was only added to `map_plot` via the `-gfov` option.  It probably wouldn't be difficult to port to `grid_plot`.

ecbland:
Closing this now that #319 has been merged. The new functionality has been added to `grid_plot`, but we have not addressed the error message. I will open a new issue for that. Thanks @egthomas for implementing this new feature!

PR #187, opened 2018-10-01
map_plot -gfov option

egthomas:
This pull request adds a new `-gfov` option which tells `map_plot` to only draw and/or fill FOVs for the radars which provided grid vectors for that particular record (in combination with `-fov` and/or `-ffov`) .

PR #319, opened 2020-08-10
grid_plot -gfov option

egthomas:
In response to #302 and similar to #187 for `map_plot`, this pull request adds a `-gfov` option which tells `grid_plot` to only draw and/or fill FOVs for the radars which provided grid vectors for that particular record (in combination with `-fov` and/or `-ffov`).

To test, you can use something like `grid_plot -x -def -fov [your_favorite_grid_file]` both with and without the `-gfov` option.  Without `-gfov`, the FOVs of all radars will be drawn, while with `-gfov` only the contributing radars' FOVs will be drawn.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/grid_plot.1.12/doc/grid_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/grid_plot.1.12/doc/grid_plot.doc.xml
@@ -112,6 +112,8 @@
 </option>
 <option><on>-ffov</on><od>plot the filled radar field of view.</od>
 </option>
+<option><on>-gfov</on><od>only plot fields of view of the radars contributing data (when using -fov and/or -ffov).</od>
+</option>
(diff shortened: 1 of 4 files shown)


ecbland:
This also works as described!

Note that the following error message still appears if I don't use the `-gfov` option, but the code still works as expected:
```
grid_plot -x -def -fov [my_second_favorite_grid_file]

ERROR: coefficients are not valid for altitudes above 2000 km: 2042.320600.
       You must either use field-line tracing (TRACE or ALLOWTRACE) or
       indicate that you know this is a very bad idea (BADIDEA)
```

This error message was reported in #302, but if you like we can treat that as a separate issue and I can merge this now.

egthomas:
Yeah it's not clear to me why the aacgm_v2 warning occurs when all the FOVs are included - I don't think any current SD FOVs are  particularly close to AACGM's undefined region.  That warning message does occur quite often when trying to draw coastlines for an entire hemisphere (at least in IDL).

ecbland:
@egthomas Ok, let's say that this error is out-of-scope for this PR since it relates to several other RST functions too. Merging now!

Issue #310, opened 2020-07-09
BUG: compilation failed at stdkey.1.5


aburrell:
Just went to upgrade to the last release of RST on my processing system.  This required updating to the latest version of CDF (3.8), and making a change to one of the makefiles:

In `$RSTPATH/codebase/general/src.lib/stdkey.1.5/src/makefile` `SLIBS` and `LIBS` are both equal to `-lm`, which results in the following command that errors:
```
================================================================================
Compiling Library:$RSTPATH/codebase/general/src.lib/stdkey.1.5
================================================================================
make clean
rm -f *.o
make
rm -fv *.o
rm -fv paltable.c
rm -fv make_palette
removed `make_palette'
cc -o make_palette make_palette.c -I$RSTPATH/include/base -lm
./make_palette
making spectra.
making color bars.
printing palette table.
printing rgb tables.
make.version $RSTPATH/codebase/general/src.lib/stdkey.1.5/src/..
make.hdr $RSTPATH/codebase/general/src.lib/stdkey.1.5/src/../include $RSTPATH/include/general 
$RSTPATH/codebase/general/src.lib/stdkey.1.5/src/../include/stdkey.h
`$RSTPATH/codebase/general/src.lib/stdkey.1.5/src/../include/stdkey.h' -> `$RSTPATH/include/general/stdkey.h'
cc -fPIC -Wall -O3 -ansi -D_GNU_SOURCE -D_LINUX -I$RSTPATH/include/base -I$RSTPATH/include/general -c paltable.c    
mkdir -p $RSTPATH/lib
ar -r $RSTPATH/lib/libstdkey.1.5.a paltable.o
ld -shared -lm -lz  -o /$RSTPATH/lib/libstdkey.1.5.so \
              paltable.o -Bstatic -L$RSTPATH/lib -lm -lm
ld: cannot find -lm
make: *** [stdkey] Error 1
Compilation Aborted.
```

Commenting out those two environment variables results in this command that succeeds:
```
================================================================================
Compiling Library:$RSTPATH/codebase/general/src.lib/stdkey.1.5
================================================================================
make clean
rm -f *.o
make
rm -fv *.o
rm -fv paltable.c
removed `paltable.c'
rm -fv make_palette
removed `make_palette'
cc -o make_palette make_palette.c -I$RSTPATH/include/base -lm
./make_palette
making spectra.
making color bars.
printing palette table.
printing rgb tables.
[...]

PR #323, opened 2020-08-13
fixed the bug in stdkey makefile

mts299:
Not sure how this slipped through the cracks but I think it is fixed. 

I needed to `include rmath.h` in make_pallette.c to make constants consistent. However, this makefile was different than others and I think I was trying to fix the error I got with it not finding `rmath.h`. That fix was including the base.  

## Test

``` bash 
git checkout master 
make.build; make.code
```
should get this error 
``` bash
paltable.o -Bstatic -L/home/marina/superdarn/rst_dev/rst//lib -lm -lm
ld: cannot find -lm
ld: cannot find -lm
make: *** [/home/marina/superdarn/rst_dev/rst//build/make/makelib.linux:25: stdkey] Error 1
``` 

```bash
git checkout FIX/stdkey_makefile
make.build; make.code 
```
should be a complete build

### DIFF
--- a/codebase/general/src.lib/stdkey.1.5/src/makefile
+++ b/codebase/general/src.lib/stdkey.1.5/src/makefile
@@ -5,10 +5,6 @@
 #
 include $(MAKECFG).$(SYSTEM)
 
-INCLUDE=-I$(IPATH)/base -I$(IPATH)/general
-SRC=make_palette.c
-SLIB=-lm
-LIBS=-lm
 OUTPUT=stdkey
 DSTPATH=$(LIBPATH)
 OBJS=paltable.o


mts299:
As a hot fix this should go to master as I mentioned in this issue #310 If not I can change it to develop

ksterne:
You've got this right on all aspects.  Though I can't vouch that this solves the issue.   I think @aburrell, you're the best situated to test this?

mts299:
What was the issue?

ksterne:
Sorry, I should clarify.  I didn't see the bug so I don't think I can test it out.

mts299:
The test shows the bug :/ Unless when you compile on master it doesn't throw an error, is that what you are seeing?

egthomas:
> Unless when you compile on master it doesn't throw an error, is that what you are seeing?

@mts299 yes I think that's right - it seems to be system (or compiler?) dependent.  For example, the [Travis-CI build](https://travis-ci.org/github/SuperDARN/rst/builds/705838971#L2374-L2405) of the `master` branch (Ubuntu 16.04.6 LTS / gcc 5.4.0) doesn't show any errors, however Angeline and Alex both experienced complete failures on `make.code`.

ksterne:
Since this is a patch version, we'll want to change the `rst.version` file right?  I can do this if it makes things easier.

PR #315, opened 2020-08-05
feature/snd library and binaries

egthomas:
This pull request adds C, IDL, and DLM library support and two new binaries (`make_snd` and `trim_snd`) for working with the new dmap `snd`-format frequency sounding files produced by the `interleavesound` control program, which is scheduled for a network-wide experiment early next month (September 2020).

To provide some context, at the ROS-style radars a `snd`-format file is produced on-site in real-time by the `interleavesound` control program and contains a subset of the radar parameter block and fitted parameters (eg, velocity, power, spectral width, phi0).  The `make_snd` binary is intended to support multi-channel or non-ROS style radars which cannot produced fitted parameters on-site.  Therefore, `make_snd` converts a `fitacf`-format (produced off-site from a `rawacf` file via `make_fit`) file to a `snd`-format file for more consistent data analysis.

To test the new library and binaries, you should be able to produce a `snd`-format file from any `fitacf`-format file using `make_snd`, and then use `dmapdump` to view the contents via the command line, or use the C or IDL (and/or DLM) functions to open and read the data in software.  Note that `snd` files are incompatible with the standard RST plotting routines for `fitacf`-format files.

*edit* - `snd` files can now be plotted with either `time_plot` or `field_plot` by including the new `-snd` option.

PR #327, opened 2020-09-01
combine_snd and updates to sounding library

egthomas:
Apologies for not having this ready in time before @ecbland so graciously merged the larger sounding library pull request (#315). This pull request adds a new `combine_snd` binary which combines sounding records from each slice of a Borealis-style radar into a single sounding file in chronological order.  An example might look like

`combine_snd -vb YYYYMMDD.hh.rad.*.snd > YYYYMMDD.hh.rad.snd`

I do not believe I am in a position to release any example Borealis data, so maybe one of our colleagues from USask could test this routine or provide access to the data if necessary.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/snd.c
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/snd.c
@@ -77,7 +77,7 @@ void snd_tplot(struct SndData *snd,struct tplot *tptr) {
   int i;
 
   tptr->bmnum=snd->bmnum;
-  tptr->channel=0;
+  tptr->channel=snd->channel;
   tptr->cpid=snd->cp;
   tptr->scan=snd->scan;
   tptr->nrang=snd->nrang;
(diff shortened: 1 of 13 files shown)


ecbland:
I'm very happy to test this PR if someone can provide some sample Borealis data :) 
@kevinkrieger @mts299 ?

ecbland:
I've tested this using some sample Borealis data that @egthomas provided from his experiment (sas, cly & pgr). The output files look correct, although I did notice that the `time_plot` output looks a bit funky---the size of the symbols varies in the RTI plot and the frequency bar above it. That's not related directly to this PR though.
20200915 20 pgr

egthomas:
Yes, unfortunately that's because of how time_plot handles the size of each pixel for irregularly sampled data - which the sounding data are, since only a few records are collected at the end of each minute. field_plot has some similar issues given the nature of sounding scan boundaries compared to a 1- or 2-min normal scan.

I just realized I forgot to update the origin fields in combine_snd, so if you can hold off on merging for a few more hours I can get that fixed.

egthomas:
Just to demonstrate that `time_plot -snd` does look more reasonable when the sounding data is regularly sampled, here is a figure from the Buckland Park radar which had a channel dedicated to frequency sounding during our September experiment:

20200909 bpk

ecbland:
> Ok, combine_snd should now correctly set the origin fields so this pull request should be good to go.

Yes, it does. Well-spotted! Merging now 😄

Issue #326, opened 2020-08-24
Testing Scripts 


mts299:
I know C doesn't have the fancy `unittests` but I bet a bunch of us have made bash, python scripts to test RST code. I was wondering if we could make a test folder to store those files? 

I just don't want to lose my files and might be handy to see what others are doing for testing :)

mts299:
*poke* *poke*

mts299:
Gonna close to this for now as we need a test data set first which is an issue on DAWG.

PR #328, opened 2020-09-02
fitacf 3.0 minimal refactor

mts299:
# Scope
To only remove the llist_for_each functions that cause a lot of type coupling and complexity in C programming language.  

I also removed other functions that were not being used to clean up the code a bit. 

This PR does not include the new elevation algorithm nor should it change the output of fitacf files

## Testing 
I have my own script using pyDARN that compares the two files if anyone wants it #326 
These are the lines of code I ran:
 ``` bash
git checkout develop
git pull origin develop
make.build; make.code
make_fit -fitacf-version 3.0 <rawacf file> > <fitacf_file>

git checkout EHN/fitacf_3.0_restructure
git pull origin EHN/fitacf_3.0_restructure
make.build; make.code 
make_fit -fitacf-version 3.0 <rawacf file> > <fitacf_file_restructure>
```
then compare the files... 
Expect no change in the files. 

## Future tasks
- [x] implement Simon's Elevation code into Fitacf 3.0 

## Note 
This version is roughly 0.025 s

Also, I noticed the library @kkotyk  copied the listed code from its Apache license which GNU states all code under it must be GPL v3.0 as well. So I suggest we wait for all modifications to be done before releasing this change. I will be removing the list file in the next modification.

### CODE REVIEW

**Comment by mts299** [comment]
@ecbland would you mind converting this to a draft PR. My thought is we will keep merging to  `develop` but as a draft for now. When I branch off this PR to do steps 2 and 3 we will merge them to this PR and once that is done re-open this PR for merging. Does that make sense? 

Not sure I can convert this PR to a draft ... or I am blind :p

**Comment by mts299** [comment]
NVM It's just in small print :) but you can still test this PR

**Comment by ecbland** [comment]
Thanks for getting this started, @mts299!
I'm getting different values for `elev_low` and `elev_high` between the two branches. Do you see this behaviour as well? I saw some "TODO" notes in the code where the phase unwrapping is done - perhaps that has something to do with it?

**Comment by mts299** [comment]
@ecbland I am not getting a difference between `elv_low` and `elv_high`

**Comment by ecbland** [comment]
@mts299 I've tested this for several `rawacf` files and the output is identical on both branches :100:

**Comment by mts299** [comment]
I will start adding in Simon's code and then reconvene on if anything else needs to be restructured or not.

File: codebase/superdarn/src.lib/tk/fitacf_v3.0/src/determinations.h [hunk 1]
```diff
@@ -46,13 +49,63 @@ struct FitRange* new_range_array(FITPRMS* fit_prms);
 void ACF_Determinations(llist ranges, FITPRMS* fit_prms,struct FitData* fit_data,double noise_pwr);
 
 void lag_0_pwr_in_dB(struct FitRange* fit_range_array,FITPRMS* fit_prms,double noise_pwr);
+
+/*
+ * Setting the quality flag in fitted data. This determines if the data is good or bad. 
+ *
+ * @param fit_range_array: list of records from RAWACF 
+ * @param fit_prms: Fitacf parameters
+ * @param noise_pwr: noise power in dB
```
ecbland, line 58: `noise_pwr` has arbitrary units

```suggestion
 * @param noise_pwr: noise power
```

mts299:
@ecbland what is the status of this shall we merge it?

ecbland:
@mts299 Sorry for not responding earlier. I noticed you added this to the PR description in December: 

> Also, I noticed the library @kkotyk copied the listed code from its Apache license which GNU states all code under it must be GPL v3.0 as well. So I suggest we wait for all modifications to be done before releasing this change. I will be removing the list file in the next modification.

Is there any more work to do on this before merging?

mts299:
Nope I think its all good now

ecbland:
Ok, I've tested this again in a similar way to #356. Both versions of FITACF were tested against `develop`, with and without the `old_elev` option.

- `20150301.2001.00.cly.rawacf` (simple Y interferometer offset)
- `20140214.2201.00.inv.rawacf` (X and Y interferometer offset)
- `20180301.0601.00.zho.rawacf` (X, Y and Z interferometer offset)
- `20180101.0001.00.dce.rawacf` (example from issue #345)
- `1997081004g.dat` (Goose Bay special case for `old-elev`, and to test this with `make_fit -old`)
- `1997081004g.dat` converted to rawacf format
- `20140301.0401.00.wal.rawacf` (no interferometer data)

For INV, ZHO & GBR(g) we expect different elevation angles for fitacf3 between the two branches, but they are within a degree or two. 

GBR with `-fitacf-version 3.0 -old_elev` is also different. This is expected because FITACF3 does not call `elev_goose()`

Everything else is identical on both branches. So this is good to go! 
Thanks @mts299! :100:

Issue #330, opened 2020-09-03
South grid_plot not plotting pwr/swg


mts299:
# Description

Wang Guowe from PRIC, contacted me about this issue. His `2014012.grd` file looked correct, so I followed up with my own `grd` to see if I can recreate the bug and I did! 

## Code 

Here are my steps:
``` bash
make_grid -tl 60 -xtd -c 20140126.*.hal.fitacf > 20140126.hal.grid 
combine_grid 20140126.*grid > 20140126.south.grd
grid_plot -png -coast -fcoast -rotate -latmin 30 -mag -grd -grdontop -tmlbl -time -pwr -xkey rainbow.key -xkeyp 20140126.south.grd
```

### Notes

Note: made a grid file for: 
- mcm
- dce
- hal
- ker
- san
- sps
- sye
- sys
- tig
- unw
- zho

based on FITACF 2.5 files following the tutorials on RST readthedocs 

## Output

Here are the plots I get:

`grid_plot -png -coast -fcoast -rotate -latmin 30 -mag -grd -grdontop -tmlbl -time -pwr -xkey rainbow.key -xkeyp 20140126.south.grd`

![0670](https://user-images.githubusercontent.com/6520530/92141421-a9bd7380-edcf-11ea-97c1-ed3a0b36ae25.png)

`grid_plot -png -coast -fcoast -rotate -latmin 30 -mag -grd -grdontop -tmlbl -time -raw -xkey rainbow.key -xkeyp 20140126.south.grd
`
![0670](https://user-images.githubusercontent.com/6520530/92141319-82ff3d00-edcf-11ea-82f5-6ed561f2fae3.png)

So from `-raw` there is some data in there and looking at the file there are power and spectral width data as well, however, they are not being plotted. 

### Extra Information

Running this off develop branch. They, also got the same results using RST 4.2:

>Dear Marina.
 Hi, this is Wang Guowei,hope the virus doesn't bother your life. Nowdays I want to use several radars to research the ionosphrer of south hemisphere. When I input the command grid_plot ... -pwr ... The power can't be plotted but if it in north hemisphere,it is normal.The cells are fulled in th north but are empty in the south. The spectrual width has the same bugs.Do you know why this happens,and if it is my fault? Thank you! And the grid file is attached.

PR #331, opened 2020-09-03
Fixing bug in grid_plot and map_plot preventing power or spectral width cells from being drawn for SH data

egthomas:
This pull request addresses the bug reported in #330 by correctly passing the latmin value from `grid_plot.c` to the `plot_cell` function. Previously, southern hemisphere grid vectors were skipped over because their latitude was always below the hardcoded `latmin` value of 0 passed to the `plot_cell` function.

edit - this pull request also addresses a similar bug in `map_plot` causing the same behavior.


mts299:
I see this relates to the issue I reported #330

mts299:
Looks like it fixed the bug in #330 using the same code but producing a plot.

mts299:
@egthomas can I merge it or wait for others to test?

Issue #322, opened 2020-08-13
AACGM_v2 error: "coefficients are not valid for altitudes above 2000 km"


ecbland:
The `grid_plot` routine sometimes returns the following error from AACGM_v2:

```
grid_plot -x -def -fov [grid_file]

ERROR: coefficients are not valid for altitudes above 2000 km: 2191.461117.
       You must either use field-line tracing (TRACE or ALLOWTRACE) or
       indicate that you know this is a very bad idea (BADIDEA)
```

In the above example, the error occurs when the `-fov` option is included. The routine still works and the plots look ok. 

This might happen with other RST routines as well - please add them in the comments if you find them!

See also #302

egthomas:
This appears to happen for the SAS, PGR, INV, RKN radars only - and I think it's because their max range is defined as 225 gates in their hdw.dat files, which is used by `grid_plot`'s `make_fov` in the call to `RPosMag`, eg

https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.bin/tk/plot/grid_plot.1.12/make_fov.c#L65)

`field_plot`, `grid_plot`, and `map_plot` all appear to use `site->maxrange` from the hdw files when calculating the radar FOVs.  One solution could be to set an upper allowable limit on the for loops using `site->maxrange` to something like 110 range gates.

aburrell:
The simplest way to change this is to default to using `ALLOWTRACE`.  As defined in aacgmlib_v2.h, `ALLOWTRACE` works by: "if height is >2000 km use tracing, else use coefficients".

egthomas:
My only concern is that the FOVs drawn by those routines are possibly meant to be more representative than wholly accurate, as they already use hardcoded "standard" values of `rsep` (45 km) and `frang` (180 km).

Or looking at it from another perspective, do the Canadian radars regularly use 225 range gates or is that more of a theoretical maximum?  So should the software draw a FOV with all 225 ranges all the time?  For example, I know we have used 200+ range gates when running 15 km (or less) range resolution modes at the CV radars but have kept a maximum range of 110 in the hardware file as that's what we use 99% of the time.

aburrell:
> Or looking at it from another perspective, do the Canadian radars regularly use 225 range gates or is that more of a theoretical maximum?

From what I have seen, a maximum of 100 is typical for INV and CLY.

asreimer:
There are experiments running 15 km range gates at INV and CLY which is why
[...]

PR #332, opened 2020-09-03
Fix max FOV range in plotting routines to 75 range gates

egthomas:
This pull request addresses issue #322 by modifying the `make_fov` routines in `field_plot`, `grid_plot`, and `map_plot` plot to use 75 range gates to create representative FOV extents and avoid AACGM-v2 errrors for radars with >110 max range gates in their hardware files. Note that this was fix previously done in `fov_plot` but the implementation in that routine has been modified to match the other plotting binaries.


mts299:
Tested this PR as well using:
`grid_plot -x -def -fov -coast -fcoast -rotate -latmin 30 -mag -grd -grdontop -tmlbl -time -swd -xkey rainbow.key -xkeyp 20140126.grd`

and no Error message appeared when I switched to this branch.

mts299:
I can merge unless @ecbland you want to test it as well?

ecbland:
@mts299 Merge away...

PR #333, opened 2020-09-03
EHN/socket version removal

mts299:
Making this a draft, for now, to check if the commit messages get stacked when we rewrite them or its actually showing them as a diff


mts299:
Not finding anything to squash the extra 300 commits except `git rebase` but not even sure that works because I cannot find the duplicates...

mts299:
Reopening this PR to reconsider this method or a simple `git mv`

mts299:
closing again as `git mv` looks to be the better option and agreed upon last telecon.

PR #334, opened 2020-09-10
What would it really take to remove hardware files but keep them? 

mts299:
# Scope 

As many scientific legends taught me (Albert Einstein, Marie Curie, etc) when people cannot see your visions you must prove it. Let's start thinking like scientists and start experimenting and listening to those that have a strong background in this stuff.  

Here it is a PR on showing how one can simply use submodules in git to pull down specific tags of another repository with tag versions. 

The hardware repository is a test one I made, and hardware files are not in RST anymore for this PR. 
This is a draft PR and **SHOULD NOT BE MERGED** this is just a proof of concept. 

## How to install RST? 
Does this change how RST installs? Nope :) 

`make.build; make.code`

What if I want to change the version? 

`make.build 1.0; make.code`

## Reproducibility

As a good scientist at least in my line of work with Numerical Analysis we had to document everything even so if a person uses the "default" RST version they can double-check the `make.build` the script, which can be documented on the README and other forms of documentation. 

If that is not loud enough we can even put a warning after making the code to ensure they know which version is being used and grab the DOI. 

If they do change the version this is on them, again there is some responsibility of the scientist to report what they use and change. Not even map files have a straight cut make method, so you need to list all the options. 

What about Zenodo? 
So what? I really am not fully sure why hdw files need to be in zenodo? If they are at least tagged then we have a reference in RST and hopefully DOI in which hardware files were used in the release. Every release we would just update the default tag number so it is like have the hardware files with RST. 

If someone downloads Zenodo zip of RST they still need to install it to use RST thus using `make.build` so they still get the hardware files. Don't believe, give this PR a test zenodo publication and see.

What if they are using RST for hardware files and cannot get them from Zenodo? 
Good! we don't want them to look at RST they should be going to DOI/GitHub repo the official location. 

Think about this do we include the Radar OS code in RST or IMF data for MAP files? No, but the scientist will reference that metadata if important. 

### Conclusion 

Can we please stop saying "leaving hardware files in RST" is for ease of installing? Reproducibility yes while there is no tag or DOI on them. Once that happens then we can revisit this idea.  
Can we please start thinking "what is going to be the best for the scientist to look at SuperDARN hdw repo and cite the DOI?"
Let's work with DSWG and instead of repeating a VT repo issue again. 

**PLEASE** this PR is not to start a discussion but a proof. Like staining wood one needs to let it set in for a couple of days and when you have questions about how things work I am more than willing to answer. 

Hopefully, this answers all the feedback I have heard from people on this issue. 

Please note this took me 30 minutes to code can we also start trying to ask people who are experts in this kind of work "how hard is it really?".


build/script/make.build
```diff
@@ -26,6 +26,11 @@ if [ -z "$BUILD" ]; then
     exit 1
 fi
 
+TAG_VERSION=$1
+if [ -z ${TAG_VERSION} ]; then
+    $TAG_VERSION=0.0
```
ecbland, line 31: I got an error here...perhaps the `$` should be removed?
```suggestion
    TAG_VERSION=0.0
```

build/script/make.build
```diff
@@ -73,5 +78,11 @@ for binname in ${binlist}
 do
   makebuild ${binname} bin
 done
+
+cd ${RSTPATH}/tables/superdarn/hdw/ 
+git checkout ${TAG_VERSION}
```
ecbland, line 83: I got an error here when I ran `make.build`: 

`error: pathspec '0.0' did not match any file(s) known to git`

So I think the script can't find the subrepo. Have I done something wrong? I get the same error with `make.build 1.0`

ecbland (commented):
Thanks for the demo, @mts299!  
I ran into some errors in `make.build`, so I might need some help with this :smile_cat:

build/script/make.build
mts299: do you have a .gitmodules? in your main RST directory?

mts299: You can also check by going to the hdw directory and going git branch and see what comes up.

mts299: It also may not work for you because of being a repo of mine... this is an example of how easy things could be to get to work not necessarily wanting to spend a lot of time troubleshooting until you actually want to consider the idea for the hdw. 

Last I heard is it is currently staying the same.

mts299:
Sorry if this PR, has a harsh tone. This PR, is due to ongoing conversation of "ease of installment" not happening when we remove hardware files. However, this is a myth so I wanted to bring light to it. With that said because it is not something we currently doing right away, I am only using this PR as a show and tell and have little time right now to dedicate to it as it isn't a priority yet :) 

Try to read with an open mind and let the stain take some time. I rather have an in-person discussion around it then a long GitHub discussion. 

Sorry if the testing fails I did this in a rush.

PR #335, opened 2020-09-16
Consolidating plotting functions into a single library

egthomas:
This pull request consolidates many of the duplicated functions used by the various plotting binaries into a single library (`binplotlib`).  For example, on the `develop` branch there are currently 23 `.c` and `.h` files in the `grid_plot.1.12` directory, while on this branch there are only 3 `.c` and `.h` files (because of duplication with `map_plot`, `field_plot`, etc).  The modified plotting binaries are:

- `field_plot`
- `fov_plot`
- `grid_plot`
- `map_plot`
- `time_plot`
- `vec_plot`

however no functionality should be affected.  (I had started this effort ~3 years ago on the `feature/plot_lib` branch and am just now wrapping it up / updating it to avoid the numerous merge conflicts that have since appeared with that branch.)


mts299:
Curious: 

1. I noticed you put numbers on the folder name which is something we are looking at not doing anymore? 

2. Also, I see you did a move on several files which I thought we didn't want to do since I was looking into a way of removing version numbers without removing the history but you did a move. 

I am just trying to get some consistency on how we are moving forward with moving files a around and extra versioning controlling. @ecbland @ksterne thoughts on this as well? I can put this in an issue to remove the scope creep here. 

Otherwise, I please to see encapsulating and modularity, good job!

ecbland:
I can test this PR today or tomorrow. 

Any comments on the `git mv` and folder numbering that @mts299 asked about? (see also https://github.com/SuperDARN/rst/pull/336#issuecomment-693648411)
Can the folder numbering can be addressed separately when we have agreed on what to do? I'll defer to the experts here...

egthomas:
In many cases there were ~2 to ~5 copies of the exact same file in each plotting binary directory - I don't know which one was actually moved to the consolidated plotting library, but git decided to arbitrarily treat one of the files as having been moved and the rest deleted.  I suppose the alternative would have been to delete all of the duplicate files and commit that change, and then create a new file to avoid a git move operation.  Although I'm not sure I understand the issue - we've moved files from one directory to another many times in the past without any problem (e.g., the elevation library).

And yes, I included a library folder number because that is how the code on the `develop` branch is currently structured (I did the same thing with the sounding libraries and binaries in #315 and #327).

ecbland:
> @ecbland  I can test this PR today or tomorrow.

Sorry, I promised to look at this almost a month ago :calendar:

I've tested this for all of the routines listed above, and I can't see any differences between the output plots produced on this branch or `develop` :smile:  It wasn't really practical to test every possible combination of options for all these routines. Mostly I used the examples from readthedocs, where available. 

Nobody has raised any further questions about the way the files were moved. I'll leave this open for a few more days just in case, but I think @egthomas has made a fair point that we've done this before (and we don't actually have a guideline about this)

ksterne:
I think it's a fair point about continuing the version numbering which is redundant or maybe unnecessary with git.  Though I think removing this directory structure numbering might be something that we want to do as a major project undertaking and try to do that all at once (though not suggesting one big pull request).  It is good to see how this would work with #336.

egthomas:
@ksterne the current `develop` and `master` branches both look for library/binary version numbers for compilation.  Other recently approved pull requests (#351, #327, #315, etc) also maintained this existing directory structure, so it seems odd to start applying that criteria now (particularly when #336 and #333 have not been merged).

ksterne:
Sorry if I wasn't clear. I meant *major* project meaning major version update.  I do not think it's necessary for our next minor release.

ecbland:
@ksterne So you're happy for me to merge this? The testing is completed and it seems that everything else mentioned here is out-of-scope.

egthomas:
@ecbland I guess the thumbs up emoji means yes?

PR #336, opened 2020-09-16
removed rtfittofitacf.1.20 version number

mts299:
Given the PR #335 where we moved the plotting binaries and libraries to new folders, I am wondering why not do the same for version numbers. 

I opened the other PR #333  if someone wants to take a look at the differences.


mts299:
I should note, using `git mv` vs. `git filter-branch ....` is a lot easier because rewriting commit history (`git filter`) does come with some gotchas and I think that will save us some time converting. 

I am not 100% sure why we need the history in tack when it will show the mv later down the line and I mean it is kind of obvious at this point since the folder name changes via removing the numbers nothing more. So this kind of "so what?" moment for me. 

Then again if some type of reasoning that people feel the history needs to be rewritten and want it that badly we could do it.

PR #349, opened 2020-10-13
Modifying time_plot to more accurately plot range-time cells for SND-format data

egthomas:
As the title implies, this pull request addresses the issue mentioned by @ecbland in #327 where `time_plot` did not correctly draw the range-time cells for SND-format data.  This is because for most radars, the sounding data are collected for only a few seconds at the end of each scan, rather than with the regular cadence of normal scan data.

For example, on the current develop branch `time_plot -snd` will produce something like this (using only ~15 minutes of real frequency sounding data from today's experiment):

cve snd old

while on this branch, the same `time_plot` command will produce this:

cve snd new


egthomas:
I've just added one more commit to this branch which allows for correct plotting of `phi0` values from `snd`-format files.  For example, using a command like `time_plot -x -a -phi -snd` on the `develop` branch you would see something like this

cve_old

where all of the `phi0` values appear to be zero (green).  On this branch, you should instead see something like this:

cve_new

where the `phi0` values are now being plotted correctly.

ecbland:
This works as described, both in terms of fixing the symbol size and the `phi0` values :+1:

PR #356, opened 2020-10-27
Simon's elevation code into fitacf 3.0

mts299:
So I have begun implementing Simon's elevation code into Fitacf 3.0; however, I need some help and guidance on a few things. 

@ecbland if you are around if you can look at this code and then meet with me video-wise to discuss a few things and help me know if I am getting correct answers. 

After that, I will update the documentation here on how to test the code!

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
+++ b/codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
@@ -121,7 +121,7 @@ int main(int argc,char *argv[]) {
   int status;
   int arg;
   int return_value = 0; 
-
+  int elv_version = 2;
   unsigned char help=0;
   unsigned char option=0;
   unsigned char version=0;
@@ -189,6 +189,11 @@ int main(int argc,char *argv[]) {
     OptionPrintInfo(stdout,hlpstr);
     OptionFree(&opt);
     exit(0);
+  }
+  
+  if (old_elev == 1)
+  {
+      elv_version = 1;
   }
 
   if (fitacf_version_s != NULL) {
@@ -366,8 +371,12 @@ int main(int argc,char *argv[]) {
       /* If the allocation was successful, copy the parameters and */
       /* load the data into the FitACF structure.                  */
       if(fit_prms != NULL) {
-    	  Copy_Fitting_Prms(site,prm,raw,fit_prms);
-    	  Fitacf(fit_prms,fit);
+    	  if (prm->stid == 1)
+          {
+              elv_version = 0;
+          }
+          Copy_Fitting_Prms(site,prm,raw,fit_prms);
+    	  Fitacf(fit_prms,fit, elv_version);
         /*FitacfFree(fit_prms);*/
     	}
       else {
@@ -380,7 +389,7 @@ int main(int argc,char *argv[]) {
   }
   else if (fitacf_version == 25) {
     fblk = FitACFMake(site,prm->time.yr);
-    fblk->prm.old_elev = old_elev;          /* passing in old_elev flag */
+    fblk->prm.old_elev = old_elev;        /* passing in old_elev flag */
     FitACF(prm,raw,fblk,fit);
   }
 
@@ -463,7 +472,7 @@ int main(int argc,char *argv[]) {
         /* load the data into the FitACF structure.                  */
         if(fit_prms != NULL) {
           Copy_Fitting_Prms(site,prm,raw,fit_prms);
-          Fitacf(fit_prms,fit);
+          Fitacf(fit_prms,fit, elv_version);
           /*FitacfFree(fit_prms);*/
         }
         else {
--- a/codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
+++ b/codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
@@ -371,7 +371,7 @@ int main(int argc,char *argv[]) {
       /* If the allocation was successful, copy the parameters and */
       /* load the data into the FitACF structure.                  */
       if(fit_prms != NULL) {
-    	  if (prm->stid == 1)
+    	  if (prm->stid == 1 && elv_version == 1)
           {
               elv_version = 0;
           }
--- a/codebase/superdarn/src.lib/tk/elevation.1.0/src/elevation_v2.c
+++ b/codebase/superdarn/src.lib/tk/elevation.1.0/src/elevation_v2.c
@@ -17,7 +17,7 @@
 
 /* SGS somehow need to pass in options for allowing:                         *
  *     negative elevation angles and residual phase                          */
-
+//TODP what do these parameters mean  like psi_obs? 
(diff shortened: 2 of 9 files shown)


codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
```diff
@@ -366,8 +371,12 @@ int main(int argc,char *argv[]) {
       /* If the allocation was successful, copy the parameters and */
       /* load the data into the FitACF structure.                  */
       if(fit_prms != NULL) {
-    	  Copy_Fitting_Prms(site,prm,raw,fit_prms);
-    	  Fitacf(fit_prms,fit);
+    	  if (prm->stid == 1)
```
egthomas, line 374:
```suggestion
    	  if ((old_elev == 1) && (prm->stid == 1))
```

The `elev_goose` function should only be used for Goose Bay if the `-old_elev` option is specified and should not override the default v2 algorithm.

ecbland: Committing this suggestion now so that I can test it (and because it's correct :smile:)

codebase/superdarn/src.lib/tk/elevation.1.0/src/elevation_v2.c
```diff
@@ -17,7 +17,7 @@
 
 /* SGS somehow need to pass in options for allowing:                         *
  *     negative elevation angles and residual phase                          */
-
+//TODP what do these parameters mean  like psi_obs?
```
ecbland, line 20: `psi_obs` is the measured/observed ("obs") phase difference, in the range [-pi,pi). See [line 123](https://github.com/SuperDARN/rst/blob/665a554e0a037bc42f9d51ee3e1b170dc7cf90bc/codebase/superdarn/src.lib/tk/elevation.1.0/src/elevation_v2.c#L123)

This code estimates the true phase lag (many multiples of 2pi) from the observed phase difference, using the assumption that the true phase lag lies in the range (`psi_max-2pi`,`psi_max`].

`psi_max` is the maximum possible value of the phase lag. It depends on the frequency, the beam orientation, the main/auxiliary array positions, and `tdiff`

ecbland: Fixing typo in case you want to search for TODO's later
```suggestion
//TODO what do these parameters mean  like psi_obs? 
```

codebase/superdarn/src.lib/tk/fitacf_v3.0/src/determinations.c
```diff
@@ -527,7 +531,40 @@ void set_nump(llist_node range, struct FitRange* fit_range_array){
...
+    double azi_offset,phi_0;
+    RANGENODE* range_node;
+    range_node = (RANGENODE*) range;
+   
+    azi_offset = fit_prms->maxbeam/2 - 0.5;
+    phi_0 = fit_prms->bmsep * (fit_prms->bmnum - azi_offset) * PI/180;
+
+    // elv_version 2 is Simon's
+    // elv_version 1 is original
+    // elv_version 0 is goosebay
```
ecbland, line 545: Just some clarification for future developers: 

```suggestion
    // elv_version 2 is the Shepherd [2017] elevation calculation
    // elv_version 1 is original elevation calculation
    // elv_version 0 is specifically for the GBR radar when -old_elev is specified
```

codebase/superdarn/src.bin/tk/tool/make_fit.1.19/make_fit.c
mts299: I switched it to match my code style as I don't use old_elev as it can be miss leading in context.

codebase/superdarn/src.lib/tk/fitacf_v3.0/src/determinations.c
```diff
@@ -27,12 +27,14 @@ ACF determinations from fitted parameters
...
 #include "determinations.h"
+#include "fitblk.h"
+#include "elevation.h"
 #include <math.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include "rmath.h"
-
+#include "fitblk.h"
```
ecbland, line 37: This was already added on line 30
```suggestion
```

ecbland:
@mts299 I found the problem we were looking at yesterday. The beam separation was missing from `fitprm` in `determinations.c` Fixed with commit https://github.com/SuperDARN/rst/pull/356/commits/acbfbc801810f26a7133e603b98445ff60c161b9

Also, FITACF 3.0 should calculate the elevation angle using the lag zero phase. This is different to FITACF 2.5, which uses the fitted phase. I've corrected this in commit https://github.com/SuperDARN/rst/pull/356/commits/38e25f85cc82ec68ff71ac4ed93f88dcd1443c27 but feel free to revert/change if you want to implement this differently.

ecbland:
@mts299 and I believe that this code is now working, but some additional testers would be appreciated. 

I have tested the code with the following files, which represent the various interferometer layouts in the network:
- `20150220.1200.02.sas.rawacf`  (simple Y-offset)
- `20180101.2201.00.lyr.rawacf`    (Y & Z offset)
- `20150220.0801.00.zho.rawacf`  (X,Y & Z offset)
- `1996102922g.dat`                      (X & Y offset, unique function `elev_goose()` exists)

I ran `make_fit -fitacf-version 3.0` and  `make_fit -fitacf-version 3.0 -old_elev` on this branch and the base branch (`EHN/fitacf_3.0_restructure`). All results are as expected. 

- SAS radar has identical results with and without `-old_elev`
- With `-old_elev` option, LYR and ZHO have the same elevation values on both branches
- Without `-old_elev`, LYR and ZHO have different elevation values, as expected
- `elev_high` and `elev_low` are identical for all radars on both branches (no functionality has changed)
- Output files are identical for `-fitacf-version 2.5` (this code only modifies `-fitacf-version 3.0`

Note that GBR is a special case:
- On the base branch (and on `develop`), FITACF 3.0 does not include the `elev_goose()` function, so the horizontal offset is not accounted for
- On this branch, `elev_goose()` is called when `-old_elev` is specified. 
-  The Shepherd (`v2`) algorithm correctly accounts for the GBR interferometer layout, so `elev_goose()` is not necessary. There are some small differences in implementation of `elev_goose()` and `elevation_v2`. Therefore, the elevation angles for GBR are slightly different (less than 1deg)

mts299:
Thank you @ecbland for looking at this! 
@egthomas  or  @aburrell do you want to test this code? 

Remember this PR merges #328 so there is more time for testing and reviewing.

mts299:
@ecbland Kathryn said she could make an announcement to get others to test code if you want?

ecbland:
> @ecbland Kathryn said she could make an announcement to get others to test code if you want?

I think we can recruit some extra testers when we make a release branch that includes this code (probably not `v4.5`). But yes that would be helpful.

ecbland:
This code has been tested so I will merge it now. Note that this is being merged into `EHN/fitacf_3.0_restructure`, not `develop`

PR #358, opened 2020-10-30
Updating .travis.yml to allow for CI builds across Ubuntu 16.04, 18.04, and 20.04

egthomas:
As the title suggests, this pull request modifies the `.travis.yml` file to allow for simultaneous continuous integration builds across multiple versions of Ubuntu (16.04 / xenial, 18.04 / bionic, and 20.04 / focal).  Currently, Travis-CI defaults to 16.04 for its linux builds.  I guess the only real test is to make sure a green check mark appears for three successful builds.  And maybe there's a cleverer way to check the linux distribution for installing the correct libpng library.

Presumably, in the future this process could be expanded to include MacOS builds as well (although I'm much less familiar with that installation process):

- https://docs.travis-ci.com/user/reference/osx/
- https://docs.travis-ci.com/user/multi-os/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,57 @@
 language: C
+
+matrix:
+  include:
+    - os: linux
+      dist: xenial
+      env: TRAVIS_DIST_NAME="xenial"
+    - os: linux
+      dist: bionic
+      env: TRAVIS_DIST_NAME="bionic"
+    - os: linux
+      dist: focal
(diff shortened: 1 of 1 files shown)


ecbland:
Awesome, thanks @egthomas! I see 3 successful builds so I think this is working. I will merge this now so that we can start using it :) 

> And maybe there's a cleverer way to check the linux distribution for installing the correct libpng library.

No idea about this part...someone can open a new PR if they want to update it.

Issue #350, opened 2020-10-13
Unable to compile RST library


mtwalach:
Hi, 
I am still trying to install RST on a new OS (I have made an advancement on my last [issue ](https://github.com/SuperDARN/rst/issues/317) (🎉) but have now run into a problem that I have never come across before:

```
================================================================================
Compiling Library:/Users/maria/rst/rst/codebase/superdarn/src.lib/tk/sim_data.1.0
================================================================================
make clean
rm -f *.o
make
make.version /Users/maria/rst/rst/codebase/superdarn/src.lib/tk/sim_data.1.0/src/..
make.hdr /Users/maria/rst/rst/codebase/superdarn/src.lib/tk/sim_data.1.0/src/../include /Users/maria/rst/rst/include/superdarn 
/Users/maria/rst/rst/codebase/superdarn/src.lib/tk/sim_data.1.0/src/../include/sim_data.h
/Users/maria/rst/rst/codebase/superdarn/src.lib/tk/sim_data.1.0/src/../include/sim_data.h -> /Users/maria/rst/rst/include/superdarn/sim_data.h
cc -Wall -O3 -ansi -D_GNU_SOURCE -D_DARWIN -I/opt/local/include -I/Users/maria/rst/rst/include/base -I/Users/maria/rst/rst/include/general -I/Users/maria/rst/rst/include/superdarn -I/Users/maria/rst/rst/include/analysis   -c sim_data.c    
In file included from sim_data.c:37:
In file included from /Users/maria/rst/rst/include/superdarn/sim_data.h:5:
/Users/maria/rst/rst/include/base/rmath.h:44:1: error: declaration of anonymous struct must be a definition
struct complex {
^
/Users/maria/rst/rst/include/base/rmath.h:44:1: warning: declaration does not declare anything [-Wmissing-declarations]
1 warning and 1 error generated.
make: *** [sim_data.o] Error 1
Compilation Aborted.
```
I think this might be a problem in `/include/base/rmath.h` but I can't spot what is wrong with it - Can anyone help?

mtwalach:
Update: Apparently this is a known problem with the compiler on Catalina. I removed `make_sim.1.0`, `sim_data.1.0` and `sim_real.1.0` which all throw up this error and RST compiles fine without them. I have never used those routines before, so I won't miss them. If anyone does want to use these on Catalina however, they may run into this issue...

egthomas:
[...]

PR #359, opened 2020-10-30
Fix sim_* compilation on Catalina

ecbland:
This PR implements @asreimer's proposed "bandaid" solution to the compilation issue on MacOS/Catalina (#350):

> You might be able to add `-Wno-gnu-anonymous-struct` to the makefile for the `sim_*` code as a bandaid work around. (https://github.com/SuperDARN/rst/issues/350#issuecomment-709560674)

I'm not able to test this myself, but I'm hoping that @mtwalach can try it 😄

If someone wants to offer a better solution to #350 before we make the release branch, then please go ahead.


asreimer:
I would encourage removing the `struct complex` definition in `rtypes.h` instead. The `sim_*` can't even use that definition (which uses .x and .y to access the real/imag components.

mts299:
@ecbland I am fixing this problem via removing the complex struct and replacing it with complex.h 

I don't think this would be a good PR to add as then we would need to remember to remove this bandaid.

ecbland:
Oh, perfect. Didn't realise that a proper solution was in the works. I'll close this then

mts299:
Sorry for not communicating that I will assign myself on that issue and it should be out next week just need to debug a bit. 

Thank you for looking into this and getting something done!

Issue #353, opened 2020-10-25
field_plot does not plot Antarctica coastline


ecbland:
While testing #335, I noticed that `field_plot` does not plot the Antarctica coastline when the `-coast` option is provided. 

Example: 
`field_plot -x -coast -fan -keyp -v -tl 60 -grd -latmin 20 20180601.1001.00.zho.fitacf`

The filled coastline option (`-fcoast`) works fine.

mts299:
Poor South

PR #360, opened 2020-11-04
Fixing bug in field_plot preventing Antarctica coast lines from being drawn

egthomas:
This pull request addresses issue #353 where the coast lines for Antarctica were not being drawn by `field_plot` when using the `-coast` option.  On the current `develop` branch, you'll see something like this for stereographic (default) and orthographic coordinates:

mcm_dev_stereo
mcm_dev_ortho

while plotting the same data using this branch should now correctly draw the Antarctica coast lines with the `-coast` option:

mcm_fix_stereo
mcm_fix_ortho


ecbland:
Thanks @egthomas for putting Antarctica back on the map! Merging it in...

PR #340, opened 2020-09-17
Fix/fitacf old

mts299:
# Scope 
Fixing issue #338 

## Problem 
There was a double freeing of `fitfp`, so I assume one of the Old functions does the freeing. 

## Solution
Just passed in NULL into my free functions which will not free it from an if statement. 

## Test 
bug on branch `develop` 
```bash 
make_fit -old 2005050520d.dat 2005050520d.fit 2005050520d.inx
free(): double free detected in tcache 2
Aborted (core dumped)
```

fixed branch `FIX/fitacf_old`
```bash
make_fit -old 2005050520d.dat 2005050520d.fit 2005050520d.inx
time_plot -old -x -a 2005050520d.fit   #to check that there's some sensible data in the file
```

@ecbland can you add more for testing and what the bug error was? 
EDIT by @ecbland : I've edited the text to include the bug I get on `develop` and how to test :+1: 

## Note

I could not recreate this bug on my Opensuse and Ubunutu's OS on LinuxMint so far shows the error. So if you cannot get the bug then you might be the best tester :p 

@ecbland might be the only one hehe

PR #337, opened 2020-09-17
EHN/removing warnings

mts299:
# Scope
This PR focuses on warnings that are left by the compiler. Leaving warnings hanging leaves potential bugs and I might have found one or two actually in `time_plot` and `fitacffex2`. 

## What warnings did I fix?

- [ ] unused variables 
- [ ] void* arithmetic 
- [ ] void* pointer casting to a function pointer 
- [ ] unintialized warning 

Not all warnings were fixed but comments were left unless these files we not allowed to be touched. 

I did not fix the overflow warnings. 

## Which files did I look at?

I only looked at the RST code and left out the convection map code by @egthomas as requested last time a PR like this popped up. 

## simple tests

**Branch** `EHN/removing_warnings`
``` bash
make.build > /dev/null; make.code 2>&1 > /dev/null | grep -i "warning" | wc -l
47
```
**Branch** `develop`

```bash
make.build > /dev/null; make.code 2>&1 > /dev/null | grep -i "warning" | wc -l
170
```

Please make sure to test various functions to make sure no changes affected the code too much. 

## Environment
Due to compilers used on various machines, the warnings may not appear or change the number of warnings. 
- OS: opensuse leap 15.1 
- compiler gcc 
```bash 
gcc --version
gcc (SUSE Linux) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

PR #363, opened 2020-11-15
Fix double free error in lmfit

ecbland:
Looks like I didn't catch this one while testing #337 
It's the same double free error that @mts299 and I fixed earlier with `make_fit -old`(#340)

On `develop` we get a core dump:
```
make_lmfit [rawfile] > /dev/null
free(): double free detected in tcache 2
Aborted (core dumped)
```

There should be no error on this branch.

Note that you may not get the error with newer versions of gcc. I get the error with `gcc (Ubuntu 9.3.0-10ubuntu2) 9.3.0`


mts299:
Yea... I never touched `make_lmfit` except for complex changes so this might be an older bug.... 

good catch

mts299:
I got the same thing and it works with your fix!

PR #364, opened 2020-11-16
Adding missing chisham keyword to RadarFldPnth in native IDL RadarPos function

egthomas:
As the title says, this pull request adds a missing `chisham` keyword to the `RadarFldPnth` function call when using the native IDL version of `RadarPos` with only a single input range-beam pair (note that the `chisham` keyword is correctly passed when an array of range-beam locations are passed to `RadarPos`).  On the `develop` branch the standard virtual height model will always be used whether or not the `chisham` keyword is set, while on this branch the correct virtual height model will be selected.


ecbland:
Tested this and it works just fine...merging it in!

Issue #282, opened 2020-03-16
Constants values in IDL code


mts299:
As @egthomas mentioned and I pointed out in PR #281 that constants I did were just for C code and not IDL. However, there are IDL users. 

If anyone wants to take on this task and make its own PR please do so. I am not as familiar so I am a bit out of my element. 

Some places to look at:
https://github.com/SuperDARN/rst/blob/839181fff7f0f3b9a0c9959b5618de8e43e804e0/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L800
https://github.com/SuperDARN/rst/blob/839181fff7f0f3b9a0c9959b5618de8e43e804e0/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L942
https://github.com/SuperDARN/rst/blob/839181fff7f0f3b9a0c9959b5618de8e43e804e0/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L982
https://github.com/SuperDARN/rst/blob/839181fff7f0f3b9a0c9959b5618de8e43e804e0/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L995
https://github.com/SuperDARN/rst/blob/839181fff7f0f3b9a0c9959b5618de8e43e804e0/codebase/superdarn/src.idl/app/make_grid.1.0/gtable.pro#L418

mts299:
**poke** **poke**

mts299:
Anyone want to take this on that knows IDL? @billetd ?

mts299:
@egthomas @billetd @ecbland whose willing to take this take on :) all my IDL coders

ecbland:
This is not a priority for me at the moment due to time.

mts299:
Thanks, I am just trying to knock some issues down :) Not sure how many people use the IDL RST code

billetd:
Since IDL has been updated a fair amount since this code was written, basically all relevent constants are now held in the IDL system variable !CONST structure (e.g. !CONST.R_earth). I can test out a branch with these implemented, but it seems pretty straightforward and shouldn't break anything assuming people are using an IDL version from this decade.

PR #366, opened 2020-11-24
Updated IDL constants

billetd:
A relatively small PR, per issue #282, which makes the usages of Earths radius consistent throughout the IDL code. I've tried to preserve floats/doubles where applicable. I've left the legacy code unchanged, as it's uhhh, legacy.

If you find any others that I missed, let me know.

### DIFF
--- a/codebase/analysis/src.idl/lib/igrf.1.0/igrflib_v2.pro
+++ b/codebase/analysis/src.idl/lib/igrf.1.0/igrflib_v2.pro
@@ -76,7 +76,7 @@ pro init_common, err=err
                   cl0:0.d, sl0:0.d}
 ; DTOR        = !const.pi/180.d
   DTOR        = !dpi/180.d
-  RE          = 6371.2d     ; magnetic reference spherical radius from IGRF
+  RE          = !CONST.R_earth/1000     ; magnetic reference spherical radius from IGRF
 ;  IGRF_FIRST_EPOCH = 1900  ; these get set in IGRF_loadcoeffs
 ;  IGRF_LAST_EPOCH  = 2015
 
@@ -1128,7 +1128,7 @@ end
 function geod2geoc, lat, lon, alt
   common IGRF_v2_Com
 
-  a = 6378.1370d;             /* semi-major axis */
+  a = !CONST.R_earth/1000;    /* semi-major axis */
   f = 1.d/298.257223563d;     /* flattening */
   b = a*(1.d -f);             /* semi-minor axis */
   a2 = a*a;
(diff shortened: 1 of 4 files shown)


mts299:
Thank you for doing this @billetd

codebase/analysis/src.idl/lib/igrf.1.0/igrflib_v2.pro
```diff
@@ -1128,7 +1128,7 @@ end
 function geod2geoc, lat, lon, alt
   common IGRF_v2_Com
 
-  a = 6378.1370d;             /* semi-major axis */
+  a = !CONST.R_earth/1000;    /* semi-major axis */
```
egthomas, line 1131: The WGS-84 ellipsoid reference has an exact definition for the semi-major axis which should probably not be modified here or elsewhere (https://en.wikipedia.org/wiki/World_Geodetic_System#Defining_Parameters).

codebase/superdarn/src.idl/lib/main.1.25/radar.pro
```diff
@@ -797,7 +797,7 @@ function RadarPos,center,bcrd,rcrd,site,frang,rsep,rxrise,$
 
 
                 
-  re=6356.779D
+  re=!CONST.R_earth/1000
```
egthomas, line 800: Won't this be different from the Earth radius defined in `rmath.h` for the C code?

codebase/analysis/src.idl/lib/igrf.1.0/igrflib_v2.pro
billetd: Oops, didn't even cross my mind. I agree.

mts299:
@billetd are you able to look at this sometime soon?

billetd:
@mts299 not until next week at the earliest I'm afraid

billetd:
I'm still not sure how to deal with the rmath.h issue, and how !CONST.R_Earth will be different.

mts299:
@ecbland any thoughts on this?

ecbland:
> @ecbland any thoughts on this?

Not at the moment -- I'm overloaded with other stuff right now.

mts299:
@ecbland I believe @billetd has been busy lately with some work at SuperDARN Canada and ongoing work with pyDARN. Wondering if someone else could be assigned to this? 

Seems like a simple PR to get done

ecbland:
This wasn't as straightforward as it looked because IDL's value of the Earth radius (`!CONST.R_earth=6378136.6`) is different to the value used in RST (which @egthomas pointed out above). 

@billetd Would it be best if we close this PR for now? We can keep the issue (#282) open and/or capture it as a TODO on the projects page.

billetd:
The only solution I can think of from looking at this again is perhaps to make a `constants` common block, but I'm not sure if that's good practice. Fine for this to be on hold if a suitable fix can't be found.

ecbland:
Closing this PR because we have higher priorities. A big thank you to @billetd for looking into this! :100:

Issue #393, opened 2021-03-01
[BUG] grid_plot  -x doesn't stay open


mts299:
# BUG 

`grid_plot -x` does not keep the display window open when displaying the data. 

## Priority

- [x] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

`grid_plot -x -raw -st 00:02 -et 00:08 -time -coast -xkey rainbow.key -fcoast 20100305.pgr.grid`

While trying to test pyDARN `plot_grid` method, I was using 'grid_plot' as verification and wanted to use `-x` option to just display instead of making an actual plot file. However, it opens then closes the window even when it hits the end time. 

### Attempts

I just tried various time options to see if it would get fixed. 

### Data Location

I am only testing with a single radar grid file so just need to make `make_grid` on a fitacf file. 

## Potential Bug Location 

In the grid_plot.c code where I think `-x` is getting closed to prevent multiple windows open. I assume the best solution is the last timestamp should not close the window. So determining when that is and then preventing the window to close. 

## Potential Solution(s)

A nice elegant solution would actually give the user control to close each window or move to the next instead of it closing them as the time stamp moves on.

mts299:
@ecbland just told me `-delay` is what I need to use to fix the problem. 

Would it be good to maybe have -delay 0 as the default so that this `-x` option doesn't seem pointless without `-delay 0`?

ecbland:
Yes, adding `-delay 0` lets the user click through the plots at their leisure. I don't feel strongly about what the default behaviour should be. Any other opinions here?

Note that `map_plot` is similar to `grid_plot`, so both should behave the same.

PR #397, opened 2021-03-04
[FIX] grid_plot set delay 0 

mts299:
# Scope

Changed delay = 0.1 to delay = 0 to address issue #393 
The new changesets the default to allow the user to click through images when using the `-x` with `grid_plot`

## Test
`grid_plot -x -raw -mag -time -coast -fcoast 20150301.2001.00.inv.grid`

Should be able to click through the images


ecbland:
@mts299 I know this is scope creep in relation to the original issue (#393), but I think the default value of `delay` should be the same for all routines that have it:

- `iqplot`
- `field_plot`
- `grid_plot`
- `map_plot`

This PR changes the default behaviour, so would it make sense to print some basic instructions in the command line? 
e.g. `Click to view next plot. Ctrl+c to exit. Use -delay option to cycle through all frames sequentially.`

mts299:
Update:
- [x] update the other plotting methods @ecbland mentioned 
- [x] add user comment "Click to view next plot. Ctrl+c to exit. Use -delay option to cycle through all frames sequentially."
- [x] add the default in XML files as well

ecbland:
Ok, there was some discussion off-github a while ago about whether to make this particular change. Thanks to everyone who has offered their view. This is a time where I think an executive decision is needed so that we can move forward and focus our energy on other things. Let's implement the change, and if users don't like it we can change it back in a later release. 

This decision is based on the following: 

- The current default behaviour is confusing when the user wants to view the data from a single time interval rather than a sequence of plots. Although the solution (set `-delay 0`) is in the documentation, it's still not particularly obvious to the general RST user. 
- The current default on `develop` (`-delay 0.1`) is quite fast, so most users will probably want to change it to something else anyway
- This PR modifies the default behaviour of the code, but all functionality is retained. Users can view a movie by setting `-delay` to any non-zero value

@mts299 The code works as described. There are some conflicts relating to the license text, so I will resolve those now and then merge. Thanks for your patience with this PR :+1:

pasha-ponomarenko:
@ecbland, I am still learning how to use GitHub and just curious if "`Merge branch 'develop' into fix/grid_plot_display`" is the right way to go...

ecbland:
> @pasha-ponomarenko I am still learning how to use GitHub and just curious if "Merge branch 'develop' into fix/grid_plot_display" is the right way to go

Yes, this is correct. This is the commit message that Github uses when resolving conflicts. Nothing to worry about.

ecbland:
Sorry for all the extra commits -- the license notices were outdated and I should have merged them from `develop` first. Anyway, it's fixed now so I will merge this in.

PR #398, opened 2021-03-10
Update instructions for map_addimf in readthedocs

ecbland:
This PR adds some basic instructions to readthedocs on how to add IMF data from CDF files in `map_addimf`

(I actually started this 1 year ago but never made the PR. Better late than never!)

### DIFF
--- a/docs/user_guide/map_grid.md
+++ b/docs/user_guide/map_grid.md
@@ -1,6 +1,8 @@
 <!--
-(C) copyright 2019 University Centre in Svalbard (UNIS)
+(C) Copyright 2019 University Centre in Svalbard (UNIS)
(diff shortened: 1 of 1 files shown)


ecbland:
Thanks @mts299! 

If you want to follow the instructions in the documentation, please use RST4.4.1 due to the bug I just reported in #399

mts299:
Got it to work however there are some underlying issues with the user interface I think... 

I will make an issue on it that may change this documentation. 

Do you want me to merge or hold off?

ecbland:
@mts299 Oops, I forgot to reply to this. I think it's fine to merge this, and then we can update the documentation again if the functionality changes later

Issue #402, opened 2021-04-08
time_plot does not load custom colour keys


ecbland:
# BUG 

In the RST plotting routines, it is possible to choose different colour keys for power/velocity/width/etc using command line options. This is not currently working for `time_plot`, where only the default colour tables work. 

## Priority

- [X] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

When I use `-pkey  [name.key]` to specify a different colour key for the power, the option gets ignored and there is no error message. This also applies to `-vkey`, `-wkey` etc.
```
time_plot -x -p -b 8 -pkey red.key 20140213.1201.00.han.fitacf 
```

The first figure below is from `develop` (incorrect behaviour), and the second is from `v4.4.1` (correct behaviour)

<img src="https://user-images.githubusercontent.com/22493623/114058570-e1275d80-9893-11eb-88dd-d4a2be898a44.png" width="350"><img src="https://user-images.githubusercontent.com/22493623/114058573-e2588a80-9893-11eb-86ed-0abd39095b11.png" width="350">

### Attempts

*What have you tried already to fix it?*
Nothing

*What have you tried to get around it?*
Use RST4.4.1

### Data Location

Any fitacf file

## Extra Notes

I also tried `field_plot` with a custom colour table and it works, so I think this bug is unique to `time_plot`

PR #403, opened 2021-04-09
Fixing bug in time_plot where user-defined color keys did not work

egthomas:
Fixed the bug in `time_plot` where user-defined color keys did not work as identifed by @ecbland in #402.  This was due to a difference in how `load_key` functioned for `time_plot` compared to the other plotting binaries, which I did not notice when consolidating everything into `binplotlib`.


ecbland:
Thanks @egthomas! :100: 
That fixes the bug, and the other plotting routines still work as before. Merging it in...

PR #405, opened 2021-04-13
Update license/copyright FAQ

ecbland:
This PR is a minor update to the copyright & licensing information that was added to the developer's guide a few weeks ago. Specifically, I have updated the FAQ section in response to questions/feedback from the PIs, plus some other tiny changes. 

Once this is merged, I will share the updated info with the PIs (hence the "high priority" label)

### DIFF
--- a/docs/developers_guide/copyright_license.md
+++ b/docs/developers_guide/copyright_license.md
@@ -144,6 +144,10 @@ Adding copyright notices is appropriate only for __substantial__ modifications. 
 
 ## Frequently asked questions
 
+__Is it possible to restrict RST to non-commercial use only?__<br/>
+No. The license prohibits us from placing any additional restrictions on the use of the software. 
+
+
 __Can RST be released with a different license?__<br/>
 To release RST with a different license, we would need permission from all of the copyright holders. This means that, in practice, it would be very difficult to re-license the software. This is the intention of the GPL--it ensures that the software can always be used freely. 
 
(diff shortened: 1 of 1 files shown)


docs/developers_guide/copyright_license.md
```diff
@@ -144,6 +144,10 @@ Adding copyright notices is appropriate only for __substantial__ modifications.
 
 ## Frequently asked questions
 
+__Is it possible to restrict RST to non-commercial use only?__<br/>
+No. The license prohibits us from placing any additional restrictions on the use of the software.
```
ksterne, line 148: Is it fair to add a caveat here that this could be added if we were not using open source standard GPL or LGPL licenses?  If we wanted to add in terminology about non-commercial use, we would be creating our own license which requires expertise in software licensing in order to properly create a non-standard license.  Does that seem correct?  I'm not advocating for it either as I believe creating a specialized license could cause problems with collaboration, maybe even amongst SuperDARN groups.

ecbland: @ksterne I've added some more detail below...does this capture what you were thinking?

```suggestion
No. The license prohibits us from placing any additional restrictions on the use of the software. This is the key principle of free software -- "free" refers to "freedom" (not price), so the software must be available to anyone to use for any purpose. Adding a "non-commercial use only" restriction would require us to release RST under a different license (possibly a non-standard license), which would be very difficult in practice (see next question). Also, any change to the license would apply only to new releases; we cannot prevent commercial use of the previous versions of RST that are already licensed under GPL. 
```

ksterne: Yes, I think that helps explains things better.  It's good to point out the use of "free" from the license perspective.

ecbland:
Could someone please merge this if it's ready to go? I'd like to send it back to the PIs, and it's simplest to keep pointing them to the version that's on `develop` :smile:

Issue #406, opened 2021-04-19
Core dump error in hmb_plot and pc_plot


ecbland:
# BUG 

`hmb_plot` and `pc_plot` produce a core dump error.

## Priority

- [X] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

```
## X-terminal output: the plot displays correctly, but there's an error when closing it
$ hmb_plot -x 20130702.n.map
free(): invalid pointer
Aborted (core dumped)

pc_plot -x 20130702.n.map 
free(): invalid pointer
Aborted (core dumped)


## PNG output
$ hmb_plot -png 20130702.n.map > out.png
munmap_chunk(): invalid pointer
Aborted (core dumped)
```

### Attempts

*What have you tried already to fix it?* 

Commenting out this line fixes the problem
https://github.com/SuperDARN/rst/blob/d38edbf6c59beb28ce476b3d75d29730c8c731f7/codebase/superdarn/src.bin/tk/plot/hmb_plot.1.6/hmb_plot.c#L528

However, that line was added by @egthomas to fix a memory leak (https://github.com/SuperDARN/rst/commit/d38edbf6c59beb28ce476b3d75d29730c8c731f7), so removing it may not be appropriate.

EDIT: `pc_plot` has the same bug. 

### Data Location

Any map file

PR #408, opened 2021-04-19
Fixing pointer freeing error in hmb_plot and pc_plot

egthomas:
This pull request fixes the error in `hmb_plot` and `pc_plot` identified by @ecbland in #406.  The problem was that this freeing of `fname` should only occur when command line options are provided in a separate file with the `-cf` option.


ecbland:
> The problem was that this freeing of fname should only occur when command line options are provided in a separate file with the -cf option.

Ahh, that makes sense. Both routines now work correctly (with and without the `-cf` option), so this PR fixes the bug. Thanks @egthomas :100:

Issue #407, opened 2021-04-19
Postscript output not working for some plotting routines


ecbland:
# BUG 

When writing a postscript file from some plotting routines, the output file is blank or garbled. The following routines are affected:
- `time_plot`
- `fov_plot`
- `istp_plot`
- `hmb_plot` (but see also #406)
- `pc_plot` (but see also #406)

These routines work with postscript:
- `field_plot`
- `grid_plot`
- `map_plot`

## Priority

- [X] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

```
time_plot -ps -a -b 5 20100305.1401.00.rkn.fitacf > out.ps
fov_plot -ps -st lyr -coast -stereo -ffov > out.ps
```
![Screenshot from 2021-04-19 11-14-59](https://user-images.githubusercontent.com/22493623/115212300-a5488f80-a100-11eb-8edb-66ee0ba2427a.png)

### Attempts

*What have you tried already to fix it?* *What have you tried to get around it?*

I haven't tried to fix this. But I did confirm that these routines work with v4.4.1. So this bug was introduced in v4.5.

### Data Location

`fov_plot` doesn't need an input file. The others need either a fitacf file or map file. 

## Extra Notes

```
gcc --version
gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
```

PR #409, opened 2021-04-19
Fixing postscript output for various plotting binaries

egthomas:
This pull request fixes the postscript plotting issue identified by @ecbland in #407.


ecbland:
Tested this for `time_plot`, `fov_plot`, `istp_plot`, `hmb_plot` and `pc_plot`. The ps output now works for all of them. Thanks @egthomas !!

PR #411, opened 2021-04-20
Fix elevation angle error bounds in FITACF2.5

### LINKED ISSUE #410: ELV_HIGH lower than ELV in many FitACF v2.5 cases
# BUG 

The "ELV_HIGH" field in fitACF v2.5 files frequently contains values lower than the corresponding "ELV" values. 
This is contrary to docs, which indicate ELV_HIGH contains the highest estimate of elevation. 

## Priority

- [ ] Major - misrepresentation of data, failure in a common situation 

## Example of the bug 

See, e.g. 20140423.cvw.fit, 20140423.kod.d.fit (created using fitACF from all BAS rawACFs on 23 April). 

In case of 20140423.cvw.fit (read using Pydarn):

elv
['32.40', '12.04', '12.94', '18.51', '35.22', '10.87', '12.63', '34.65', '20.58', '21.48', '17.18', '22.70', '20.33']
elv_high
['32.08', '9.59', '9.45', '15.82', '34.88', '6.50', '11.06', '34.09', '19.05', '19.80', '14.95', '19.93', '17.54']

## Potential Bug Location 

cross-correlation code

## Potential Solution(s)

fix the elevation angle algorithm

ecbland:
This PR fixes #410 reported by @alexchartier, where the elevation angle error bounds (`elv_low` and `elv_high`) are flipped for some radars. The problem occurs for radars with the interferometer array behind the main array. 

I fixed this problem for the default (Shepherd) elevation angle calculation, and also for the `-old_elev` option. For consistency I corrected the function call to `elev_goose()`, but that radar has a front inteferometer so it doesn't change the behaviour. 

To test (with and without `-old_elev`):
```
make_fit [-old_elev] 20100305.1401.00.inv.rawacf > inv.fitacf      # front inteferometer 
make_fit [-old_elev] 20100305.1401.00.rkn.rawacf > rkn.fitacf      # rear interferometer

# Display the phase error, elevation, elv_low and elv_high (see note below)
dmapdump -d rkn.fitacf | grep -A 7 phi0_e | head -8
```

- Front interferometer case: the output files should be the same on this branch and `develop`, with `elv_high` > `elv_low` as expected. 
- Rear interferometer case: `elv_high` > `elv_low` on this branch, and `elv_high` < `elv_low` on `develop`

**Note:** This PR does not account for phase flipping due to the 2pi ambiguity. If a 2pi jump occurs when calculating the elevation angle error bounds, it's possible to get `elv_high` < `elv_low`. You can see this problem quite well for the `inv` example I provided above. I don't want to change this behaviour because it would be misrepresenting the data.

### DIFF
--- a/codebase/superdarn/src.lib/tk/fitacf.2.5/src/do_fit.c
+++ b/codebase/superdarn/src.lib/tk/fitacf.2.5/src/do_fit.c
@@ -368,30 +368,30 @@ int do_fit(struct FitBlock *iptr, int lag_lim, int goose,
           xptr[k].phi0 = xptr[k].phi0*iptr->prm.phidiff;
 
 
-      /* phi_sign indicates whether interferometer array is in front (+) or behind (-) main array
+      /* Y_offset_sign indicates whether interferometer array is in front (+) or behind (-) main array
          used for elv_low and elv_high calculation */
-      int phi_sign;
+      int interfer_sign;
       if (iptr->prm.interfer[1] > 0.0)
-        phi_sign= 1.0;
+        Y_offset_sign= 1.0;
       else
-        phi_sign= -1.0;
+        Y_offset_sign= -1.0;
 
       if (iptr->prm.old_elev) {
         /* use old elevation angle routines */
         if (goose == 0) {
           elv[k].normal = elevation(&iptr->prm, xptr[k].phi0);
-          elv[k].low = elevation(&iptr->prm, xptr[k].phi0 + phi_sign*xptr[k].phi0_err);
-          elv[k].high = elevation(&iptr->prm, xptr[k].phi0 - phi_sign*xptr[k].phi0_err);
+          elv[k].low = elevation(&iptr->prm, xptr[k].phi0 + Y_offset_sign*xptr[k].phi0_err);
+          elv[k].high = elevation(&iptr->prm, xptr[k].phi0 - Y_offset_sign*xptr[k].phi0_err);
         } else {
           elv[k].normal = elev_goose(&iptr->prm, xptr[k].phi0);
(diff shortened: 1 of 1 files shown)


pasha-ponomarenko:
@ecbland, this two-pi flipping-over was one of the reasons why in FITACF3 we suggested using a single elevation error value instead of low/high boundaries.

ksterne:
Tested this out with some data from cvw and from kap and appears to work.  There's some confusion for the kap data, but I think that is from the 2 pi ambiguity.  My only comment is (and I'll admit it's a little nit-picky): coming from the hdw.dat perspective, is there a better variable name to use than phi_sign?  There's a hdw.dat parameter that's called phase sign that relates to elevation angle calculation and though it's fairly clear on the definition in the code, I didn't want to introduce a point of confusion.  Maybe I'm being overly cautious or worrying, but thought I would share in case others had it.  I do think it's fine as is and coming from the angle of arrival calculation, understand the choice of variable name.

ecbland:
@ksterne Sure, I can change the variable name. I copied `phi_sign` from `elevation.c`, but I see that it was replaced with `sgn` in `elevation_v2.c`:

https://github.com/SuperDARN/rst/blob/b667628c85580b9bbcff6b167345a33a5d0c1c56/codebase/superdarn/src.lib/tk/elevation.1.0/src/elevation.c#L51-L52

https://github.com/SuperDARN/rst/blob/b667628c85580b9bbcff6b167345a33a5d0c1c56/codebase/superdarn/src.lib/tk/elevation.1.0/src/elevation_v2.c#L31

ksterne:
Sounds good @ecbland!  That was a good thought to use the same variable to do the same function or for the same purpose.

PR #416, opened 2021-05-04
Fix undeclared variable in FITACF2.5

ecbland:
I just realised that compilation on `develop` fails now that #411 has been merged. This bug was introduced when I changed a variable name but forgot update the variable declaration.

To test, just run `make.build && make.code` on this branch and check that the package compiles correctly


mts299 (approved):
Compiles fine on Open Suse

Issue #413, opened 2021-04-21
Historical version log in README.md


ecbland:
# Question

Does the "Historical Version Log" in [`readme.md`](https://github.com/SuperDARN/rst/blob/master/README.md) serve any useful purpose nowadays? I'm wondering if it can be removed...

mts299:
Totally agree! actually, the RST wiki would be a good place for that. Given we don't actually add the more recent history I really see not much point by now.

PR #417, opened 2021-05-06
Remove historical version log from README.md

ecbland:
This PR removes the historical version log from `readme.md` (see #413)

I moved this information to the wiki so that it will be retained somewhere: https://github.com/SuperDARN/rst/wiki


mts299 (approved):
Looks way better!

Issue #419, opened 2021-05-13
Very high SNR value sporadically occurring across all range gates in BOREALIS data processed with FITACF3


pasha-ponomarenko:
# BUG 

[Sporadic srreaks of high SNR in BOREALIS data processed with FITACF3 .pdf](https://github.com/SuperDARN/rst/files/6474905/Sporadic.srreaks.of.high.SNR.in.BOREALIS.data.processed.with.FITACF3.pdf)

## Priority

- [X] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

 **Please see the attached PDF document** 

### Data Location

*e.g., 20210203.1600.00.cly.a.rawacf (beam 5)*

##  Bug Location 

Plese see the attached PDF

## Solution

Commenting out line 827 in rst/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c :

*if (min_pwr < 1.0) min_pwr = fit_prms->noise;* 

It has been tested and seems to work (see the attached PDF).

PR #420, opened 2021-05-28
Revising the noise calculation in FITACF3.0

pasha-ponomarenko:
This pull request deals with the problem described in issue [#419](https://github.com/SuperDARN/rst/issues/419) when very high (Inf) SNR values sporadically occurr across all range gates in BOREALIS data processed with FITACF3 (note that those are absent if you process the same data with FITACF2.5). This effect is regularly observed in CLY data:

20210203 1600 00 cly a v3

Testing of the bug fix is very simple: run _make_fit_ from this branch on the respective _rawacf_ file and plot it:

          make_fit -fitacf-version 3.0 20210203.1600.00.cly.a.rawacf > 20210203.1600.00.cly.a.v3_m.fitacf
              
          time_plot -png -km -frang 0 -erang 3500 -a -phi -vmax 100 -vmin -100 -b 5 -st 16:00 -ex 02:00 20210203.1600.00.cly.a.v3_m.fitacf > 20210203.1600.00.cly.a.v3_m.png 

The vertical stripes should disappear: 

20210203 1600 00 cly a v3_m

### DIFF
--- a/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
+++ b/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
@@ -823,7 +823,7 @@ double ACF_cutoff_pwr(FITPRMS *fit_prms){
     ni = (ni > 0) ? ni :  1;
     cpc = cutoff_power_correction(fit_prms);
     min_pwr = min_pwr/ni * cpc;
-    if (min_pwr < 1.0) min_pwr = fit_prms->noise;
+    /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
 
     free(pwr_levels);
     return min_pwr;
--- a/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
+++ b/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
@@ -21,9 +21,8 @@ Author(s): Keith Kotyk July 2015
 Modifications: 
     2020-03-11 Marina Schmidt (SuperDARN Canada) removed all defined 
                               constants and included rmath.h 
-    2021-05-26 Pasha Ponomarenko (SuperDARN Canada) commented out repacement of the
-                                skynoise by search noise as it caused problems with 
-                                processing BORELAIS data (no search noise)
+    2021-05-26 Pasha Ponomarenko (SuperDARN Canada) check that the search noise
+               is nonzero before using it to replace the skynoise when min_pwr < 1
 
 */
 
@@ -828,13 +827,12 @@ double ACF_cutoff_pwr(FITPRMS *fit_prms){
     min_pwr = min_pwr/ni * cpc;
     
     
-    /*  Commenting the line below out as it causes problems with processing
-            BOREALIS data by setting Inf values to SNR as the search noise is not 
-            estimated. Amending it temporarily with the check
-            for "skynoise" data to be nonzero. This is a quick fix that allows 
-            BOREALIS data to be processed properly but also makes no impact 
-            on the older data. A more elaborate/universal solution to this issue 
-            should be developed for th enext release. */
+    /*  The commented line below causes Inf values of the SNR when the search noise 
+        is zero. Therefore, we introduce an additional condition that the search noise 
+        must be nonzero. This is a temporary fix that allows the affected data to be 
+        processed properly while not affecting the processing of any other data. 
+        A more universal solution to this issue should be developed for the 
+        next release. */
    
     /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
     if ((min_pwr < 1.0) && (fit_prms->noise != 0.0)) min_pwr = fit_prms->noise;
@@ -1492,5 +1490,4 @@ void Fill_Data_Lists_For_Range(llist_node range,llist lags,FITPRMS *fit_prms){
 
 
 }
-
 
--- a/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
+++ b/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
@@ -21,6 +21,9 @@ Author(s): Keith Kotyk July 2015
 Modifications: 
     2020-03-11 Marina Schmidt (SuperDARN Canada) removed all defined 
                               constants and included rmath.h 
+    2021-05-26 Pasha Ponomarenko (SuperDARN Canada) commented out repacement of the
+                                skynoise by search noise as it caused problems with 
+                                processing BORELAIS data (no search noise)
 
 */
 
@@ -823,6 +826,12 @@ double ACF_cutoff_pwr(FITPRMS *fit_prms){
     ni = (ni > 0) ? ni :  1;
     cpc = cutoff_power_correction(fit_prms);
     min_pwr = min_pwr/ni * cpc;
+    
+    
+    /*  Commenting the line below out as it causes problems with processing
+            BOREALIS data by setting Inf values to SNR as the search noise is not 
+            estimated. */
+   
     /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
 
     free(pwr_levels);
--- a/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
+++ b/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
@@ -830,10 +830,15 @@ double ACF_cutoff_pwr(FITPRMS *fit_prms){
     
     /*  Commenting the line below out as it causes problems with processing
             BOREALIS data by setting Inf values to SNR as the search noise is not 
-            estimated. */
+            estimated. Amending it temporarily with the check
+            for "skynoise" data to be nonzero. This is a quick fix that allows 
+            BOREALIS data to be processed properly but also makes no impact 
+            on the older data. A more elaborate/universal solution to this issue 
+            should be developed for th enext release. */
    
     /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
-
+    if ((min_pwr < 1.0) && (fit_prms->noise != 0.0)) min_pwr = fit_prms->noise;
+    
     free(pwr_levels);
     return min_pwr;
 


codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
```diff
@@ -823,7 +823,7 @@ double ACF_cutoff_pwr(FITPRMS *fit_prms){
     ni = (ni > 0) ? ni :  1;
     cpc = cutoff_power_correction(fit_prms);
     min_pwr = min_pwr/ni * cpc;
-    if (min_pwr < 1.0) min_pwr = fit_prms->noise;
+    /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
```
ecbland, line 826: Do you want to add a brief comment here to explain why this line was commented out?

ecbland (changes requested):
@pasha-ponomarenko On slide 7 of your notes in #419 you said the following:

> A sanity check flagging zero skynoise values can be added here as well.

This sounds like a good idea. Do you want to add it?

---------
Also, could you please add a new entry under "modifications" at the top of the file?

https://github.com/SuperDARN/rst/blob/866f3b64b010210c898ec667c87705f556d9c9dd/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c#L21-L23

codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
pasha-ponomarenko: Thanks a lot for leading me through the correct procedures, @ecbland ! 
:-) 

On the sanity check, I am not sure what should be the output of this procedure if the background noise value is zero for some reason (probably, only possible for older data).

On a more general note, the original replacement of "sky" noise < 1  with the "search" noise seems to be a bit strange. The latter is estimated through a completely different procedure involving FFT which is performed over a short time interval and, I suspect, at a higher sampling rate. I have had a look at the code used by ROS, and there are virtually no comments throughout the text and some interesting modification remarks like this:

_Revision 1.7  2004/05/14 16:27:31  barnes
Divided noise value by three to correct for signal abnormalities._

I assume that some sort of _ad hoc_ scaling has been applied to the "search" noise level in order to match that of the "sky" noise. From what I have seen, this scaling is not working for at least some datasets. On the other hand, this replacement might still be required for processing old datasets... Looks like we might need to open another issue to dig deeper into this.

pasha-ponomarenko: Further on the noise value replacemet, I feel that we may need to re-think the original way of determining the "sky" noise that arbitrarily utilises only one third of the available range gates:

 /*  look for the lowest 10 values of lag0 power and average to
            get the noise level.  Ignore values that are exactly 0.  If
            you can't find 10 usable values within the first 1/3 of the
            sorted power list, then just use whatever you got in that
            first 1/3.  If you didn't get any usable values, then use
            the NOISE parameter */

    while ((ni < 10) && (i < fit_prms->nrang/3)) {
        if (pwr_levels[i]) ++ni;
        min_pwr += pwr_levels[i++];
    }

ecbland: > I assume that some sort of ad hoc scaling has been applied to the "search" noise level in order to match that of the "sky" noise.

Oh dear :rofl: :rofl:

> we may need to re-think the original way of determining the "sky" noise that arbitrarily utilises only one third of the available range gates

I've also been wondering about this. It sounds like another ad-hoc legacy thing, perhaps from the days when not all pwr0 values were stored??  We need to be careful that it doesn't break anything for the really old data.

pasha-ponomarenko: Yep, there was another layer of pre-selection by the ROS when the data with lag 0 power below a certain A/D converison level were not stored at all. I tried to find how the rejection threshold was calculated but did not succeed as there were no documentation on this. 

Zero values in raw data have been discovered by Keith when he was testing FITACF3 and obtaining infite spectrl width for some older data.

I also recall that ANge,ine might have had a closer look at this issue as well but don't remeber the details....

pasha-ponomarenko:
I just want to outline the problem background in a bit more detail so that we can find a suitable solution.  
We have a significant portion of a historic dataset from which ACFs with lag 0 power below a certain (still unknown) threshold were dropped. 
Currently, the data used for the mean noise calculation are extracted from the bottom third of the sorted lag 0 power values.  If the average of non-zero values in this subset is below 1(???), the search noise level is used instead. 
I can only guess about the logic behind this arbitrary selection procedure. It looks like the code developers anticipated that if too many ACFs are excluded form analysis due to their low power, then the noise estimate becomes unreliable, and they designed an alternative way to calculate it based on the “search” noise. According to the comments in the modification history, this procedure required some power level adjustment for the “search” noise to become compatible with the “sky” noise.
Based on the above, we have some issues to deal with: 
(1) The removal of the low-power ACFGs leads to overestimating the background noise level. 
I am not sure if there is a proper way to compensate for it as an unknown amount of information about the low-power "tail" of the noise distribution is irreversibly lost.
(2) The use of the “clear frequency search” data as a substitute is rather questionable due to the way this parameter is evaluated. As follows form observations it might be just incompatible with the "normal" noise estimates.   
(3) It is hard to understand why the replacement would be triggered at a non-zero mean noise value of one (1) rather than zero (0). 
With this respect I have a strategic question: what is the reproducibility policy with respect to different versions of FITACF?

I would prefer to get rid of yet another ad hoc procedure and we select the “weakest” lag 0 powers from the whole set of non-zero values and abandon using the “search noise” as a substitute. 
It will lead to overestimating the noise level/underestimating SNR in the old data. 
I am pretty sure that this effect should be minimal but to be on the safe side, we would just need to look at a couple of examples with "dropped" ACFs.

ecbland:
> @pasha-ponomarenko With this respect I have a strategic question: what is the reproducibility policy with respect to different versions of FITACF?

My view is that we should avoid making modifications to FITACF2.5 that would have any impact on the data pre-selection or the fitted parameters. I think small "self-contained" changes are ok if there is a good reason to make them (e.g. #411), but nothing more significant than that. So I'd recommend that we review the noise calculation in FITACF3.0 only. We can rename it to FITACF3.1 if that seems appropriate. 

Two general notes about reproducibility:
- Changes to the analysis software that may affect the scientific interpretation of the data should be discussed with the PIs. Changing the noise calculation _may_ fall into this category (depending on what effect it has on the data)
- Each RST release has a DOI, so authors can specify which version was used to process their data so that their work is reproducible. 

> @pasha-ponomarenko I would prefer to get rid of yet another ad hoc procedure: to select the “weakest” lag 0 powers from the whole set of non-zero values and to abandon using the “search noise” as a substitute.

Completely agree...the search noise is not a good substitute for the "sky noise" -- If anyone is interested, I have some fun examples buried in my notes :stuck_out_tongue_winking_eye:

ecbland:
Does anyone know when the radars stopped discarding ACFs below the threshold?

pasha-ponomarenko:
Adding  for reference the clear frequency search code. I have got it from Dieter few years ago:
[fclrgc214.txt](https://github.com/SuperDARN/rst/files/6656366/fclrgc214.txt)
Do we have a repository for this sort of documents?

ecbland:
> @pasha-ponomarenko  Adding for reference the clear frequency search code...Do we have a repository for this sort of documents?

Yes, there is a separate ROS archive on Github: https://github.com/SuperDARN/ros-archive

https://github.com/SuperDARN/ros-archive/blob/ros.1.25/qnx/code/src.lib/radarqnx4/fclrgc214.1.15/src/fclrgc214.c

pasha-ponomarenko:
@ecbland, I made a quick fix  (now in lane 840 of preporcessing.c) that performs the skynoise replacement by the search noise only if the search noise value is non-zero. In this way we preserve the old analysis output but resolve the "streaks" issue for BOREALIS data (and any other data that have zero search noise values). 
This is a temporary measure, and this issue should be resolved in a more fundamental fashion in the next release.
Please check my syntax :-)

ecbland:
Thanks @pasha-ponomarenko! I'll take a look ASAP. There are indeed some other radars that haven't used the clear frequency search at one point or another, so I will test with them as well.

codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
```diff
@@ -823,8 +826,19 @@ double ACF_cutoff_pwr(FITPRMS *fit_prms){
...
+            BOREALIS data by setting Inf values to SNR as the search noise is not 
+            estimated. Amending it temporarily with the check
+            for "skynoise" data to be nonzero. This is a quick fix that allows 
+            BOREALIS data to be processed properly but also makes no impact 
+            on the older data. A more elaborate/universal solution to this issue 
+            should be developed for th enext release. */
+   
+    /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
+    if ((min_pwr < 1.0) && (fit_prms->noise != 0.0)) min_pwr = fit_prms->noise;
+
```
ecbland, line 841: Just to avoid implying that this problem can only occur with the BOREALIS radars, would you mind if we rephrase this slightly?
Also, I think you meant to write "Amending it temporarily with the check for **search noise** data to be nonzero"

```suggestion
    /*  The commented line below causes Inf values of the SNR when the search noise 
        is zero. Therefore, we introduce an additional condition that the search noise 
        must be nonzero. This is a temporary fix that allows the affected data to be 
        processed properly while not affecting the processing of any other data. 
        A more universal solution to this issue should be developed for the 
        next release. */
   
    /*if (min_pwr < 1.0) min_pwr = fit_prms->noise;*/
    if ((min_pwr < 1.0) && (fit_prms->noise != 0.0)) min_pwr = fit_prms->noise;
    
```

codebase/superdarn/src.lib/tk/fitacf_v3.0/src/preprocessing.c
```diff
@@ -21,6 +21,9 @@ Author(s): Keith Kotyk July 2015
 Modifications: 
     2020-03-11 Marina Schmidt (SuperDARN Canada) removed all defined 
                               constants and included rmath.h 
+    2021-05-26 Pasha Ponomarenko (SuperDARN Canada) commented out repacement of the
+                                skynoise by search noise as it caused problems with 
+                                processing BORELAIS data (no search noise)
```
ecbland, line 26: Is this a better description of the change?
```suggestion
    2021-05-26 Pasha Ponomarenko (SuperDARN Canada) check that the search noise
               is nonzero before using it to replace the skynoise when min_pwr < 1
```

ecbland (commented):
I tested this for a few radars. I didn't find any non-BOREALIS examples where this change would matter, but it is possible there are examples out there in the historical data. In any case, this PR fixes the divide by zero problem.

pasha-ponomarenko:
@ecbland, yes these modificaitons do make sense! Please go ahead and implement those changes!

ecbland (approved):

ecbland:
I'm adding another example here to answer a specific question from a PI about the Clyde River data on 29 May 2021.

The plots below show that this PR solves the issue with the vertical stripes in BOREALIS data.

1. fitacf v3.0 from RST4.5
2. fitacf v3.0 from RST4.6 (which includes the fix)
3. fitacf v2.5 provided for comparison

20210529 cly fitacf3_RST4 5
20210529 cly fitacf3_RST4 6
20210529 cly fitacf2

Issue #345, opened 2020-10-05
Bad phi0 calculation at DCE


aburrell:
I have been looking into the DCE elevation angles recently, and found that a bug had been introduced into fitacf version 3.0 in the last release.  I didn't see any issues for this, but it's possible someone else has seen it too.  The problem is, all phi0 values are zero and they aren't if you run fitacf 2.5.  They also were not zero in the previous release of RST.  You can see where my processing was when the last release came out:

![dce_phi0_status_2018](https://user-images.githubusercontent.com/7045886/95096161-ec89a880-06f9-11eb-9a8b-271984a23f2f.png)

You can see the difference between running version 2.5 and 3.0 on the same rawacf file:

![dce_phi0_status_fitacf_version](https://user-images.githubusercontent.com/7045886/95096224-00350f00-06fa-11eb-9819-bc4fd6348665.png)

But this is not the case for all radars:

![han_phi0_status_fitacf_version3](https://user-images.githubusercontent.com/7045886/95097304-53f42800-06fb-11eb-9926-ac836affca9b.png)

egthomas:
I noticed this behavior (phi0 = 0) for some recent BPK data processed using FITACF 3.0 as well, but I wasn't sure if it was a radar-specific issue or related to the FITACF 3 algorithm.

ecbland:
@aburrell Could you please clarify which two versions of RST you're comparing here?

aburrell:
1) The current master branch (v4.4.1) and the one right before it (v4.4) (I think!).  
2) Yes, it does apply to all DCE data (I went back and checked).  The date change just indicates that I updated RST at that point in the processing.

egthomas:
I thought the only change between v4.4 and v4.4.1 was to the stdkey makefile, which shouldn't affect any of the fitting algorithms, right?

mts299:
yea... I just changed a make thing which essentially going back

aburrell:
Maybe I had the version before that then... Sorry, I can see what I have now, but I am not sure how to see what I had before!

egthomas:
This is getting a little off-topic, but maybe we should think about appending the RST version number to the `origin.command` string since the FITACF major/minor revision numbers are not always sufficient to capture changes between releases?

mts299:
@egthomas can you make an issue for this comment so we can discuss it there and focus on the issue about the change in elevation here. I don't want to burry this comment. Gonna do some testing myself soon on this bug.

aburrell:
[...]

PR #421, opened 2021-05-28
Remove unnecessary fprintf statements from FITACF 3.0

ecbland:
This PR removes some `fprintf()` statements from the function `set_xcf_phi0` in `determinations.c`. I think these are leftover from debugging when we added the Shepherd (2017) elevation angle calculations. 

To Test: 
```
make_fit -fitacf-version 3.0 20100305.1428.05.sas.rawacf > /dev/null
```

On develop, you will see the following printed in the terminal:
```
xcf_phi0 before elv_a: -1.267452
xcf_phi0 after elv_a: -1.267452
phidiff: 1.000000
```

On this branch, you will see nothing. 

This PR is not related to #345 but I'm planning to take a look at that soon.


mts299:
Did you want me to take this on since it was my PR that introduced this?

mts299 (approved):

ksterne:
I know this makes this pretty quick, but found this works, the mods are simple and there's possibly an interest in time for a proposal to the PIs.  So merging this in.

PR #423, opened 2021-05-31
Fix comment about FITACF1-2 noise estimation in ACF_cutoff_pwr

ecbland:
@pasha-ponomarenko:  I noticed that the description of the`ACF_cutoff_pwr()` routine in FITACF 3.0 says "overestimated" instead of "underestimated" when referring to the noise level calculation in FITACF1-2.

I also reworded the previous sentence for clarity.


pasha-ponomarenko:
Thaks, @ecbland ! I merge the corrections into develop.

Issue #399, opened 2021-03-10
CDF data not being handled correctly


ecbland:
# BUG 

IMF data from CDF files is not handled correctly in RST4.5.

This bug affects the following routines: `map_addimf`, `istp_plot` and `istp_text` (and others?)

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [X] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

The simplest way to reproduce this bug is
```
istp_text -ace -sd 20200101 -path [path/to/cdf-file]
```
**NB:** the cdf file needs to be placed in a subdirectory of the path given above: `path/to/cdf-file/ace/[year]`

For the above date, the last 3 columns are all zeros. For 20161231 I also got nans and other garbage values.

### Attempts

*What have you tried already to fix it?*
(see next section)

*What have you tried to get around it?*
Use RST4.4.1

### Data Location

CDF files can be downloaded from CDAWeb
e.g. https://cdaweb.gsfc.nasa.gov/pub/data/ace/mag/level_2_cdaweb/mfi_h0/2020/

## Potential Bug Location 

These two lines in `rcdf.c` seem to be the problem
https://github.com/SuperDARN/rst/blob/6697c65e3b2acb4c317168375087b79db4d66335/codebase/analysis/src.lib/cdf/rcdf.1.5/src/rcdf.c#L221
https://github.com/SuperDARN/rst/blob/6697c65e3b2acb4c317168375087b79db4d66335/codebase/analysis/src.lib/cdf/rcdf.1.5/src/rcdf.c#L332

## Potential Solution(s)

Removing `(long*)` from the above lines solves the problem, but then there will be a warning: `pointer of type ‘void *’ used in arithmetic`

## Extra Notes

From the example given above, the correct IMF values for the last few lines are:
```
2020 01 01 23 58 39.000000       2.7400       1.2230      -1.2720
2020 01 01 23 58 55.000000       2.6250       1.2620      -1.3700
2020 01 01 23 59 11.000000       2.2590       1.3540      -1.6290
2020 01 01 23 59 27.000000       2.4790       1.3710      -1.3970
2020 01 01 23 59 43.000000       2.9510       1.2900      -0.9520
```

ecbland:
Closing this with a note in the [cleaning up RST project](https://github.com/SuperDARN/rst/projects/18) to look at this warming again another time.

PR #428, opened 2021-06-03
Fix/cdf not handled

mts299:
@ecbland I followed your suggestion if you want to test it out in #399, I think for now we keep the warning if it means not having code break. 

I can look at this later.


ecbland (approved):
Yes, that works...thanks! I added a note to the [cleaning up RST project](https://github.com/SuperDARN/rst/projects/18) to look at this warming again another time.

Issue #223, opened 2019-02-28
Missing fitted fields in FITACF files 


mts299:
Recently I have been testing field checking in pydarn for reading and writing files, and to my surprise, I kept running into records in fitacf files that are missing the fitted fields:
{'x_sd_l', 'x_p_l_e', 'p_l_e', 'x_v_e', 'phi0', 'x_w_l_e', 'nlag', 'w_s', 'v', 'x_gflg', 'w_l_e', 'qflg', 'p_s', 'x_p_s', 'elv_high', 'x_w_s', 'p_s_e', 'x_qflg', 'x_w_s_e', 'gflg', 'sd_phi', 'w_l', 'sd_s', 'phi0_e', 'x_v', 'elv', 'v_e', 'elv_low', 'x_p_s_e', 'x_w_l', 'x_p_l', 'w_s_e', 'sd_l', 'p_l', 'x_sd_s', 'x_sd_phi', 'slist'}
At first I thought I found a bug but soon realized that this happened in the other fitacf files I produced so now I am thinking these fields are not written when there is "bad data" however the record scalar information is still written in. 

@asreimer told me lmfit2 would not do this because it fits everything. So now I see inconsistencies in the code and file generation. 

I am wondering if we need to review this?  
@egthomas @asreimer @pasha-ponomarenko thoughts or explanations?

egthomas:
Certain fields are only included if the XCF data is present in the rawacf file (eg elevation angles, everything beginning with `x_`, etc).  I'm guessing the other fields may not be written due to bad quality flags.  Here's what an old [tutorial](https://github.com/SuperDARN/rst/wiki/Using-The-IDL-DataMap-Library-With-New-Format-Files) had to say:

> The problem with fitacf files is that depending on whether a good fit was obtained, a range may not be stored in the file. The array variable slist contains a list of the ranges that were stored.

mts299:
@egthomas you are the librarian of RST documentation. 

Moving forward, is this something we might want to change to handle it a bit more elegantly? Otherwise people like myself might think it is a bug or think something is wrong with the data. It can be a bit misleading.

asreimer:
Exactly what @egthomas said. This is only the case for fitacf2.5 though. Here's the specific references in the code that explains why this happens:

[...]

PR #435, opened 2021-06-10
[Ehn] Remove partial records

mts299:
# Scope 

Removing partial records from FITACF files which may remove partial records for GRID and MAP files (still need to double-check but will make a PR for the others if its something else)

# TEST

On `develop`
```bash
make_fit 20180404.0200.12.lyr.rawacf > test.fitacf 
dmapdump test.fitacf | less
```
I search `/ptab`
```
        float   "noise.sky" = 508.299988
        float   "noise.lag0" = 0.000000
        float   "noise.vel" = 0.000000
arrays:
        short   "ptab" [7]
        short   "ltab" [19][2]
        float   "pwr0" [70]
scalars:
        char    "radar.revision.major" = 1
        char    "radar.revision.minor" = 1
        char    "origin.code" = 1
        string  "origin.time" = "Thu Jun 10 23:14:58 2021"
        string  "origin.command" = "make_fit 20180404.0200.12.lyr.rawacf"
        short   "cp" = 153
```
On `ehn/remove_partial_record`

```bash
make_fit 20180404.0200.12.lyr.rawacf > test.fitacf 
dmapdump test.fitacf | less
```
I search `/ptab`
```bash
      short   "ptab" [7]
        short   "ltab" [19][2]
        float   "pwr0" [70]
        short   "slist" [70]
        short   "nlag" [70]
        char    "qflg" [70]
        char    "gflg" [70]
        float   "p_l" [70]
        float   "p_l_e" [70]
        float   "p_s" [70]
```
you will notice there are now variables where they were missing before

### DIFF
--- a/codebase/superdarn/src.lib/tk/fit.1.35/src/fitwrite.c
+++ b/codebase/superdarn/src.lib/tk/fit.1.35/src/fitwrite.c
@@ -18,6 +18,8 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 Modifications:
+2921-06-10 Marina Schmidt (SuperDARN Canada) removed the qflg check that
+causes partial records to be written
 */
 
 #include <stdio.h>
@@ -114,8 +116,9 @@ int FitEncode(struct DataMap *ptr,struct RadarParm *prm, struct FitData *fit) {
 
   snum=0;
   for (c=0;c<prm->nrang;c++) {
-    if ( (fit->rng[c].qflg==1) ||
-         ((fit->xrng !=NULL) && (fit->xrng[c].qflg==1))) snum++;
+    if (((fit->xrng !=NULL)) 
+            snum++;
+            break;
   }
 
   if (prm->xcf !=0) xnum=snum;
--- a/codebase/superdarn/src.lib/tk/fit.1.35/src/fitwrite.c
+++ b/codebase/superdarn/src.lib/tk/fit.1.35/src/fitwrite.c
@@ -119,7 +119,6 @@ int FitEncode(struct DataMap *ptr,struct RadarParm *prm, struct FitData *fit) {
     if (fit->xrng !=NULL)
     {
             snum++;
-            break;
     }
   }
 
--- a/codebase/superdarn/src.lib/tk/fit.1.35/src/fitwrite.c
+++ b/codebase/superdarn/src.lib/tk/fit.1.35/src/fitwrite.c
@@ -116,9 +116,11 @@ int FitEncode(struct DataMap *ptr,struct RadarParm *prm, struct FitData *fit) {
 
   snum=0;
   for (c=0;c<prm->nrang;c++) {
-    if (((fit->xrng !=NULL)) 
+    if (fit->xrng !=NULL)
+    {
             snum++;
             break;
+    }
   }
 
   if (prm->xcf !=0) xnum=snum;


ecbland:
@mts299 This works, although it substantially increases the file size by adding zeros:

Using the same rawacf file: `20180404.0200.12.lyr.rawacf`
- `develop`: output file is 12M
- this branch: output file is 40M

Where do these partial records cause problems?

mts299:
1. Inconsistency in the data, for people unaware of this behavior it looks like our data product is incomplete
2. You are throwing away potentially good data or user data due to the quality flag not equaling = 1 which was the if statement I changed. I don't know what scientific basis this flag is set but I assume the scientist should decide what is good and what is bad
3. pyDARN runs into a bunch of problems with it, so much now users report partial records being a bug in pyDARN when really its not 
4. It has broken Kathryn's IDL code when a partial record is at the beginning and then the code can not set the array size correctly. 

If the size is a concern we should consider other formats tbh, we rewrite the same things in every records like:
''fitacf.version" and "radar.version" 
"origin.command"
"origin.time"
many other fields...

mts299:
I don't want this to be a debate but it has been discussed many times that we shouldn't have partial records for the fact it's not user-friendly or scientific proper. Again this can wait for another release, I am just finally getting around to doing it.

mts299:
#223 there is where the discussion happened.

pasha-ponomarenko:
@mts299, @ecbland, I am not sure if this is an actual issue. 

I must admit I have got confused a bit here. From the first sight, the term "partial records" seems to imply that some legitimate fitted data are recorded and some are not. However, after I have read the preceding discussion, it seems to mean that the records that did not pass the quality selection criteria (i.e., not enough "good" lags or fitting is not converging etc) are actually not written into *.fitacf files.

>     You are throwing away potentially good data or user data due to the quality flag not equaling = 1 which was the if statement I changed. I don't know what scientific basis this flag is set but I assume the scientist should decide what is good and what is bad

I know these criteria quite well, and we have already relaxed them in FITACF3 by leaving only those which do make physical and/or statistical sense (FITACF3 White Paper).

Right now we have enough confusion among our highly qualified users even with the relaxed selection criteria in FITACF3, so I imagine it will be a lot more trouble if we actually remove all of them, especially when feeding the fitted data into gridding/mapping routines. 

Furthermore, changing this basic writing routine will also create data reproducibility problems.

Your thoughts?

pasha-ponomarenko:
@ecbland, does it just add zeroes?

ecbland:
> @pasha-ponomarenko does it just add zeroes?

Yes, for the file I tested it just adds zeros. And there has indeed been some confusion with data pre-selection criteria, which came up in #223 

I had a brief chat with @mts299 earlier today and we agreed to revisit this after we've finished with the RST4.6 release. I'll convert this to a draft in the meantime.

mts299:
@ecbland done 

`20210801.2001.00.dcn.fitacf` is good file to use since it has all partial records (will contact PI and @kevinkrieger on this file being in the distro) 

This will create a fitacf file with empty arrays instead of filled with zeros. 

``` bash
       -10.671273,     3.390764,       -0.670371,      -50.000000,     -2.981003,      -9.257982,      4.157047,       -1.136881,      -2.935142,
      1.456526,       -5.315253,      -4.691495
        short   "slist" [0]=
        short   "nlag" [0]=
        char    "qflg" [0]=
        char    "gflg" [0]=
        float   "p_l" [0]=
        float   "p_l_e" [0]=
        float   "p_s" [0]=
        float   "p_s_e" [0]=
        float   "v" [0]=
        float   "v_e" [0]=
        float   "w_l" [0]=
```

mts299:
@ecbland this is ready for review but will wait for the next release to add this as it might be more of a major release.

mts299:
Note this is a priority for Kathryn as mentioned in the DAWG meeting today, since this seems to trouble the SuperDARN Canada group and students we have.

PR #437, opened 2021-06-15
MLT_v2 DLM inverse keyword

egthomas:
This pull requests adds inverse MLT functionality for the IDL DLMs.  Previously when adding the MLT_v2 functions to the RST DLMs I forgot to include the ~~inverse~~ `mlt2mlon` keyword (which already exists / works for the native IDL function), so on the `develop` branch one cannot go from MLT to magnetic longitude (MLON), only MLON to MLT.

To test, try calling `MLTConvertYMDHMS_v2`, `MLTConvertYrsec_v2`, and/or `MLTConvertEpoch_v2` with the ~~inverse~~ `mlt2mlon` keyword set and DLMs enabled.  On this branch it will work, while on the `develop` branch it will not, eg

```
IDL> mlt = MLTConvertYMDHMS_v2(2021,6,15,0,0,0,18.)         
% Loaded DLM: MLTDLM.
IDL> lon = MLTConvertYMDHMS_v2(2021,6,15,0,0,0,18.,/mlt2mlon)
% Keyword MLT2MLON not allowed in call to: MLTCONVERTYMDHMS_V2
% Execution halted at: $MAIN$ 
```

*edited to use the correct IDL keyword*


ecbland:
@egthomas I can't get this to work on my machine. I tested it from a fresh install of RST. On this branch I get the same error as `develop`:

```
IDL> lon = MLTConvertYMDHMS_v2(2021,6,15,0,0,0,18.,/inverse)
% Loaded DLM: MLTDLM.
% Keyword INVERSE not allowed in call to: MLTCONVERTYMDHMS_V2
% Execution halted at: $MAIN$

IDL> help,/dlm,'mltdlm'
** MLTDLM - MLTDLM (loaded)
    Version: 2.0, Build Date: June 2017, Source: R.J.Barnes (JHU/APL) & E.G.Thomas (Dartmouth)
    Path: /home/emma/gitrepos/rst/dlm/mltdlm.so
```

I will take a closer look at the code ASAP, but do you have any suggestions for what could be wrong?

egthomas:
@ecbland sorry about that - I got my variables mixed up!  The IDL keyword for the inverse coordinate transformation is actually `mlt2mlon` (not `inverse`, which is used internally by the DLMs).  So the example from above would instead be something like

```
IDL> lon = MLTConvertYMDHMS_v2(2021,6,15,0,0,0,18.,/mlt2mlon)
% Loaded DLM: MLTDLM.
IDL> print, lon
      -20.361230
```

ecbland:
Ahhh, that explains it! I tested this for a few times & locations -- the `mlt2mlon` keyword works and gives the expected results. Merging this in...thanks @egthomas!

PR #438, opened 2021-06-16
Fixing segmentation fault in fov_plot when using -dot with invalid date

egthomas:
This pull request fixes a bug in `fov_plot` causing a segmentation fault when using the `-dot` command line option and the radar site information cannot be found.  For example, on the develop branch the following command will generate a valid figure (note the date is before the SCH radar end date):

```
fov_plot -x -st sch -d 19950101 -fov -ffov -coast -stereo -latmin 30 -fcoast -grd -grdontop -dot
```

sch_dot

while this second command will cause a segmentation fault (note the date is after the SCH radar end date):

```
fov_plot -x -st sch -d 19960101 -fov -ffov -coast -stereo -latmin 30 -fcoast -grd -grdontop -dot
```

The same behavior (ie segmentation fault) occurs on the `develop` branch for dates before a radar's official start date as well.  On this branch however, the latter statement will generate a valid figure (without a radar FOV or dot) as expected.

sch_empty


ecbland:
This works as described...thanks @egthomas!

PR #439, opened 2021-06-16
More fixes to fov_plot

egthomas:
This pull request makes a few more fixes to `fov_plot` (apologies for not including these all in one pull request).

First, this pull request adds date and time labels via the new `-time` flag (similar to `grid_plot` and `map_plot`), eg

```
fov_plot -x -fov -ffov -coast -stereo -fcoast -grd -grdontop -dot -d 19950101 -t 06:30 -st sch -time
```

time_label

Next, this pull request adds an `AACGM_v2` label to the bottom-right corner of the plot when using the `-mag` flag, or an `AACGM` label with the `-mag` and `-old_aacgm` flags (again similar to `grid_plot` and `map_plot`), eg

```
fov_plot -x -fov -ffov -coast -stereo -fcoast -grd -grdontop -dot -d 19950101 -t 06:30 -st sch -mag
```

aacgm_v2_label

```
fov_plot -x -fov -ffov -coast -stereo -fcoast -grd -grdontop -dot -d 19950101 -t 06:30 -st sch -mag -old_aacgm
```

aacgm_label

Finally, this pull request fixes a bug when using both the `-grdontop` and `tmlbl` flags together causing the plotted grids to be misaligned.  For example, the following command on the `develop` branch will cause the grids to be misaligned, eg

```
fov_plot -fov -ffov -coast -fcoast -stereo -grd -grdontop -dot -d 19950101 -t 06:30 -time -st sch -tmlbl
```

tmlbl_old

while on this branch the bug will be fixed, eg

tmlbl_new

(note - keeping this as a draft for now in case any other unexpected bugs appear)

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/fov_plot.1.10/doc/fov_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/fov_plot.1.10/doc/fov_plot.doc.xml
@@ -152,6 +152,8 @@
 </option>
 <option><on>-yoff <ar>yoff</ar></on><od>open the window <ar>ypad</ar> pixels from the top edge of the screen.</od>
 </option>
+<option><on>-time</on><od>plot the date and time of the plotted field of view.</od>
+</option>
 <option><on>-chisham</on><od>use the Chisham virtual height model.</od>
 </option>
 <synopsis><p>Plot the radar fields of view.</p></synopsis>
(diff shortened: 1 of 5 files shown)


ecbland:
I tested all the new options & bugfixes here with various dates and radars. Everything works as described. 

Something that would be nice to do here is to automatically set `-lat -90` if the radar is in the southern hemisphere. I'll leave that up to you if you want to include that in this PR

egthomas:
@ecbland agreed, that was  a pain always having to remember to set `-lat -90` to get a plot of the Southern Hemisphere.

`fov_plot` should now automatically adjust the sign of the `-lat` and `-latmin` values (whether or not the user provides them) according to the hemisphere of the selected radar.  If no radar is provided, then it will default to the Northern Hemisphere / TST radar (as it currently does on the `develop` branch).

ecbland:
> agreed, that was a pain always having to remember to set -lat -90 to get a plot of the Southern Hemisphere.

Poor southern hemisphere, all tucked away down there! Everything in this PR seems to be working, so I will merge it now :smile:

PR #441, opened 2021-06-21
Release: RST 4.6

ecbland:
# Release RST 4.6 

- [ ] PATCH
- [X] MINOR
- [ ] MAJOR

## Deadline: 19 July 2021
Please comment in the Pull Request conversation if you need additional time to test the release.

## Testing

- [ ] Mac OSx
- [x] Linux deb - Debian, Ubuntu, Mint
- [x] Linux rpm - OpenSuse, Fedora, CentOS

*Do any binaries/libraries require special attention during testing?*

Below are the major updates. Of course these have already been tested, but feel free to try to "break" them :smile: 

- `make_fit -fitacf-version 3.0` - fixed "divide by zero" bug that affects noise level determination for BOREALIS radars (#420 . Added Shepherd (2017) elevation angle algorithm
- `make_grid` - see #373 
- `fit_speck_removal` - new binary, see #400 
- `datdump` - new binary, see #376
- `fov_plot` - new functionality, see #440 
- IDL DLM for MLT_v2 - see #437 
- [Documentation](https://radar-software-toolkit-rst.readthedocs.io/en/develop/index.html) - lots of updates, comments welcome. 

### How to test

Install git, and then run:

```
git clone https://github.com/SuperDARN/rst
cd rst
git checkout release/rst-4.6
make.build && make.code
```

Alternatively, download a [zip archive of the release branch](https://github.com/SuperDARN/rst/archive/refs/heads/release/rst-4.6.zip) and then run `make.build && make.code` as usual. You may need to update some environment variables for this to work. 

From there, please use RST as normal and play with the new features listed above. 

## Checklist

- [X] Update `radar.dat` -- not required this time
- [x] Update hardware files (if required)
- [X] Update `.rst.version`
- [x] Update author list in `.zenodo.json`. **@ksterne and I will provide a draft author list ASAP**
- [x] Write a list of major updates for the release description
 
## Extra Notes

All hardware files have been updated to match the new file format on the `hdw` repo. 
The DCE and DCN radars have updated information (see https://github.com/SuperDARN/hdw/pull/1). 

## Approving the release

After testing, please state which operating system you used and what you tested.

If you found no bugs during testing, please *approve* the PR:

1. Click on `files changed` top right
2. Click on `start review` green top right  button
3. In the comment box, state which operating system you used for testing, and any other helpful information
4. Select `approve` and `submit review`

PR #445, opened 2021-07-29
Cleanup readthedocs

ecbland:
This PR is a quick cleanup of `readthedocs` to be merged into the `release/rst-4.6` branch (https://github.com/SuperDARN/rst/pull/441#issuecomment-888627542)

I wanted to collapse the headings in the navigation bar by default, but that doesn't seem to be possible at the moment. As a compromise, I moved the developer guide to the end of the table of contents & navigation bar, so that the other content is more visible to users. 

I also fixed some broken links and warnings.

To test, you can compile the documentation locally: 
```
cd $RSTPATH
pip install mkdocs
mkdocs serve
```
The online build for this branch is here: https://radar-software-toolkit-rst.readthedocs.io/en/fix-readthedocs/


ksterne:
Looks like this all works.  That's odd on the links (I think some were ones I added), I guess the auto-magically figure out the right path?  I'm guessing we might not be able to have two files of the same name if that's the case.  Not that I think we will, but just something to keep in mind.  This looks better with the developer info towards the bottom.  We can looking making it look even cleaner in the future.

PR #446, opened 2021-08-03
Merging `master` back into `develop`

ecbland:
Thanks to everyone who contributed to development and testing of RST 4.6! This PR is just admin: merging `master` back into `develop`. Once this is merged we can continue development as usual on the `develop` branch...there are already several PRs waiting for review :smile: 

Feel free to click the big green button if you feel like celebrating :rocket:

PR #447, opened 2021-08-04
Revert `0026.rfc.xml` to previous version

ecbland:
[`0026.rfc.xml`](https://github.com/SuperDARN/rst/blob/master/codebase/superdarn/src.doc/rfc/0026.rfc.xml) was accidentally modified while cleaning up the license notices. This PR reverts this file to the previous version using `git checkout v4.5 -- codebase/superdarn/src.doc/rfc/0026.rfc.xml`

Test: `make.doc` should run on this branch without skipping over this file or crashing.

### DIFF
--- a/codebase/superdarn/src.doc/rfc/0026.rfc.xml
+++ b/codebase/superdarn/src.doc/rfc/0026.rfc.xml
@@ -2,6 +2,7 @@
 <rfc>
 <id>0026</id>
 <title>Radar Operating System: Proposed System Architecture</title>
+<author>R.J.Barnes</author>
 <summary><p>An overview of the proposed system architecture for the Radar Operating System.</p></summary>
 <body>
 <h4>1. Introduction</h4>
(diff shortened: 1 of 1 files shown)


egthomas:
@ecbland thanks, this fixes the compilation error on `0026.rfc.xml`.  There are still a lot of duplicate Copyright lines in the example `.c` files used by the XML documentation, eg 

https://github.com/SuperDARN/rst/blob/ed604f958384b390fb48dc5f120899f62024e533/codebase/superdarn/src.lib/tk/cnvmap.1.17/doc/src/CnvMapFread/CnvMapFread.c#L1-L7

In total there are nearly 300 files affected.

ecbland:
> @egthomas There are still a lot of duplicate Copyright lines in the example .c files used by the XML documentation

Ok, I'll look into this ASAP in a separate PR

ksterne:
Is this something that needs to be merged into the `master` branch?  I could see either way since this fixes something that seems to crash on at least one system (I think there was at least one test where it didn't cause a crash).  On the other side, this is on documentation and not critical for the main processing routines.  Again, I'm indifferent either way and just wanted to stop to think some on it.

ecbland:
> @ksterne Is this something that needs to be merged into the master branch?

I don't think it matters either way. Note that if we merge this PR before merging #446 , the same file will end up on both branches anyway.

> (I think there was at least one test where it didn't cause a crash)

It doesn't crash on my system (Linux Mint 20, `gcc (Ubuntu 9.3.0-10ubuntu2) 9.3.0`). It just skips over the affected file.

ksterne:
As might be expected, Ubuntu 20 has a similar performance where it skips over the RFC 0026 file.  @egthomas, can you pass along the OS where `make.doc` is crashing?

egthomas:
@ksterne sorry for any confusion, but I'm not sure where it was said that `make.doc` crashes entirely.  It fails to compile `0026.rfc.xml` as @ecbland noted, where a failure means that it skips over that file and continues to the next in the list.

egthomas (approved):

ksterne:
Thanks for the clarification @egthomas, reading over it again, I see where my 'skim-reading' thought it was failing to compile but you had originally noted the same behavior that it was finishing but this one was skipped over.  Since it's not breaking anything and this particular documentation is very old and out-of-date, I'd say it's better to merge it into `develop`.  Thoughts @ecbland?

Issue #451, opened 2021-08-17
[EHN] make system refactored 


mts299:
# New Feature

**Name**: make.code to make code? 

## Description & User Interface

As I look into how to resolve some of the ongoing RST make issues as posted in #450 along with many other annoying issues that have been brought up a lot:
- [ ] long compile time when a few files are changed (I know you can compile a single directory with their makefile but if you change various folders then it gets a bit confusing what to run first)
- [ ] not including netcdf/NASA cdf lib if you are not using those binaries/features of RST 
- [ ] make -g ability (though easy to solve) 

I noticed we have our own "make" code written in c called `makeall` this causes a lot of tight coupling to this one function as I cannot parse or want to stare at what would be innards of `makeall`. This also explains the crashes I get with `make` when I swap it out for `makeall` 

I plan to do some investigation work and refactor the make code to use GNU make instead of our own `makeall` to allow better flexibility and understandability of how things compile in our code. 

https://www.oreilly.com/library/view/managing-projects-with/0596006101/ch06.html 

The plan is to make sure nothing changes on the user level with the advantages of the above listings and the same for the developer side. 

## RST Checklist

- [x] Does the feature fit within RST's scope?
- [x]Is this a major or minor change?
    - [x] Major (e.g. includes large amounts of new code, existing code significantly restructured, or impacts on other RST libraries/binaries)
-  [x] Can the code be released under the GPL v3.0 license?
-  [x] Does the source code include copyright and authorship information? (can be added later at the pull request stage)

## Timeline

**1 week - month** making this a high priority for myself as it is showing to be a problem for using it on my HPC system where I don't have admin privileges and it forces me to use compilers that seem to crash RST. #450 

## Extra Notes

This might not be as hard since we actually follow the suggested structure provided by the link: https://www.oreilly.com/library/view/managing-projects-with/0596006101/ch06.html

I will give this a try and see where I get to and report either in PR if worth it or close the issue if not.

PR #456, opened 2021-08-25
[New] make system for RST

mts299:
# Scope 

So this is less refactoring right now as I just made a makefile in the root directory to essentially do what `make.build` and `make.code` does. In the future, we can deprecate those two scripts and use the `makefile` as it allows more flexibility and is the built-in make system than what RST has built into it. Also, the hope is more flexibility and less maintenance. 

**issue** #451 

I will be building off of this branch with other branches for more in-depth refactoring to fix ongoing users and including myself problems with install RST on remote servers. 

## Approvals: 3 
- [ ] Mac OS user @billetd or @mtwalach ?
- [ ] Debian user (Ubuntu) @ksterne ? @ecbland ? 
- [ ] Fedora/RPM user 
- [ ] IDL test @ecbland ?

## Test

``` bash 
git checkout ehn/refactor_make
cd rst 
make
```
Should compile RST and you should be able to use `make_fit` and other binaries with no issues. 

```bash 
make TARGET=clean
```
is a nother option to run make clean on all of RST files. 

It does not compile the documentation.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1,0 +1,37 @@
+# (C) Copyright 2021 SuperDARN Canada, University of Saskatchewan
+# Author(s): Marina Schmidt
+# 
+#
+lib_analysis = $(wildcard ./codebase/analysis/src.lib/*/*/src/)
+lib_base = $(wildcard ./codebase/base/src.lib/*/*/src/)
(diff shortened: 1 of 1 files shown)


ecbland:
@mts299 I think you mentioned to me that you were going to close this PR and try another approach?

PR #463, opened 2021-09-17
Checking the value of qflg in fit_speck_removal

ecbland:
This PR adds a check in `fit_speck_removal` to ensure that `qflg` can only have the value 0 or 1. This will always be the case for data processed with `fitacf3.0`, but @egthomas pointed out to me that `fitacf2.5` uses other integer values of `qflg` to mark rejected ACFs. I believe that only the records with `qflg=1` are written to file, but we cannot find which part of the code would reject the data assigned `qflg=8` here:

https://github.com/SuperDARN/rst/blob/c9345ca78fe88c0279dc09fbb9eaf91da512d684/codebase/superdarn/src.lib/tk/fitacf.2.5/src/do_fit.c#L268-L271

I searched 4 months of `fitacf2.5` data from 11 radars and did not find a single record with `qflg=8`. But it's straightforward to check the `qflg` value so that's what I did.

**Testing:** Since I can't find a specific test case where this modification would matter, please just check that `fit_speck_removal` produces identical output on this branch and `develop` for a fitacf file of your choice. 

*I also updated the description and comments in this routine to avoid using the word "filter". There is some potential for confusion with traditional median filtering procedures that modify the fitted values in the neighbouring cells. In contrast, `fit_speck_removal` just removes isolated points without modifying any of the remaining fitted data.*

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/doc/fit_speck_removal.doc.xml
+++ b/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/doc/fit_speck_removal.doc.xml
@@ -17,12 +17,13 @@
 </option>
 
 <synopsis><p>Removes salt and pepper noise from a <code>fitacf</code> file.</p></synopsis>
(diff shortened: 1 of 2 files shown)


egthomas:
Tested and approved.

PR #471, opened 2022-01-20
Adding Christmas Valley virtual height model and modularized rpos_v2 library

egthomas:
This pull request adds a new set of modularized functions in `rpos_v2` of the existing `rpos` library, including stand-alone functions for each of the following virtual height models:

- standard model
- Chisham et al [2008] Saskatoon model
- Thomas and Shepherd [2022] Christmas Valley model (submitted to Radio Sci)

There is also a new binary added to the `testing` directory called `rpos_v2_test` which allows users to test each of the different virtual height models for different radars, beams, and range gates (the output is geographic latitude and longitude printed to stdout).  This binary is primarily intended for testing the `rpos_v2` functions and can be deleted before merging into `develop` if desired.

Note that none of the existing geolocation, gridding, plotting, etc code has been modified - I have simply added a parallel set of (more generalized) functions so that if someone wants to use them they can.


ksterne:
Should the new files have copyright lines added to them?  I only ask because it's been a while, but I'm fairly certain that they should.  This reference should help:  [copyright_license.md](https://github.com/SuperDARN/rst/blob/master/docs/dev/copyright_license.md).

ecbland:
@egthomas does this update replace all of the functionality of `rpos` without changing the results? Just wondering whether the original `rpos` functions can be completely replaced with these and keep the same names (i.e. without adding `_v2` to everything)

pasha-ponomarenko:
I just wonder what is applicability of this model in terms of mid-latitude, auroral and polar cap radars?  
How much testing has been done in that respect? From what I can deduce, this model might be restricted to mid latitudes. 
If this is the case, it should probably be explicitly mentioned somewhere.

JWiker:
@egthomas Would it be possible to update the `rpos_v2` and `rpos_v2_test` headers to include file/function descriptions?

JWiker:
> 
> > @egthomas Would it be possible to update the rpos_v2 and rpos_v2_test headers to include file/function descriptions?
> 
> @JWiker thanks for the suggestion. For most previous libraries/functions (and in this pull request) I've tried to add in-line comments to the functions themselves in the `.c` files as appropriate. If there's a push for including that information in the header files moving forward then it could be done.

@egthomas That'd be really helpful, thanks. As a newcomer to the repo, it'd be super useful to be able to understand what each file does without having to go through more detailed in-line comments.

mts299:
> @egthomas That'd be really helpful, thanks. As a newcomer to the repo, it'd be super useful to be able to understand what each file does without having to go through more detailed in-line comments. 

Even as a non-newcomer code documentation is key especially understanding the general gist of a function, parameters (what they mean and types), possible return values and possible errors are always desired in any code base. 

Would it be possible to just add your algorithm in as a new algorithm library instead of rposgeo_V2? version numbers are miss leading at best especially in already versioned git repository. V2 also could signfy better then current where this is not true for all radars? Maybe spherical_rpos or something more meaningful. 

Granted I am curious to prevent bloat, if this is not added to anything like plotting or gridding, is there intent to add it there some time in the future? Otherwise, would it be better to live in another repository like AACGM does?

pasha-ponomarenko:
> @pasha-ponomarenko yes that is absolutely true (and something we discuss in the referenced manuscript), hence why I am only adding the libraries to use this model and not actually connecting it to any of the gridding or plotting binaries. Of course, a similar warning could be applied to the standard and Chisham et al virtual height models that they were each derived using data from only a single auroral-zone radar and may not be suitable for mid-latitude or polar radars.
> 
> We do compare our Christmas Valley / midlatitude ionospheric model to the Chisham et al model, and other than the range of the 0.5-hop E- to F- transition and 0.5- to 1.5-hop F-region transition, they largely agree with one another (except we drop the pseudo virtual height concept for implementation of 1.5-hop scatter in our model).

Thanks, @egthomas! As your RSci manuscript is not readily available, I wonder what are the main advantages granting addition of this model to the RST. As you stated before, it mainly agrees with Gareth's model.  Does abandoning the pseudo-VH concept have anything to do with the discrepancies at the transition zones? Also, it would be prudent to check how the model works at the other (mid-latitude?) radars with reliable elevation measurements.

egthomas:
> Thanks, @egthomas! As your RSci manuscript is not readily available, I wonder what are the main advantages granting addition of this model to the RST. As you stated before, it mainly agrees with Gareth's model. Does abandoning the pseudo-VH concept have anything to do with the discrepancies at the transition zones? Also, it would be prudent to check how the model works at the other (mid-latitude?) radars with reliable elevation measurements.

@pasha-ponomarenko this is my first time doing so, but there is actually a pre-print available through ESSOAR here: https://doi.org/10.1002/essoar.10510245.1 . I would prefer to let the peer review process to run its course, and will be talking about this next week at USask, so hopefully we can follow up there.  Maybe one high-level difference between the new CV and Chisham models which was not obvious from my prior comments is that the new CV model has a ground scatter component as well as an ionospheric scatter component.  So (in our opinion) the ground scatter component performs dramatically better than any existing virtual height model when properly applied to ground scatter.

pasha-ponomarenko:
@egthomas, thanks for the reference to the preprint. I believe that it should have been added to this issue from the beginning as there was virtually no information about this apparently significant new option in the geolocation package. 

On the RSci peer review process, it has its own limitations, especially if the reviewers are not closely familiar with SuperDARN specifics. From that prospective presenting such material at our seminar is a good idea. 

I would reiterate that the statement on the network-wide applicability would need to be supported by a network-wide testing.

egthomas:
After some unfortunate delays, our Radio Sci manuscript has now been accepted for publication and I have updated the model coefficients to match the final values.

ecbland:
Nice paper @egthomas! To move this discussion along (since @egthomas has done a good job answering questions so far, and there is a peer-reviewed article connected to this work): 

1. Let's support the publication of new analysis software so that the research can have an impact. Including this library in the RST makes it findable, accessible, etc.
2. Information about the model and its potential limitations are described in the accepted manuscript. 
3. These routines will only be called if the user chooses to call them. Until @egthomas or someone else wants to connect these routines to `make_grid` etc, I'd say that network-wide testing is unnecessary. 

A general observation about item 1 above: Authors would of course like to say in their manuscripts that the software has been added to the RST, but the SuperDARN community often prefers to see an accepted manuscript (and perhaps additional validation) before adding new analysis routines to the RST. So we have a chicken-and-egg type problem. This would be worth a broader discussion during a working group meeting.

egthomas:
@ecbland thanks for the kind words.

> A general observation about item 1 above: Authors would of course like to say in their manuscripts that the software has been added to the RST, but the SuperDARN community often prefers to see an accepted manuscript (and perhaps additional validation) before adding new analysis routines to the RST. So we have a chicken-and-egg type problem. This would be worth a broader discussion during a working group meeting.

I agree that this is a difficult line to walk - we wanted to indicate that we had taken the extra step of adding our algorithm to the software, without being too presumptuous about the code review aspect (note we did note say which branch the code was added to!).

In general, I believe that requiring a journal publication is an unnecessarily high bar to set for all algorithms submitted for potential inclusion in the RST.  However, if the algorithm is to be included as part of a manuscript submission (or is the main focus) then the full peer review process should have a chance to run its course before including the code in an official release.

codebase/superdarn/src.bin/tk/testing/rpos_v2_test.1.0/rpos_v2_test.c
```diff
@@ -0,0 +1,188 @@
...
+
+  char *dtetxt=NULL;
+  double dval=0;
+  int yr,mo,dy,hr,mt,isc,usc;
+  double sc;
+
+  int stid=1;
+  int bm=7;
+  int rn=20;
+  double height=300;
```
aburrell, line 88:
```suggestion
  double height=300.0;
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/rpos_v2.c
```diff
@@ -0,0 +1,255 @@
...
+ * This function converts a gate/beam coordinate to geographic
+ * position. The height of the transformation is given by height -
+ * if this value is less than 90 then it is assumed to be the
+ * elevation angle from the radar. If center is not equal to zero, then
+ * the calculation is assumed to be for the center of the cell, not the
+ * edge. The calculated values are returned in geocentric spherical
+ * coordinates (rho,lat,long). If model is equal to one the Chisham
+ * et al [2008] virtual height model is used; if model is equal to
+ * two the Thomas and Shepherd [2022] virtual height model is used;
+ * otherwise,the standard virtual height model is used.
```
aburrell, line 46: I recommend reformatting this as a list, so it is easier to read when scanning what the different `model` values mean.

codebase/superdarn/src.lib/tk/rpos.1.7/src/rpos_v2.c
```diff
@@ -0,0 +1,255 @@
...
+ **/
+void RPosGeo_v2(int center, int bcrd, int rcrd, struct RadarSite *pos,
+                int frang, int rsep, int rxrise, double height,
+                double *rho, double *lat, double *lng,
+                double hop, int model, int gs, int rear) {
+
+  double d,rx;
+  double azi,elv,vh,grng;
+  double rrad,rlat,rlon,del;
+  double range_edge=0;
```
aburrell, line 56:
```suggestion
  double range_edge=0.0;
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/rpos_v2.c
```diff
@@ -0,0 +1,255 @@
...
+  int pseudo=0;
+
+  if (center==0) range_edge=-0.5*rsep*20/3;
```
aburrell, line 59:
```suggestion
  if (center==0) range_edge=-0.5*rsep*20.0/3.0;
```

Make it clear that there is no integer division going on.

codebase/superdarn/src.lib/tk/rpos.1.7/src/rpos_v2.c
```diff
@@ -0,0 +1,255 @@
...
+}
+
+
+/* Calculate the coning angle correction for a given radar, beam
+ * number, elevation angle, and FOV direction (ie front vs rear) */
+double calc_coning_angle(struct RadarSite *pos, int bmnum, double elv,
+                         int center, int rear) {
+
+  double azi,offset,psi;
+  double bm_edge=0;
```
aburrell, line 219:
```suggestion
  double bm_edge=0.0;
```

aburrell (commented):
Code review with suggestions.  Going to try and run it soon.

In general, I think you could stand to add more comments to the code, but I could follow it so I'm not going to complain beyond this sentence.

egthomas:
Thanks @aburrell - I've just updated the code to address your specific comments.

codebase/superdarn/src.lib/tk/rpos.1.7/src/rpos_v2.c
```diff
@@ -0,0 +1,255 @@
...
+}
+
+
+/* Thomas and Shepherd [2022] Christmas Valley virtual height model */
+double calc_cv_vhm(double r, int gs, double *hop) {
+
+  double xh;
+
+  if (gs) {
+    /* Ground scatter virtual height model */
```
aburrell, line 152:
```suggestion
    /* Ground scatter virtual height model {1E, 1F, 2F} */
```

codebase/superdarn/src.lib/tk/rpos.1.7/src/rpos_v2.c
```diff
@@ -0,0 +1,255 @@
...
+    *hop=1.0;
+
+    if (r<1140.0) xh=A_const[0]+B_const[0]*r+C_const[0]*r*r;
+    else if (r<=3262.5) xh=A_const[1]+B_const[1]*r+C_const[1]*r*r;
+    else {
+      xh=A_const[2]+B_const[2]*r+C_const[2]*r*r;
+      *hop=2.0;
+    }
+  } else {
+    /* Ionospheric scatter virtual height model */
```
aburrell, line 166:
```suggestion
    /* Ionospheric scatter virtual height model {1E, 1F, 2F} */
```

aburrell (commented):
Ok, I read the paper and verified the coefficients are correct.  I tried running `rpos_v2_test` and is it supposed to work without any input?  That was confusing, but I see there are defaults in the test function.  It would be nice if the test output specified some of the inputs, but that's just my preference and not terribly important.

If I run `rpos_v2_test -stid 10 -bm 7 -rn 25 -cv` I get a seg fault.  Probably because HAN is down right now.  A graceful exit would be better.  `rpos_v2_test -stid 10 -bm 7 -rn 25 -cv -d 20040101` works.

I used a SPEAR experiment at HAN to see how well the CV IS model worked.  The midpoint of the range gates was ~40 on beam 9 (2006-10-12) and yields a lat of 77.33.  This is very close to the reference lat of 77.89.  The Chisham model gives 77.48 (closer, as expected for a higher lat model) and the standard model gives 78.97 (too high, as expected).  This agrees well with the histograms I made of the calibrated elevation angles, Chisham model, and standard model for the selected backscatter (figure below).  So I think the model implementation is working as desired.
spear_test

egthomas:
> I tried running rpos_v2_test and is it supposed to work without any input?

Yes, if you run `--help` it should indicate the default station (1 / GBR), beam number (7), range gate (20), etc.

> If I run `rpos_v2_test -stid 10 -bm 7 -rn 25 -cv` I get a seg fault. Probably because HAN is down right now. A graceful exit would be better. `rpos_v2_test -stid 10 -bm 7 -rn 25 -cv -d 20040101` works.

Thanks for catching that - I just added a check for invalid radar / date combinations.  Ultimately though the `rpos_v2_test` binary was meant as a simple testing tool and doesn't necessarily need to be kept in the RST.

aburrell (approved):
I think your test script is a useful tool for exploring the different virtual height models, I hope it stays in.  I can totally picture using it for a presentation or paper.

PR #474, opened 2022-02-23
Print despecking statistics in the terminal

ecbland:
@alexchartier suggested that we update `fit_speck_removal` to print some basic statistics in the terminal about how much scatter was removed. 

Here's an example of the new output: 
```
$ fit_speck_removal 20180308.0201.00.cly.fitacf3 > /dev/null

Number of echoes removed: 3999 of 33074 (12.1%)
```

I also added a new command line option `-quiet` which allows the user can suppress this output if they want to. By default, the statistics will be printed in the terminal.

------------

**Note:** The percentage of scatter removed can seem alarmingly high when the amount of ground/ionospheric scatter is low. In the following example, we see that the time periods with almost no ionospheric/ground scatter have a high percentage of echoes removed (mostly meteor scatter in this example). There is no problem here of course...and we leave it to the user to interpret the results. 

```
$ for file in 20180308*cly.fitacf3; do echo $file ; fit_speck_removal $file > /dev/null ; done

20180308.0001.00.cly.fitacf3
Number of echoes removed: 3002 of 53806 ( 5.6%)
20180308.0201.00.cly.fitacf3
Number of echoes removed: 3999 of 33074 (12.1%)
20180308.0401.00.cly.fitacf3
Number of echoes removed: 1761 of 3576 (49.2%)
20180308.0601.00.cly.fitacf3
Number of echoes removed: 2180 of 7612 (28.6%)
20180308.0801.00.cly.fitacf3
Number of echoes removed: 2285 of 6632 (34.5%)
20180308.1001.00.cly.fitacf3
Number of echoes removed: 3280 of 7232 (45.4%)
20180308.1201.00.cly.fitacf3
Number of echoes removed: 5658 of 37055 (15.3%)
20180308.1401.00.cly.fitacf3
Number of echoes removed: 5379 of 41404 (13.0%)
20180308.1601.00.cly.fitacf3
Number of echoes removed: 5565 of 36754 (15.1%)
20180308.1801.00.cly.fitacf3
Number of echoes removed: 5765 of 46067 (12.5%)
20180308.2001.00.cly.fitacf3
Number of echoes removed: 4548 of 46838 ( 9.7%)
20180308.2201.00.cly.fitacf3
Number of echoes removed: 3917 of 52648 ( 7.4%)
```

20180308 cly fitacf3
20180308 cly despeck fitacf3

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/doc/fit_speck_removal.doc.xml
+++ b/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/doc/fit_speck_removal.doc.xml
@@ -5,13 +5,15 @@
 <location>src.bin/tk/tool/fit_speck_removal</location>
 
 <syntax>fit_speck_removal --help</syntax>
-<syntax>fit_speck_removal [-vb] <ar>fitacfname</ar></syntax>
+<syntax>fit_speck_removal [-vb] [-quiet] <ar>fitacfname</ar></syntax>
 
 <option><on>--help</on><od>print the help message and exit.</od>
 </option>
 <option><on>--version</on><od>print the RST version number and exit.</od>
 </option>
-<option><on>-vb</on><od>verbose. Log information to the console.</od>
+<option><on>-vb</on><od>print extra information in the terminal.</od>
+</option>
+<option><on>-quiet</on><od>prevent statistics from being printed in the terminal.</od>
 </option>
 <option><on><ar>fitacfname</ar></on><od>filename of the input <code>fitacf</code>-format file.</od>
 </option>
--- a/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/fit_speck_removal.c
+++ b/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/fit_speck_removal.c
@@ -25,9 +25,9 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <https://www.gnu.org/licenses/>.
        
 Modifications:
-	E.C.Bland, University Centre in Svalbard, 2021-09-17: fix handling of qflg>1 data from fitacf2.5, 
+    2021-09-17 E.C.Bland, University Centre in Svalbard (UNIS): fix handling of qflg>1 data from fitacf2.5, 
 	           which in rare cases might include values other than qflg=0 for rejected ACFs.
-
+    2022-02-23 E.C.Bland, (UNIS): add statistics on number of rejected ACFs
 */
 
 
@@ -253,6 +253,8 @@ int main (int argc,char *argv[]) {
     free_parameters(prm, fit, fp, qflgs);
     exit(-1);
   }
+  int echoes_total=0;
+  int echoes_removed=0;
   do {
   
     if (vb) {
@@ -319,7 +321,11 @@ int main (int argc,char *argv[]) {
         
         // Remove record if median=0 (sum of qflgs < 5)
         if (sum < 5) 
+        {
             fit->rng[range].qflg=0;
+            echoes_removed+=1;
+        }
+        echoes_total+=1;
       }
     }
     irec[beam][channel]++;
@@ -359,6 +365,9 @@ int main (int argc,char *argv[]) {
   } while (FitFread(fp,prm,fit) !=-1);
 
   fclose(fp);
+  
+  // Print statistics
+  fprintf(stderr,"Number of echoes removed: %d of %d (%4.1f%%)\n",echoes_removed,echoes_total,100*(float)(echoes_removed)/(float)(echoes_total));
 
   // Free memory
   free_parameters(prm, fit, NULL, qflgs);
--- a/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/fit_speck_removal.c
+++ b/codebase/superdarn/src.bin/tk/tool/fit_speck_removal.1.0/fit_speck_removal.c
@@ -103,6 +103,7 @@ void free_parameters(struct RadarParm *prm, struct FitData *fit, FILE *fp, qflgD
 int main (int argc,char *argv[]) {
   
   unsigned char vb=0;
+  unsigned char quiet=0;
   unsigned char help=0;
   unsigned char option=0;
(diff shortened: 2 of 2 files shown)


pasha-ponomarenko:
@ecbland, I wonder if the "quiet" mode is more desirable here just to avoid being spammed by all uninitiated users rather then by those who are actually interested in the stats.
Another thought is that the chosen metric -- removed echoes/"good" echoes -- is intrinsically somewhat confusing as it indeed strongly depends on the number of physically meaningful echoes. I am not sure how this would work, but what about choosing the total number of  range/time cells as a reference (something like 16 beams x 70 gates x 120 scans for a two-hour datafile). In this case for the "best" case scenario (20180308.0001.00.cly.fitacf3) we will have  2.2% of removed echoes  on the background of  40% of of the "good" echoes, while for the "worst" case (20180308.1001.00.cly.fitacf3) it will be 2.4% versus  5.4%, respectively. I feel that this metric is less confusing.

Just a thought.

alexchartier:
@ecbland Thanks for putting this in. I think the percentage is great. Maybe "removed echoes / total echoes" would be a better metric than "removed / good"? 

My suggestion is that we include fit_speck_removal (or at least the interference rejection component of it) into fitACF (of course taking care not to reject the near-field "meteor" stuff in the first ~5-10 rangegates). At that point, the commandline text output would let the non-expect user know that data-points are being rejected. The expert user could either flag "no_despeckling", or "quiet" as appropriate.

pasha-ponomarenko:
@alexchartier, this issue has already been discussed extensively within this WG, and the output has been posted at the end of this report: 
https://superdarn.github.io/dawg/fitacf3/
(click on "_Integration of make_fit with despecking routine_")

pasha-ponomarenko:
I meant  integration of the de-specking routine into make_fit, of course ;-)

ecbland:
> @alexchartier Maybe "removed echoes / total echoes" would be a better metric than "removed / good"?

Perhaps there's some confusion here? The code is already calculating "removed echoes / total echoes"

pasha-ponomarenko:
@ecbland, I guess Alex meant the same stuff as me: total number of **potential** echoes, i.e., the number of cells in the beam /range gate/time matrix.

alexchartier:
@pasha-ponomarenko No I think @ecbland's current metric is good. Earlier you described it as "_removed echoes/"good" echoes "_ which would cause problems. 

Regarding the question of what filtering happens inside make_fit, I do remember the discussion with the PIEC, but I still think there's a good case to include interference rejection in make_fit as default. 
The big issue is that non-expert users (e.g. me a couple of years ago) will find all sorts of interference in their data, and might either assume it's all junk or waste effort re-making their own interference rejection algorithms.
The main objections seem to be:
a) reduced transparency (the user would not know how much interference is in their data)
b) Loss of meteor scatter

I think a) can be dealt with by Emma's text-printing and b) can be accomplished by leaving the first few rangegates untouched. Maybe I'm missing something though - why does the non-despeckled fitACF-v3 have so much more interference than the equivalent v2.5 data?

alexchartier:
@mts299 I'm not speaking for the PIEC, just as someone who was tripped up by a change in the way the code works. So the following is just a suggestion from an interested party, nothing more. I have read the reports, and I'm not questioning the fact that fitACF 3 is doing better overall and has many improvements (in particular the accuracy of the data has likely been much improved, in my opinion). However, I am seeing a lot of interference in v3 data that was not there in v2.5. Note this is removed by the despeckling, so it's not a "dealbreaker". Interference rejection is just something that a non-expert might expect to be done by default, as early as possible, so when they see it in the fitACF output they might question the data quality. I think we should reject interference by default, as early as possible in the chain, and report what we're doing via @ecbland 's new onscreen messaging.

mts299:
@alexchartier thank you for the feedback. Not trying to offend, just reminding that this topic is a scope creep for this PR and maybe a better dealt-in an issue or somewhere else.  

@ecbland I will test this tomorrow 👍🏻

alexchartier:
@mts299 None taken :) I was just trying to explain the context for this request.

mts299:
Alright... 2 weeks later... I tested the code and works as follows: 

# Test code

``` bash
#!/bin/bash

FITACF_FILE=/home/marina/superdarn/data/fitacf/20060916.0402.00.han.fitacf
SOUTH_FILE=/home/marina/superdarn/data/fitacf/20220108.0801.00.dcn.fitacf

echo "Normal Test"

fit_speck_removal ${FITACF_FILE} > test.despeck.fitacf
fit_speck_removal ${SOUTH_FILE} > test.despeck.fitacf 

echo "with option -quiet "

fit_speck_removal -quiet ${FITACF_FILE} > test.despeck.fitacf
fit_speck_removal -quiet ${SOUTH_FILE} > test.despeck.fitacf 

rm test.despeck.fitacf
```

output: 

``` bash 
Normal Test
Number of echoes removed: 10996 of 50470 (21.8%)
Number of echoes removed: 3514 of 33081 (10.6%)
with option -quiet 
```

Looks to be working :)

mts299 (approved):

ecbland:
@pasha-ponomarenko Thanks for the comments and sorry for the delay in responding to you:

> I wonder if the "quiet" mode is more desirable here just to avoid being spammed by all uninitiated users rather then by those who are actually interested in the stats.

Maybe we can revisit this if we do end up getting spammed? I expect that anyone using this routine will quickly go to the documentation and figure it out themselves.

> Another thought is that the chosen metric -- removed echoes/"good" echoes -- is intrinsically somewhat confusing as it indeed strongly depends on the number of physically meaningful echoes. I am not sure how this would work, but what about choosing the total number of range/time cells as a reference (something like 16 beams x 70 gates x 120 scans for a two-hour datafile).

This would need additional information printed in the terminal so that the user would know how the calculation has been done. Indeed it's not obvious how to determine a suitable reference (especially since time resolution, number of beams, number of channels, etc, etc, can change). I see more potential for confusion compared to just a simple percentage. I expect that anyone using this routine will have read the documentation and would be able to interpret an unexpectedly-high percentage for what it is.

pasha-ponomarenko (approved):

pasha-ponomarenko:
Apologies for the delayed review -- just overloaded with other stuff.

Issue #457, opened 2021-08-25
Removal or relocating of testing/map_potential


mts299:
# Question

While working on #456 I ran into some inconsistencies in RST folder systems and I found one folder with no makefile. 
`codebase/superdarn/src.bin/tk/testing/map_potential.1.0/` Wondering if we want to remove it or relocate it as it doesn't creat a binary and it looks like testing scripts and data files?

## Category

- [ ] Usage
- [ ] Troubleshooting
- [x] Software Design
- [ ] Data related
- [ ] Capabilities
- [x] Clarity
- [ ] Workflow

PR #476, opened 2022-03-01
Dep/remove map potential test

mts299:
issue: #457 

Removing the folder map_potential.1.0 as it is just scripts and data files. This should stay with the user testing it not in the software code.


ecbland:
This looks good, merging in in...

PR #480, opened 2022-03-07
Fixing bug in IDL DLM RadarEpochGetSite and RadarYMDHMSGetSite

egthomas:
This pull request fixes a bug when calling `RadarYMDHMSGetSite` or `RadarEpochGetSite` via the IDL DLMs which prevents loading the new hardware file format information.  The solution was to modify the two appropriate functions in `rposdlm.c` to match the logic in the native C `RadarEpochGetSite` function in `radar.c`, eg

https://github.com/SuperDARN/rst/blob/develop/codebase/superdarn/src.lib/tk/radar.1.22/src/radar.c#L36-L50


ecbland:
This fixes the bug...merging it in!

PR #373, opened 2021-01-11
Fix/make grid

mts299:
# Scope 

Fixing `make_grid` problem with `-c -st` combination highlighted in #370 

## Changes 

`make_grid.c` and `make_grid.md` was changed. 
 
- [x] added  errno start of #314 
- [ ] removed `-c`
- [x] got `-st` to work for multiple files 

Note: I removed `-c` because:
- Reduce code in `make_grid`
- Reduced copied code in `make_grid` which was probably why this bug happened in the first place
- keep consistent with `combine_grd` didn't use `-c`
- Less user mistake by forgetting the `-c` option

## Test Reading In Multiple Files with a Start Time

Pull and compile the branch `FIX/make_grid`

`make_grid -tl 60 -st 04:06 -xtd 20100101.*.tig.fitacf > tmp.grd`

**Output**

```
Opening file:20100101.0000.00.tig.fitacf
File does not contain the requested start time, ignoring this file.
Opening file:20100101.0200.00.tig.fitacf
File does not contain the requested start time, ignoring this file.
Opening file:20100101.0400.00.tig.fitacf
Opening file:20100101.0600.00.tig.fitacf
Opening file:20100101.0800.00.tig.fitacf
Opening file:20100101.1000.00.tig.fitacf
Opening file:20100101.1200.00.tig.fitacf
Opening file:20100101.1400.00.tig.fitacf
Opening file:20100101.1600.00.tig.fitacf
Opening file:20100101.1800.00.tig.fitacf
Opening file:20100101.2000.00.tig.fitacf
Opening file:20100101.2200.00.tig.fitacf
```

`dmapdump tmp.grd | less`

```  
      short   "start.hour" = 4
      short   "start.minute" = 6
```

## Test File not Found

```
make_grid -tl 60 -st 04:05 -xtd 20100101.*.dog.fitacf > tmp.grd
Opening file:20100101.*.dog.fitacf
Error: No such file or directory
```

PR #483, opened 2022-03-15
Update make_grid tutorial on readthedocs

ecbland:
This PR updates the readthedocs tutorial for making grid files. 

The documentation was written before we removed the `-new` and `-c` (concatenate) command line options from `make_grid`, so that's the motivation for the update.

### DIFF
--- a/docs/user_guide/make_grid.md
+++ b/docs/user_guide/make_grid.md
@@ -3,7 +3,7 @@
 author: Emma Bland, UNIS
 -->
 # FITACF to GRD
-Grid files contain combined data from multiple SuperDARN radars which have been organized into a latitude/longitude grid. Grid files are created in a two-step process:
+Grid files contain combined data that have been organized into a latitude/longitude grid. They may contain data from a single radar or from multiple radars. Multi-radar grid files are created in a two-step process:
 
 1. Create separate grid files for each radar
 2. Combine the separate grid files into a single file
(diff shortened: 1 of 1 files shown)


docs/user_guide/make_grid.md
```diff
@@ -17,13 +17,14 @@ combine_grid -vb *.grd > [outputfile].grd
...
 	
-## Full-day grid files
-The standard practice is to generate 24-hour grid files. If you have already made a concatenated 24-hour fitacf file, then use that file in the manner shown above. Otherwise, you can use the `-c` flag to get RST to concatenate the input fitacf files on-the-fly. This works only when the input files are all from the same radar; the method for combining data from multiple radars into a single grid file is shown in the next section.
+	
+### Full-day grid files
+The standard practice is to generate 24-hour grid files. If you have already made a concatenated 24-hour fitacf file, then use that file in the manner shown above. Alternatively, you can provide several fitacf files as input to `make_grid`, and the routine will automatically concatenate the files for you. This works only when the input files are all from the same radar; the method for combining data from multiple radars into a single grid file is shown in the next section.
```
mts299, line 22: Didn't we find a bug with 24-hour fitacf file in make_grid where it didn't read in the record right if it jumped too much.

ecbland: @mts299 I am not aware of this bug. Do you have an example?

mts299: Actually you were the one that found it way back when... I can try but it may take me a day or two.

ecbland: @mts299 This PR is about updating the documentation to reflect the changes made to the `make_grid` command line options in #373. I suggest we open a new issue if there's a bug in `make_grid`

mts299: Correct, I am just saying we may want to take the line out about using 24-hour fitacf files.

PR #488, opened 2022-03-21
Update `radar.dat`

ecbland:
This PR updates the `radar.dat` file with the end dates for several radars that have been decommissioned. 

@egthomas I guess we should add a new line to the LYR hardware file since it has not been operational since October 2018? 
https://github.com/SuperDARN/rst/blob/6ec842a19c51abea7e32476db279898fc7ce6c70/tables/superdarn/hdw/hdw.dat.lyr#L15

I noticed that the Halley hardware file, for example, has a line that reflects a temporary decommissioning: https://github.com/SuperDARN/rst/blob/6ec842a19c51abea7e32476db279898fc7ce6c70/tables/superdarn/hdw/hdw.dat.hal#L13-L15


egthomas:
@ecbland thanks for making these edits.  

> @egthomas I guess we should add a new line to the LYR hardware file since it has not been operational since October 2018?

I think there was a general consensus from the PIs that they didn't want to track every down-time for a given radar, and only wanted to set the `-1` flag once the site was permanently decommissioned (or a major status change such as hardware removal, eg below for Halley).  So I don't think we need to change the status for LYR if that radar is being rebuilt.  And in hindsight, maybe we shouldn't have set the HAN flag to `-1`, although I think that radar's refurbishment status is slightly less certain.

> I noticed that the Halley hardware file, for example, has a line that reflects a temporary decommissioning:

We added multiple `-1` flags for Halley because the radar hardware was physically removed from the site for a few years when the base was being rebuilt, which we felt warranted a change in status.

tables/superdarn/radar.dat
```diff
@@ -32,44 +32,44 @@
...
-206 1 20101104 25000101 "Christmas Valley West"   "Dartmouth College"                   "hdw.dat.cvw" "cvw"
-207 1 20101104 25000101 "Christmas Valley East"   "Dartmouth College"                   "hdw.dat.cve" "cve"
-208 1 20120906 25000101 "Adak West"               "Penn State University"               "hdw.dat.adw" "adw"
-209 1 20120906 25000101 "Adak East"               "Penn State University"               "hdw.dat.ade" "ade"
-512 0 20110101 25000101 "EKB Radar"               "Russian Academy of Sciences"         "hdw.dat.ekb" "ekb"
+0    0 20080203 25000101 "Test Radar"              "United Nations of SuperDARN"         "hdw.dat.tst" "tst" "x"
+1    1 19831001 25000101 "Goose Bay"               "Virginia Tech"                       "hdw.dat.gbr" "gbr" "g"
+2   -1 19861001 19950930 "Schefferville"           "CNRS/LPC2E"                          "hdw.dat.sch" "sch" "s"
+3    1 19930901 25000101 "Kapuskasing"             "Virginia Tech"                       "hdw.dat.kap" "kap" "k"
+4   -1 19880101 20160119 "Halley"                  "British Antarctic Survey"            "hdw.dat.hal" "hal" "h"
```
egthomas, line 39: Should this be `20160120` to still allow plotting of the FOV on `20160119`?  I'm not actually sure of the software behavior if the `hdw.dat` entry extends a few hours past the `radar.dat` end date.

ecbland: Good catch. I did a quick test with `fov_plot`. It works with `-d 20160119 -t 00:00` but not `-d 20160119 -t 00:01`.

Update on the way...

ecbland:
> @egthomas  if Dag wants to change LYR's status for that ~2018-present interval, that is of course his decision as PI

Ok, I'll follow that up with the PI another time. 

I guess this is ready to merge?

egthomas:
@ecbland looks good, thanks!

Issue #490, opened 2022-03-23
Enabling separate tdiff values for non-stereo two-channel data


pasha-ponomarenko:
# Discussion

**Topic:**
Enabling separate tdiff values for non-stereo two-channel data
## 

The new hardware file format intorduced in https://github.com/SuperDARN/rst/pull/455 contains two different values of hardware time delay, _tdiff,_ which allows to calibrate two different channels of the stereo mode data within a single run of the fitting code, _make_fit._ This feature represent a significant improvement as compared to the original version of the hardware files. 
It relies on the fact that the stereo mode data are recorded into two separate channels, 1 and 2, while the "mono" data are generally recorded into channel 0. The current implementation utilises the _offset_ field in the parameter structure to distinguish between the stereo and "mono" modes so that all non-stereo data utilise only a single value of _tdiff_ regardless of the channel number. This pre-selection is applied to both FITACF2.5  https://github.com/SuperDARN/rst/blob/db5204d4f049e2e84817b6f89752fea0ad79c659/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/fitacftoplevel.c#L174 
and FITACF3.0
https://github.com/SuperDARN/rst/blob/db5204d4f049e2e84817b6f89752fea0ad79c659/codebase/superdarn/src.lib/tk/fitacf.2.5/src/fitacf.c#L69
While the mono data are usually recorded with zero offset and channel = 0, **there is a significant volume of non-stereo data recorded with zero offset but split into channels 1 and 2** in order to distinguish between two radar frequency bands. These are two-frequency data from SAS, PGR, INV, RKN and CLY from late 2016 until the radar hardware was replaced by the new BOREALIS system during 2019-2021. 

**I suggest to extend the benefits of the two-channel calibration to these and any other data recorded into channels 1 and 2 by removing the zero offset condition and retaining data pre-selection by the channels only.**

Implementing these changes requires very simple modifications of the code, and their testing is straightforward. Suitable test data would be from CLY in 2017 when there were two distinct _tdiff_ values at two channels:

- channel 1 (10 MHz): -13.5 ns
- channel 2 (12 MHz): -17.0 ns

egthomas:
@pasha-ponomarenko thank you for elaborating on your proposed changes.  I can see the value added by enabling multiple tdiff values to support the dual-frequency operation, however I do have a few questions and concerns (which are my own thoughts and not on the behalf of the other Tdiff task force members):

[...]

PR #487, opened 2022-03-20
Release: RST 4.7

ecbland:
# Release RST 4.7

- [ ] PATCH
- [X] MINOR
- [ ] MAJOR

## Deadline: 11 April
Please comment in the Pull Request conversation if you need additional time to test the release.

## Testing

- [ ] Mac OSx
- [ ] Linux deb - Debian, Ubuntu, Mint
- [ ] Linux rpm - OpenSuse, Fedora, CentOS

*Do any binaries/libraries require special attention during testing?*

Key updates in this release are
- New hardware file format (#455)
- Find frequency bands (#444)
- Data simulator updates (#436)
- Bug fixes for compilation on various systems
    - #450
    - #459
    - #462
- Anything else?

## Checklist

- [ ] Update `radar.dat` (if required)
- [x] Update hardware files (if required)
- [ ] Update `.rst.version`
- [ ] Update author list in `.zenodo.json`

## Approving the release

After testing, please state which operating system you used and what you tested.

If you found no bugs during testing, please *approve* the PR:

1. Click on `files changed` top right
2. Click on `start review` green top right  button
3. In the comment box, state which operating system you used for testing, and any other helpful information
4. Select `approve` and `submit review`

PR #496, opened 2022-03-31
Tdiff for twof channels

pasha-ponomarenko:
This pull request implements changes discussed in https://github.com/SuperDARN/rst/issues/490.

Testing consists of making sure that different tdiffs are applied to different channels in both stereo and USask twofrequency mode data (CLY, SAS, PGR, INV, RKN starting from late 2016, cp_id=3501 or similar). The most straightforward way to test is to compare elevation data for four combinations of tdiff values:
- both tdiff[0] and tdiff[1] are zero -- reference set
- tdiff[0] is non-zero and tdiff[1] is zero -- ch1 changes, ch2  does not change
- tdiff[0] is zero and tdiff[1] is non-zero -- ch1 does not change, ch2  changes
- both tdiff[0] and tdiff[1] are non-zero -- both ch1 and ch2 change

I recommend to do this testing for both FITACF2.5 and FITACF3.0.

Please remember that in the new hardware format the date is given for the first occurrence of the respective set of parameters, while in the old format it is provided for the last occurrence, so one needs to make sure that the correct record in the HDW file is modified.

I used following commands to process and plot data from `20170319.1601.00.cly.rawacf`:
- Processing rawacf files:
FITACF2.5
`make_fit -fitacf-version 2.5 20170319.1601.00.cly.rawacf  > probe_v25_cly_11.fitacf`
FITACF3.0
`make_fit -fitacf-version 3.0 20170319.1601.00.cly.rawacf  > probe_v30_cly_11.fitacf`
- Plotting elevation data (beam 8):
channel 1;
`time_plot -png -e -b 8 -c A -st 16:00 -ex 02:00 -emax 45 probe_v30_cly_11.fitacf > probe_v30_cly_11_ch1.png`
channel 2:
`time_plot -png -e -b 8 -c B -st 16:00 -ex 02:00 -emax 45 probe_v30_cly_11.fitacf > probe_v30_cly_11_ch1.png`

For example, the reference elevation data (tdiff[0]=0, tdiff[1]=0) for FITACF2.5 in the above data looks like this:

- channel 1

probe_v25_cly_00_ch1

- channel 2

probe_v25_cly_00_ch2

Then I used tdiff[0]=-0.014 and tdiff[1]=0 and obtained the following graphs:

- channel 1 (changed as compared to the reference set above)

probe_v25_cly_10_ch1

- channel 2 (unchanged)

probe_v25_cly_10_ch2

### DIFF
--- a/codebase/superdarn/src.lib/tk/fitacf.2.5/src/fitacf.c
+++ b/codebase/superdarn/src.lib/tk/fitacf.2.5/src/fitacf.c
@@ -99,7 +99,7 @@ int fill_fit_block(struct RadarParm *prm, struct RawData *raw,
     input->prm.cp=prm->cp;
     input->prm.channel=prm->channel;
     input->prm.offset=prm->offset;  /* stereo offset */
-    if ((input->prm.offset == 0) || (input->prm.channel < 2)) {
+    if (input->prm.channel < 2) {
       input->prm.tdiff=hd->tdiff[0];
     } else {
       input->prm.tdiff=hd->tdiff[1];
--- a/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/fitacftoplevel.c
+++ b/codebase/superdarn/src.lib/tk/fitacf_v3.0/src/fitacftoplevel.c
@@ -171,7 +171,7 @@ void Copy_Fitting_Prms(struct RadarSite *radar_site, struct RadarParm *radar_prm
   fit_prms->bmoff=radar_site->bmoff;
   fit_prms->bmsep=radar_site->bmsep;
   fit_prms->phidiff=radar_site->phidiff;
-  if ((radar_prms->offset == 0) || (radar_prms->channel < 2)) {
+  if (radar_prms->channel < 2) {
     fit_prms->tdiff=radar_site->tdiff[0];
   } else {
     fit_prms->tdiff=radar_site->tdiff[1];


egthomas:
As previously mentioned in #487, the Tdiff Task Force is currently working on a more flexible and comprehensive `tdiff` implementation to support all radar operating modes, frequencies, channels, calibration techniques, etc.  Unfortunately this will not be ready in time for the upcoming 4.7 release, so we kindly ask to hold off on making any changes to the interpretation of the new `hdw.dat` file format until then.

pasha-ponomarenko:
@egthomas , I am not sure if I understand your problem with interpreting the HDW format as two-channel non-stereo data do exist so one cannot misinterpret this fact. Furthermore, the proposed changes do not affect in any way the stereo mode processing and, as far as I can see from our exchange in https://github.com/SuperDARN/rst/issues/490, have no ill effects on non-stereo data.

As you seem to talk on behalf of the task force (of which I am still a member I believe), has this particular issue -- usage of two tdiff values with non-stereo multi-channel data -- been discussed among the task force members? Are there any documents/notes/minutes reflecting this?

codebase/superdarn/src.lib/tk/fitacf.2.5/src/fitacf.c
```diff
@@ -99,7 +99,7 @@ int fill_fit_block(struct RadarParm *prm, struct RawData *raw,
     input->prm.cp=prm->cp;
     input->prm.channel=prm->channel;
     input->prm.offset=prm->offset;  /* stereo offset */
-    if ((input->prm.offset == 0) || (input->prm.channel < 2)) {
```
mts299, line 102: Can you comment on why the removal of offset == 0? Sorry I don't understand this much to begin with :p

mts299:
I am little confused here @pasha-ponomarenko by noticed the channel = 2 how does it select the frequency tdiff vs. what is in the hardware files which I thought was based on stereo mode? Sorry if this is a dumb question :p 

I will test to make sure it runs and give you some plots to confirm

pasha-ponomarenko:
Thanks @mts299! 
As you know, the new hardware format contains two tdiff values, tdiff[0] and tdiff[1] which are currently applied to channels 1 and 2 of the stereo mode data. For USask two-frequency data we also use channels 1 and 2 to distinguish between two frequency bands so it makes lots of sense to apply different tdiff values to those in the same way as done with stereo data.  As far as I am aware all other "mono" radars use channel 0. 
For stereo radars `offset` (shift between pulse sequences in different channels) is non-zero while for mono radars it is always zero. 
Under the current implementation the condition 
`(input->prm.offset == 0) || ` 
means that tdiff[0] is applied to all non-stereo data regardless of the channel number. By removing this condition we allow to utilise both tdiff[0] and tdiff[1] for calibrating channels 1 and 2 in USask two-frequency data.

mts299:
@pasha-ponomarenko I assume our hardware files already contain the two tdiffs for twofrequency mode?

pasha-ponomarenko:
@mts299 , no, right now these are all zeroes
https://github.com/SuperDARN/rst/blob/tdiff_for_twof_channels/tables/superdarn/hdw/hdw.dat.cly
```
#  Hardware parameters for Clyde River radar.
#
# **********************************************************************
# ==Notes==
# 13/Jan/2020 - Date that Borealis installed in Clyde
# 01/Feb/2022 - When Borealis changed to 3.24 degree beam separation
#
# **********************************************************************
#
#                            GEOGRAPHIC             BORE-   OFF  BEAM V  |--------- interferometer ---------| |--- RX ---| RNG MX
#      YYYYMMDD HH:MM:SS    LAT      LON      ALT   SIGHT   SET  SEP  SG PS  TDIFF  TDIFF    X      Y     Z    RXRIS  ATTN     BM
 66  1 20120806 00:00:00   70.487  -68.504    28.0  -55.62 0.00  3.24  1  1  0.000  0.000    0.0  100.0   0.0    0.0  10 0 100 16
 66  1 20200113 00:00:00   70.487  -68.504    28.0  -55.62 0.00  3.50  1  1  0.000  0.000    0.0  100.0   0.0    0.0  10 0 100 16
 66  1 20220201 18:00:00   70.487  -68.504    28.0  -55.62 0.00  3.24  1  1  0.000  0.000    0.0  100.0   0.0    0.0  10 0 100 16
# EOF
```
so for testing one would need to change them manually. 
For the example above (as for any CLY data form 2017) the actual calibration values are close to -0.014 and -0.018 for the first and second tdiff, respectively. For example, for the year of 2017 the changes should be made in the first record as it covers a period between 2012/08/06 and 2020/01/13.

mts299:
Sounds good make sure to make those changes. One last thing, do any other radars run twofsound especially stereo radars? I can see where that could cause some cobfusion

pasha-ponomarenko:
@mts299 , as far as I am aware, no stereo radars were running our twofsound mode, also some of those were using different frequencies.

The only other radar on which twofsound (cp_id=3501) was run is DCE before January 2017, but all these data were recorded as channel 0, so there would be no side effects or confusion.

mts299:
@pasha-ponomarenko Doing some testing and I am very confused still on this minor "or" change. From my understanding if you keep the `offset==0` doesn't change how twofsound is evaluated because you have an "or" the `channel < 2` will still be triggered? 

So the one radar I could find with 2 tdiff values and channels with various offsets in their is **han**  20170303.1800.01.han.fitacf 

Could not find where it had a `0` or `>= 2` channel or `offset == 0`  

Since this file has a `offset=400` it triggers the `else` but with your change this would not happen instead it would trigger `if` because the `channel=1` which is `< 2` so you would be using the wrong `tdiff` value for this radar. 

I can do more testing but I am seeing a issue with this PR. I would suggest closing it for now and maybe work witht he TDIFF task force some more how to differ stereo tdiffs and frequency tdiff

*Suggestion* Maybe an option to put in your own frequency based tdiff or even a file that gives the various channel based frequency tdiffs :)

pasha-ponomarenko:
In view of quickly approaching release date, I guess this issue can be put on the backburner.

egthomas:
@pasha-ponomarenko thank you for your contributions and bringing attention to this multi-channel / multi-frequency tdiff issue.  Given the inclusion of the new software and calibration file support in the upcoming 5.0 release (which can now accommodate multi-frequency data from individual radars), would it be appropriate to close this pull request?

pasha-ponomarenko:
@egthomas , sure.

PR #498, opened 2022-04-20
Merging `master` back into `develop`

ecbland:
Thanks everyone for your contributions to RST4.7 :100:  

This PR merges the `master` branch back into `develop`.

Once this PR is merged, I suggest that we take the opportunity to rename `master` to `main` (following guidelines from github and elsewhere). Please comment here if you have any concerns about renaming our default branch.


ksterne:
Think the renaming is an excellent idea!

egthomas:
> Once this PR is merged, I suggest that we take the opportunity to rename master to main (following guidelines from github and elsewhere). Please comment here if you have any concerns about renaming our default branch.

I just checked Zenodo and it shouldn't be any problem on that end (ie everything seems to be tied to releases and not branches).

ecbland:
Ok, I have renamed the default branch to `main`. You may need to update your local environment.

pasha-ponomarenko:
@ecbland,  I guess we need to inform the other users/developers about the name change as they may be using outdated repositories. 
I have just heard about this from Carley and Kevin K...

ecbland:
@pasha-ponomarenko The name change should really only affect current RST developers, so I'm reluctant to announce this more widely on the user mailing list. Users should be installing the citable releases published on Zenodo for official work (or equivalently, the tagged versions on Github, e.g. `v4.7`). The `master`/`main` (and `develop`) branches often differ from the official releases and therefore there are consequences for reproducibility.

PR #499, opened 2022-04-26
Updating html documentation for 4.7 release

egthomas:
This pull request is to update the API documentation on the `gh-pages` branch (ie https://superdarn.github.io/rst/index.html).

There appear to be a lot of changes mostly because there was a find-and-replace applied to the license information in the example C code.  Note that this worked correctly for some files, but not others (eg some are missing a closing `*/` at the end of the new license information, like `FrameBufferBezier.c`).


ecbland:
Thanks for doing this, @egthomas!

PR #500, opened 2022-05-18
Fixing bug in dattodmap.c causing usr_resL1 to be stored in usr_resL2 (and vice versa)

egthomas:
As indicated by the title, this pull request fixes a bug in `datdump` where the value of `usr_resL1` was being stored in `usr_resL2`, and vice versa.

Note that sometime in the early 2000's, `usr_resL1` became the field in which `offset` information for Stereo radars was stored, eg

https://github.com/SuperDARN/rst/blob/main/codebase/superdarn/src.lib/tk/oldraw.1.16/src/rawreadcurrent.c#L221

Unfortunately the `channel` information for the very earliest Stereo data (eg `2001091721fa.dat`) does not appear to be properly captured by either `datdump` or `dattorawacf`, but I'm not sure how widespread that issue is and it can probably be left for another pull request.


ecbland:
Nice catch @egthomas! I did a quick test with the file `2005021200f.dat` and (unsurprisingly) this fixes the bug. Merging it in...

PR #511, opened 2022-07-01
AACGM_v2 lock update

egthomas:
This (draft) pull request fixes an issue I noticed several years ago after we implemented AACGM_v2 / MLT_v2 into the RST (and replaces #510). Specifically, I noticed a dramatic slowdown when using `map_addhmb` with the new -v2 coordinate transformations as compared to using the old AACGM coefficients (via the `-old_aacgm` flag).

The three new functions (`AACGM_v2_Lock`, `AACGM_v2_Unlock`, and `AACGM_v2_Locked`) allow users the ability to bypass certain date/time checks within `MLTConvert_v2` which were causing the performance decrease (and were typically unnecessary when working with <= 24 hours of data from a single file).

To test, one can use something like

```
time map_grd yyyymmdd.hemi.grdmap | map_addhmb > yyyymmdd.hemi.hmb.cnvmap
```

and compare the `time` output on this branch versus `develop`, and with or without including the `-old_aacgm` flag.  On this branch there should be a ~3x speed increase compared to `develop`, and the timing should be comparable to using the `-old_aacgm` flag.

### DIFF
--- a/codebase/analysis/src.lib/aacgm_v2/aacgm.1.0/include/aacgmlib_v2.h
+++ b/codebase/analysis/src.lib/aacgm_v2/aacgm.1.0/include/aacgmlib_v2.h
@@ -66,7 +66,7 @@ int AACGM_v2_GetDateTime(int *year, int *month, int *day,
 int AACGM_v2_SetNow(void);
 int AACGM_v2_Lock(void);
 int AACGM_v2_Unlock(void);
-int AACGM_v2_GetLock(void);
+int AACGM_v2_Locked(void);
 
 #endif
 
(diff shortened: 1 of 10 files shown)


egthomas:
Anyone had a chance to test this?  We were hoping the HMB crew would be pleasantly surprised by the speed increase.

ecbland:
@egthomas Sorry, it wasn't clear that this was ready for review. I've just tested it and I can confirm that there is a ~3x speed improvement on this branch as described, and the output files are identical to `develop`.

egthomas:
@ecbland yes, I think this is ready to merge.  If Simon wants to make any changes for the next version of the AACGM_v2 library then we can address those in a future pull request.

PR #512, opened 2022-07-06
Tdiff calibration file update

egthomas:
This is a draft pull request on behalf of the Tdiff Task Force with a preliminary implementation of "tdiff calibration files" and the C and IDL software needed to interface with them.  From a user perspective, the important new features are the following:

- `fitacf` files now contain an additional field which keeps track of the `tdiff` value applied to each record by `make_fit`
- `make_fit` has a new `-tdiff` option that allows the user to specify a fixed value of `tdiff` (in units of microseconds) to all of the records for a given input file or data stream
- `make_fit` has a new `-tdiff_method` option that supports reading from a `tdiff.dat` file located in the `tables/superdarn/tdiff` directory (and pointed to by `$SD_TDIFFPATH`) which contains calibrated tdiff values for different techniques (ie methods), frequency bands, time periods, and (stereo) radar channels.  So the user can specify a particular calibration method and the code will search for a value for the appropriate frequency, time, and channel; if none can be found, then it will default back to the `hdw.dat` file.

By default, `make_fit` will still use the `tdiff` value from a radar's `hdw.dat` file.  These changes should be compatible with both FITACF 2.5 and 3.0, although there are likely lots of conflicts with @ecbland's #504.  The new `tdiff` field added to the `fitacf`-format files will also likely require some coordination with `pydarn` / `pydarnio`.

@aburrell has provided an example `han.tdiff.dat` file for testing purposes only, so please do not treat the values as fully calibrated (ie some are made up).

----------

Examples for testing (added by @ecbland)
*Note the new syntax for `make_fit`*

```
# Use tdiff from hardware file
make_fit -fitacf3 20150308.1401.00.han.rawacf > default.fitacf3

# Set fixed tdiff value
make_fit -fitacf3 -tdiff 0 20150308.1401.00.han.rawacf > tdiff0.fitacf3

# Use tdiff values from calibration file located in $SD_TDIFFPATH
make_fit -fitacf3 -tdiff_method 4 20150308.1401.00.han.rawacf > method4.fitacf3
```

The output `fitacf` files can be read using `dmapdump` or IDL. 
PyDARNio has not yet been updated to read the new `tdiff` field (this will be done soon)

Issue #479, opened 2022-03-03
Capturing data processing metadata in dmap files


ecbland:
# Discussion

At the DAWG meeting on 1 March 2022, we briefly discussed improving the way that data processing options are recorded in `dmap` files.

## Category

- [ ] User-friendly
- [ ] Software Design
- [X] Data related
- [ ] Capabilities
- [ ] Clarity
- [X] Workflow

## Examples

1. `origin.command` records the command used to generate the output file. However, this overwrites the command used to generate the input file:

```
make_fit -fitacf-version 3.0 20180323.0801.00.lyr.rawacf > 20180323.0801.00.lyr.fitacf
string	"origin.command" = "make_fit -fitacf-version 3.0 20180323.0801.00.lyr.rawacf"

trim_fit -st 09:00 -ex 00:02 20180323.0801.00.lyr.fitacf3 > lyr.trim.fitacf3
string	"origin.command" = "trim_fit -st 09:00 -ex 00:02 20180323.0801.00.lyr.fitacf3"
```

2. If another fitting algorithm was used (`lmfit`, `fitex`), this information is not captured at all (except `origin.command`)

```
make_lmfit 20180323.0801.00.lyr.rawacf > lyr.lmfit
int	"fitacf.revision.major" = 0
int	"fitacf.revision.minor" = 0
```

3. `fitacf.revision.major` and `fitacf.revision.minor` are not included in grid/map files, so it is impossible for a user to determine which version of FitACF was used to produce these files. 

-------------

**Note:** I'm not planning to work on this right now; just opening the issue so that we capture the action item from the meeting. If someone is keen to take the lead with this, please "assign" yourself to this issue. We can discuss this at our next telecon meeting, and we can seek input from the PIs when we are ready.

egthomas:
> 3. `fitacf.revision.major` and `fitacf.revision.minor` are not included in grid/map files, so it is impossible for a user to determine which version of FitACF was used to produce these files.

We should also keep in mind that a grid file can contain information from at least 1 to any number of different radars and channels, each of which may have been processed using a different fitting algorithm.  So a new field could be added to the `GridSVec` structure (and output grid file) which retains whatever new fitting algorithm string (or field) is also added to the fitacf-format files.

egthomas:
@ecbland I think we've accomplished the goal of this pull request with regards to `fitacf`-format files, right?

[...]

PR #513, opened 2022-07-08
Fitting algorithm field

egthomas:
This pull request adds a new `algorithm` field to `fitacf`-format files, which uses @ecbland's new consolidated `make_fit` to store an appropriate string for each fitting algorithm and (partially) addresses #479.  Similar to #512, the addition of this new field will eventually need to be coordinated with `pydarn` and `pydarnio`.

Note the target of this pull request is the `feature/make_fit` branch rather than `develop`.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/field_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/field_plot.c
@@ -1196,8 +1196,12 @@ int main(int argc,char *argv[]) {
     stid=prm->stid;
     cptab[0]=prm->cp;
     cpnum=1;
-    sprintf(revtxt,"Revision:%d.%d",fit->revision.major,
-            fit->revision.minor);
+    if (fit->algorithm !=NULL) {
+      sprintf(revtxt,"%s",fit->algorithm);
+    } else {
+      sprintf(revtxt,"Revision:%d.%d",fit->revision.major,
+              fit->revision.minor);
+    }
   }
   if (cfitflg) {
     stid=cfit->stid;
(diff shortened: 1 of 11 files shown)


ecbland:
@egthomas I'm not able to compile this branch. Are there some commits you forgot to push?

The `;` is missing on this line...

https://github.com/SuperDARN/rst/blob/ca7d3c1a64dfa46c41e7448a482b793060c7dc6f/codebase/superdarn/src.lib/tk/idl/fitidl.1.5/src/fitidl.c#L115

And I got an error: 
```
In file included from fitidl.c:34:
fitidl.c: In function ‘IDLCopyFitDataFromIDL’:
fitidl.c:56:34: error: ‘struct FitIDLData’ has no member named ‘algorithm’
   56 |   if (strlen(IDL_STRING_STR(&ifit->algorithm)) !=0)
      |                                  ^~
```

egthomas:
@ecbland sorry about that - I was just in the process of fixing those errors when you made your comment.  It should compile properly now, and I just verified that the IDL DLM is working.  The last thing for me to test is the native IDL support for the new field.

egthomas:
Ok so I changed my mind - one final commit modifies `time_plot` and `field_plot` to display the fitting algorithm string (if available) instead of the major and minor revision numbers, eg

20200626 0001 00 cve

If the fitting algorithm is not available from the `fitacf` file, then the major and minor revision numbers are displayed as usual.

ecbland:
Thanks for doing this @egthomas! I've tested this with `dmapdump`, the IDL DLM code and the native IDL code. The changes to `time_plot` and `field_plot` also work as described.

PR #514, opened 2022-07-08
Adding the `lmfit2` library

ecbland:
This PR adds the [`lmfit2` library](https://github.com/asreimer/lmfit2) by @asreimer to the RST, as requested by several people during the DAWG meeting in June.

The routine can be called using the new (temporary) binary `make_lmfit2`:
```
make_lmfit2 -vb 20180308.0001.00.cly.rawacf > 20180308.0001.00.cly.lmfit2
```
After this PR is merged, I will update #504 to call this new library from `make_fit`, and then remove the `make_lmfit2` binary. I'm adding the separate binary for now to simplify the testing. 

**Note:** I had to rename several header files in `lmfit2` that had the same names as the `fitacf3.0` header files (but slightly different contents). This was necessary because all these header files are copied into `$RSTPATH/include/superdarn` during compilation, so the `fitacf3.0` header files were being overwritten and causing a segfault. It might be possible to combine these files to remove the duplicated code, but I would prefer to address that in a separate PR. I found some other duplicated code as well (e.g. `mpfit.c`). Any suggestions about this are most welcome, now or later.

20180308 cly lmfit2


egthomas:
@ecbland I think I was able to remove the duplicate `mpfit` code, although `make_lmfit2` has been working on a 2-hour `rawacf` file for about 20 minutes now and still not done, so I assume it's still working correctly.

egthomas:
@ecbland I did some minor reorganization of the `lmfit2` header files so that not all of them are copied into the main `include` directory, and removed `llist.h` which is already contained in `fitacf3`.  Everything still compiles and `make_lmfit2` still appears to function correctly.

That raises another point, where there are nearly identical `llist.h` files in both the `include` and `src` directories of `fitacf3` - these should be consolidated and the one in the `src` directory removed, but that can be done in a separate pull request.

egthomas:
@ecbland I'm ok with approving / merging this pull request, but maybe someone else should also test since I made some changes too.

Do you have a general sense of how long it can take to process a 2-hour `rawacf` file?  I think it was on the order of ~30 min the last time I tried (granted that was with ~100 range gates).

ecbland:
> Do you have a general sense of how long it can take to process a 2-hour `rawacf` file? I think it was on the order of ~30 min the last time I tried (granted that was with ~100 range gates).

@egthomas The files I tested (with `nrang`=75) took about 12.5min for 2 hours, and [@asreimer's notes](https://github.com/asreimer/lmfit2/blob/18ee3e58a70436001ecdd54cffa067bd8bfc54e2/README.md?plain=1#L30-L31) say 9.5min/hour.


> I'm ok with approving / merging this pull request, but maybe someone else should also test since I made some changes too.

Yes, this one was a collaborative effort! It would be great if someone else could do a quick test, and let us know how long it takes to process a 2-hour file...  @pasha-ponomarenko would you be willing to have a look?

pasha-ponomarenko:
Sure!

________________________________

pasha-ponomarenko:
@ecbland, I have run the script on my Linux box, and it took just under 10 minutes to process one RAWACF file, 20180404.1801.inv.rawacf. The output looks reasonable and compatible with that obtained from FITACF3 except LMFIT2 does not do any pre-selection. Also, it does not look like there any elevation data processing performed in LMFIT2. Is it supposed to be done there?

LMFIT2
image

FITACF3

image

ecbland:
Thanks for testing this, @pasha-ponomarenko! That's correct that `lmfit2` does not do any data pre-selection, since that task is left to the user. And there's no elevation angle determination at present...that was on the [TODO list for `lmfit2`](https://github.com/asreimer/lmfit2/blob/master/README.md#limitations-and-to-do), and I won't be addressing it in this PR.

If you're happy that this PR is working, then please feel free to merge! Then I can add this fitting algorithm to `make_fit` in #504

pasha-ponomarenko:
@ecbland, done!

pasha-ponomarenko:
@ecbland, while adding this option to `make_fit` I believe you need to add a warning about the processing time. 
I guess if XCF processing is added, it will double...

asreimer:
Thanks for the interest and for the work integrating this into RST. While I don't have the bandwidth these days to support code development, I thought it would be useful to point out a couple things that might be helpful:

1. Figure 4 here: https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2017RS006450 shows that there are several local minima in chi-squared when fitting for velocity, so what LMFIT2 is doing is fitting ~30 initial guesses and picking the fit with the global mimimum chi-squared value. That means we fit each ACF ~30 times. So for XCF fitting, where we can use the fitted velocity from the ACF, we only need to perform 1 fit. This means XCF fitting won't double the processing time.
2. Note that the main limitation of LMFIT2 is the 1970s implementation of the Levenburg-Marquardt algorithm, back in the day when computers could only do 1 thing at a time. LM is just linear algebra though, so it's totally possible to significantly speed things up, which is what is done in 3
3. https://github.com/SuperDARNCanada/placeholder_fitting_name/tree/initial_branch implements a modern and highly vectorized Levenburg-Marquardt algorithm written in python. It can be run using CPU or GPU, both with a significant speedup compared to this C version of LMFIT2.

ecbland:
Thanks for the extra information @asreimer! That's good to hear that adding the XCF fitting would not significantly increase the processing time, and that there are other possibilities to improve the speed as well. I'll add a note in the documentation about the python algorithm in case someone wants to use the faster version available in that repo :smile:

PR #515, opened 2022-07-09
Consolidating llist.h files in the fitacf3 library

egthomas:
As noted in #514, this pull request consolidates the `llist.h` files in the `include` and `src` directories of the `fitacf3` library.  The only real difference are a few comment lines - `fitacf3` should (hopefully) still function the same on this branch as on `develop`.


ecbland:
Thanks @egthomas :100: 

I've checked that this doesn't modify the output of `fitacf3.0`. Merging now!

PR #516, opened 2022-07-20
Modifying LMFIT2 include guards

egthomas:
This pull request modifies the include guards used by the header files in the `lmfit2` library so they are unique and no longer duplicates of those in the `fitacf3` library.  For testing, the output of `make_lmfit2` should (hopefully) be identical between this branch and `develop`.


ecbland:
@egthomas I've compared the output of `make_lmfit2` on this branch and develop, and the results are identical. Merging now!

PR #517, opened 2022-07-21
Modifying LMFIT2 to include support for bmoff parameter

egthomas:
Sorry for not including this in the previous pull request - this one final commit adds support for the recently introduced `bmoff` parameter in the `hdw.dat` files to the `lmfit2` library (it looks like @ecbland already included the relevant `tdiff` functionality).

### DIFF
--- a/codebase/superdarn/src.lib/tk/lmfit_v2.0/include/lmfit_structures.h
+++ b/codebase/superdarn/src.lib/tk/lmfit_v2.0/include/lmfit_structures.h
@@ -124,6 +124,7 @@ typedef struct fit_prms {
   double **acfd;
   double **xcfd;
   int maxbeam;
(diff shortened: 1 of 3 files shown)


ecbland:
@egthomas  Ahhh, yes, I updated `tdiff` but forgot about `bmoff`. I've tested that this PR doesn't change or break anything. `lmfit2` doesn't determine elevation angles, so there's no difference between the output files on this branch or `develop` even when `bmoff` is nonzero. 

I just noticed that the `find_elevation()` function in `lmfit_determinations.c` is a copy of the elevation code that was originally included in `fitacf_v3` before we modified it to call the `elevation_v2` library. There's no rush to fix that up because we don't have `phi0` anyway. I've added a comment in the code to highlight this to future developers.

PR #519, opened 2022-08-04
Adding beam number color to fitacfclientgui

egthomas:
This pull request makes a small change to `fitacfclientgui` such that the current beam number is highlighted in red so that it is easier to tell which row has been most recently updated, eg beam 5 in the screenshot below:

fitacfclientgui


ecbland:
This is a helpful addition to the gui and it works as described. 

I noticed that the current beam is highlighted only if I use the `-color` flag. Do you also want to indicate the current beam when `-color` is not supplied (e.g. with an asterisk)? I'm happy to merge this now if you'd rather keep things as they are.

egthomas:
@ecbland thanks for testing.  Adding another symbol like an asterisk for non-color displays is an interesting idea.  That would ultimately affect the column spacing, which I unfortunately don't have the time to experiment with right now.  So maybe that feature could be left for later until someone really wants it?

PR #521, opened 2022-08-18
More simulator updates

egthomas:
This pull request adds support for the new `bmoff` parameter in the hardware files to the radar simulator (`make_sim`).  This pull request also makes major changes to `sim_real`, which (I think) was originally intended to take the contents of a `fitacf`-format file as input and then run the data simulator at each range where scatter was actually observed to produce an output `rawacf` (or `iqdat`) file.  I don't know if the intention was to use user-defined irregularity characteristics (like in `make_sim`) or the actual measured velocity/width/power values, so I've gone with the latter.  `sim_real` also seems to have used a custom plain-text file as input, whereas in this pull request I've modified it to read an actual `fitacf`-format file.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/make_sim.1.0/doc/make_sim.doc.xml
+++ b/codebase/superdarn/src.bin/tk/tool/make_sim.1.0/doc/make_sim.doc.xml
@@ -19,7 +19,11 @@
 </option>
 <option><on>-oldscan</on><od>use the old 7-pulse sequence.</od>
 </option>
-<option><on>-xcf</on><od>calculate XCFs (using a virtual height model).</od>
+<option><on>-xcf</on><od>calculate interferometer samples and XCFs (default is to use a virtual height model).</od>
+</option>
+<option><on>-elv <ar>elv</ar></on><od>set the elevation angle to <ar>elv</ar> degrees (instead of using a virtual height model).</od>
+</option>
+<option><on>-vht <ar>vht</ar></on><od>set the virtual height to <ar>vht</ar> degrees (instead of using a virtual height model).</od>
 </option>
 <option><on>-freq <ar>f</ar></on><od>set the radar frequency to <ar>f</ar> MHz (default is 12 MHz).</od>
 </option>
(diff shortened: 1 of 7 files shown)


pasha-ponomarenko:
_I don't know if the intention was to use user-defined irregularity characteristics (like in make_sim) or the actual measured velocity/width/power values, so I've gone with the latter. sim_real also seems to have used a custom plain-text file as input, whereas in this pull request I've modified it to read an actual fitacf-format file._
@egthomas, I believe you are right as the very name of the routine, `sim_real`, suggest reproduction of real data.

egthomas:
> @egthomas, I believe you are right as the very name of the routine, `sim_real`, suggest reproduction of real data.

@pasha-ponomarenko thanks for the confirmation!  It's possible that I've botched some of the simulator options, but otherwise the `sim_real` routine should now be in a much more usable state than before.

egthomas:
@pasha-ponomarenko you might want to verify this - I've just added a `-xcf` option to `make_sim` which will calculate a phase difference between the main and interferometer antenna arrays to populate the `xcfd` arrays for subsequent XCF / elevation angle processing with `make_fit`.  The simulated elevation angles come from the [Thomas and Shepherd [2022]](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2022RS007429) Christmas Valley virtual height model for ionospheric scatter (really any virtual height or elevation angle could be specified), and the phase differences are calculated using Equation 14 of [Shepherd [2017]](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2017RS006348) with the interferometer offset and tdiff values from the appropriate `hdw` file.

Initial testing looks reasonable but I may have misinterpreted something about how the simulator works.

pasha-ponomarenko:
@egthomas , I would be interested indeed but am rather busy right now. :-(
Do you have any test plots to show?

egthomas:
@pasha-ponomarenko not a problem, there isn't any urgency with this.

Here's an example plot of the ACF and XCF and their phase for a range gate using these options:

```
make_sim -srng 20 -n_good 20 -decay -xcf -beam 7 -freq 10
```

rawacf_sim_test

egthomas:
Users can now set the elevation angle or virtual height directly via the `-elv` and `-vht` options to `make_sim` when calculating interferometer samples for XCFs (as an alternative to the virtual height model which is the default). Users can also manually specify tdiff instead of using the default value from the hdw file.

That's probably the last of the updates I'll be making to this pull request (pending any feedback).

ecbland:
@egthomas This compiles correctly and I can use the various options as expected. The new options are a great addition. Could you please add a brief description of your modifications to all the `*.c` files, and then I will merge it :smile:

egthomas:
@ecbland sure - let me know if I missed any.

Issue #525, opened 2022-09-06
IDL FitSeek DLM segmentation fault


egthomas:
I'm not sure if this is because I'm using a file with a non-standard control program, but I'm encountering a segmentation fault when trying to use `FitSeek` from within IDL via the DLM (`RawSeek` is working fine, and so are `FitOpen` / `FitRead` / `FitClose`).  Can anyone else replicate this?  This may be my first time trying to use `FitSeek` from within IDL.  Thanks!

pasha-ponomarenko:
@egthomas, I have never used it, sorry!

egthomas:
@pasha-ponomarenko no problem - it ended up being a simple fix!

PR #526, opened 2022-09-07
Fixing FitSeek DLM bug

egthomas:
This pull request fixes a typo in `fitdlm.c` causing `FitSeek` to throw a seg fault when trying to use the DLM with IDL (#525).  To test, try something like the following on this branch and on `develop`:

```
inp = FitOpen('your_filename_here.fitacf', /read)
s = FitSeek(inp, yr, mo, dy, hr, mt, sc)
ret = FitRead(inp, prm, fit)
```

As far as I can tell, this bug dates back to when the DLMs were first introduced in RST 3.2.


ecbland:
This fixes the bug! Merging now :)

PR #528, opened 2022-10-12
Fixing likely typo in dmap.c

egthomas:
This pull request fixes a likely typo in `dmap.c` where a break occurs before an error flag can be set, which is inconsistent with the surrounding code.


ecbland:
Agree this looks like a typo. Merging now :)

PR #530, opened 2022-10-13
Fix flatten and expand functions

egthomas:
This pull request reverts several changes previously made to the `IQ` / `Fit` / `Raw` / `RadarParm` `Flatten` and `Expand` functions which had attempted to address compilation warnings.  While these previous changes got rid of the compilation warnings, the functions no longer work properly - in some cases garbage values are passed back and forth, and in others they fail completely.  I've also gone ahead and added support for the new `algorithm` field to `FitFlatten` and `FitExpand` for completeness.

Note that these functions are (as far as I know) only used by control programs in the ROS to pass data to the various `write` functions, so the changes in this pull request are not so straightforward to test.  One could either take my word for it, write their own program to test, or I can provide some instructions offline for using the radar simulator.

### DIFF
--- a/codebase/superdarn/src.lib/tk/fit.1.35/src/fit.c
+++ b/codebase/superdarn/src.lib/tk/fit.1.35/src/fit.c
@@ -154,23 +154,23 @@ void *FitFlatten(struct FitData *ptr,int nrang,size_t *size) {
 
   r=(struct FitData *) buf;
 
-  memcpy((int*)buf,ptr,sizeof(struct FitData));
+  memcpy(buf,ptr,sizeof(struct FitData));
   p=sizeof(struct FitData);
 
   if (ptr->rng !=NULL) {
-    memcpy((int*)buf+p,ptr->rng,nrang*sizeof(struct FitRange));
+    memcpy(buf+p,ptr->rng,nrang*sizeof(struct FitRange));
     r->rng=(void *) p;
     p+=nrang*sizeof(struct FitRange);
   }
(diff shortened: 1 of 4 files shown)


ecbland:
> Note that these functions are (as far as I know) only used by control programs in the ROS to pass data to the various write functions

Out of curiosity, why do we have these functions in RST at all if they are only used by the ROS? Is this related to the radar simulator?

> the changes in this pull request are not so straightforward to test. One could either take my word for it...

I have checked that RST compiles correctly on the branch. Beyond that I'm happy to take your word for it :smile:

egthomas:
> Out of curiosity, why do we have these functions in RST at all if they are only used by the ROS? Is this related to the radar simulator?

I suppose as an alternative to this pull request these functions could be removed entirely.  I think we left them in the RST because they were located within libraries under the `tk` (and not `os`) path and just never noticed.

Sorry, I should clarify that by radar simulator I didn't mean AJ/Pasha's `make_sim` but rather a different `radar` binary (removed 5+ years ago) that simulated how a control program on the Linux side of the 3.x RST-ROS would talk to the QNX6 side and write to `iqdat`, `rawacf`, and `fitacf` files in the correct output directories (but with no actual data).  I'm not sure if the QNX4 ROS has a comparable routine.

ecbland:
Thanks for the explanation @egthomas. I'll merge this now, and if you someone else wants to remove the functions entirely then that can happen in a separate PR.

PR #531, opened 2022-10-18
fitacfclientgui update

egthomas:
This pull request makes a small change to `fitacfclientgui` to include `nave` and `xcf` for each record in the output display header, for better assessment of real-time data streams.


egthomas:
One last update - `fitacfclientgui` should now cleanly exit on any key press, whereas before it would keep running and remain attached to the real-time server (eventually causing a "Too many connections" error to occur if you keep calling it which prevented any other clients from attaching).

ecbland:
Thanks @egthomas, both changes work as described.

image

PR #533, opened 2022-10-28
Release RST 5.0

ecbland:
# Release RST 5.0

- [ ] PATCH
- [ ] MINOR
- [X] MAJOR

As announced in #529, the next version of the RST will be a major release. Key changes are: 

- All fitting algorithms are now called by `make_fit`, and there is no "default" fitting algorithm (#504)
- C & IDL software for TDIFF elevation angle calibration (#512)
- New fields in the `fitacf` file format to support the two points above (`algorithm`/`tdiff`), and the elevation angle error calculations in `fitacf3` (`elv_error`/`elv_fitted`)
- Added the `lmfit2` fitting algorithm by @asreimer (#514). *Note: this has a separate binary `make_lmfit2` and cannot be called by `make_fit`*
- Updates to the data simulator (#521)
- New `fitacfclientgui` binary by @egthomas (#503)
- New virtual height model by @egthomas (#471)
- Option to use spectral width to constrain the HMB (#509)
- Speed improvements to `map_grd` and `map_addhmb` (#511)

## Deadline: 30 November 2022
Please comment in the Pull Request conversation if you need additional time to test the release.

## Testing

- [x] Mac OSx
- [x] Linux deb - Debian, Ubuntu, Mint
- [x] Linux rpm - OpenSuse, Fedora, CentOS

*Do any binaries/libraries require special attention during testing?*

We have made major changes to the following binaries (please add the ones I have missed!) 

- [ ] `make_fit`
- [ ] `make_grid`
- [ ] map potential (`map_addhmb`)
- [ ] `sim_real` & `make_sim`
- [ ] `make_lmfit2` (new in RST 5.0)
- [ ] `fitacfclientgui` (new in RST 5.0)

The IDL code for reading `fitacf` files has also been changed, and there are many other changes related to the tdiff updates.

## Checklist

- [x] Update `radar.dat` (if required) **Check closer to the testing deadline**
- [x] Update hardware files (if required) **Check closer to the testing deadline**
- [X] Update `.rst.version`
- [x] Update author list in `.zenodo.json` **check again before merging**
- [x] Update author list in `AUTHORS.md`

## Extra Notes

A description of the release will appear soon. 

## Approving the release

After testing, please state which operating system you used and what you tested.

If you found no bugs during testing, please *approve* the PR.

PR #536, opened 2022-11-04
Fix `fitacfclientgui` compilation on macOS

ecbland:
@agrocott _https://github.com/SuperDARN/rst/issues/533#issuecomment-1303662216_

> ...does the script that produces the makefile not have access to the OSTYPE system variable, such that the makefile could be modified based on that?

Yes, I think this might work. I've modified the makefile to check `$OSTYPE` before adding `-ltinfo`. It works for me on linux. Would you mind testing on your mac?


agrocott (approved):
Yes, this worked.

PR #537, opened 2022-11-05
Fitacf3 variable documentation add

ksterne:
I originally made this branch as I thought there were a few more places that needed updates to the documentation for the new variables, but turns out this is all I could find (for now).  We will need to update the `gh-pages` branch separately from the release branch to add in this info and other parts to keep things consistent.

### DIFF
--- a/docs/references/general/fitacf.md
+++ b/docs/references/general/fitacf.md
@@ -29,7 +29,7 @@ Sometimes modes like `twofsound` will write data into a single file. In this cas
 RST includes several fitting algorithms which are used to determine the physical properties of the received backscatter (power, velocity, spectral width, elevation) from the mean autocorrelation and crosscorrelation function estimates. 
 RST includes the following fitting algorithms:
 
-- FITACF 2.5 (default)
+- FITACF 2.5 (optional)
 - FITACF 3.0 (optional)
 - [FITEX2](https://doi.org/10.1002/rds.20031) (optional)
 - [LMFIT](https://doi.org/10.1002/rds.20031) (optional)
--- a/docs/references/general/fitacf.md
+++ b/docs/references/general/fitacf.md
@@ -24,19 +24,6 @@ Each file contains data from a different frequency channel.
 !!! Note
 Sometimes modes like `twofsound` will write data into a single file. In this case the two frequencies are marked as two separate channels, denoted using the `channel` parameter. However, it is important to note that some SuperDARN radars have stereo capability (transmitting and receiving on 2 frequencies simultaneously), which was the original intended usage of the `channel` parameter.
 
-## Fitting Algorithms
-
-RST includes several fitting algorithms which are used to determine the physical properties of the received backscatter (power, velocity, spectral width, elevation) from the mean autocorrelation and crosscorrelation function estimates. 
-RST includes the following fitting algorithms:
-
-- FITACF 2.5 (optional)
-- FITACF 3.0 (optional)
-- [FITEX2](https://doi.org/10.1002/rds.20031) (optional)
-- [LMFIT](https://doi.org/10.1002/rds.20031) (optional)
-
-Other fitting algorithms that are not yet implemented in RST:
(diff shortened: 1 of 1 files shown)


docs/references/general/fitacf.md
```diff
@@ -29,7 +29,7 @@ Sometimes modes like `twofsound` will write data into a single file. In this cas
 RST includes several fitting algorithms which are used to determine the physical properties of the received backscatter (power, velocity, spectral width, elevation) from the mean autocorrelation and crosscorrelation function estimates. 
 RST includes the following fitting algorithms:
 
-- FITACF 2.5 (default)
+- FITACF 2.5 (optional)
```
ecbland, line 32: @ksterne I wonder if we can remove this entire section? This information is already available in more detail in the `make_fit` tutorial, and I think it belongs there. https://radar-software-toolkit-rst.readthedocs.io/en/release-rst-5.0/user_guide/make_fit/

ksterne: I've removed the hold `Fitting Algorithms` section now.  Is the `File Structure` a carryover from the RFCs?  I could see the intent of this section is to help defined some parameters and how the file is written, and read, from one record to the next.  It's true that it's not specific to fitacf files, I'm guessing its there since fitacf files are the general user level radar data file.  There's possibly another page as a tutorial or under SuperDARN data that could contain some of this more generic information about variables in the files.

ecbland:
@ksterne, based on your comment above, in d685390a54878303a0503e44de083a54a60d5599 I have consolidated the FITACF file documentation a bit. If people are happy with this, then I will go through the other file formats and update in a similar way (in a separate PR).

docs/references/general/fitacf.md
```diff
@@ -117,52 +114,27 @@ The arrays for the XCF fit will only be present in the record if an xcf is calcu
...
 | *phi0_e* | *radians* | *[num_pts]* | ***float*** | Phase determination error at lag zero of the ACF | 
 | *elv* | *degrees* | *[num_pts]* | ***float*** | Elevation angle estimate recommended for scientific use |
 | *elv_fitted* | *degrees* | *[num_pts]* | ***float*** | Alternative elevation angle determined from fitted phase (FitACF v3 only) |
 | *elv_error* | *degrees* | *[num_pts]* | ***float*** | Least-square elevation angle error (FitACF v3 only) |
-| *elv_low* | *degrees* | *[num_pts]* | ***float*** | Lowest estimate of elevation angle (FitACF v1-2 only) |
-| *elv_high* | *degrees* | *[num_pts]* | ***float*** | Highest estimate of elevation angle (FitACF v1-2 only) |
-| *x_sd_l* | **None** | *[num_pts]* | ***float*** | Standard deviation of lambda fit of XCF|
+| *elv_low* | *degrees* | *[num_pts]* | ***float*** | Lowest estimate of elevation angle (FitACF v2 only) |
+| *elv_high* | *degrees* | *[num_pts]* | ***float*** | Highest estimate of elevation angle (FitACF v2 only) |
+| *x_sd_l* | **None** | *[num_pts]* | ***float*** | Standard deviation of lambda fit of XCF |
```
egthomas, line 136: @ecbland are all of the `(FitACF v2 only)` notes accurate with regards to `fitacfex`, `fitacfex2`, `lmfit`, etc?  I see that some of those fitting libraries call the `FitSetXrng` function - will that populate the `x_` fields as well?

ecbland: @egthomas Thanks, you're correct. I could just remove these algorithm-specific notes "(FitACF vX only)" from the table entirely. The note in this section about the XCF data is probably sufficient. What do you think?

egthomas: @ecbland I think that sounds fair.  It would be easy enough to add it back in if the community expresses a desire for it.

ecbland:
I'd say this PR is ready to merge :smile:  Could someone please press the big green button?

PR #539, opened 2022-11-15
Modifying rtcfit, rtgrid, and rtsnd to sleep 5 seconds if a connection fails

egthomas:
As the title / commit suggest, this pull request modifies `rtcfit`, `rtgrid`, and `rtsnd` to add a `sleep(5)` command if the connection to a real-time data stream fails, times out, or is reset.  This should both limit the number of attempted connections if a stream goes down, and also reduce the size of the rt* logs.

For example, currently `rtgrid` will try to connect >9 times per second if the connection fails, leading to a log file of >100 MB per day.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/rtcfit.1.12/rtcfit.c
+++ b/codebase/superdarn/src.bin/tk/tool/rtcfit.1.12/rtcfit.c
@@ -342,20 +342,9 @@ int main(int argc,char *argv[]) {
     if (resetflg==1) loginfo(logfname,"Connection timed out.");
     if (resetflg==2) loginfo(logfname,"Connection reset by signal.");
     ConnexClose(sock);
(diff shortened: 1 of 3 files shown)


ecbland:
@egthomas There's no obvious way to test this, so should I just merge it in? 🚀

egthomas:
@ecbland you could try running rtgrid with an invalid host / port number, eg

`rtgrid localhost 1024` (without any active processes like `fitacfserver`)

and watch how quickly the log file fills up on this branch compared to `develop`.  Otherwise, you're right that there isn't an obvious way to test this.  We may be the only group currently running `rtgrid` on a regular basis (besides maybe USask?).

ecbland:
@egthomas Ok, I can see a difference in the log file size on this branch compared to `release/rst-5.0`. Merging now!

PR #540, opened 2022-11-15
Modifying fitacfclientgui to add elevation angle support and keyboard toggle

egthomas:
This pull request makes a few small changes to the new `fitacfclientgui` to add support for plotting elevation angles when in color mode, and also support for toggling between different parameters by pressing the `p` (power), `v` (velocity), `w` (width), or `e` (elevation) keys; pressing any other key will still cleanly exit the program.


egthomas:
Ok, I promise these are the last two edits to `fitacfclientgui` - the code now automatically draws the correct number of range gate labels based on the `nrange` option, and the default maximum value of the power color scale has been changed from 40 to 30 dB.

PR #543, opened 2022-11-18
Fix/rbposlib.pro

agrocott:
I was testing the IDL routines, and happened to test it on a file time 2002010100, i.e. yrs(year seconds)=0, upon which the code failed. It tests if the "yrs" keyword is set, using the "keyword_set" function. This is unadvisable ("keyword_set" is best used when the keyword is binary) since it evaluates FALSE if the keyword has a (valid) value of zero. In this case, the time being 0:0:0 on Jan 1st, yrs=0, and "keyword_set(yrs)" was FALSE. I have replaced "keyword_set" with "n_elements" which returns 1 (TRUE) for any value (including zero). It evaluates 0 if the variable is undefined (i.e. unset).


agrocott:
Requested into wrong branch

PR #544, opened 2022-11-18
Update rbposlib.pro

agrocott:
I was testing the IDL routines, and happened to test it on a file time 2002010100, i.e. yrs(year seconds)=0, upon which the code failed. It tests if the "yrs" keyword is set, using the "keyword_set" function. This is unadvisable ("keyword_set" is best used when the keyword is binary) since it evaluates FALSE if the keyword has a (valid) value of zero. In this case, the time being 0:0:0 on Jan 1st, yrs=0, and "keyword_set(yrs)" was FALSE. I have replaced "keyword_set" with "n_elements" which returns 1 (TRUE) for any value (including zero). It evaluates 0 if the variable is undefined (i.e. unset).

### DIFF
--- a/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro
+++ b/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro
@@ -121,7 +121,7 @@ function rbpos,range,height=height,beam=beam,lagfr=first_lag,smsep=smsp, $
   if (keyword_set(station)) then st_id = station else st_id = dp.p.st_id
 
   if (keyword_set(yr)) then year=yr else year=dp.p.year
-  if (keyword_set(yrs)) then yrsec=yrs else $
+  if (n_elements(yrs)) then yrsec=yrs else $
       yrsec=TimeYMDHMSToYrsec(dp.p.year,dp.p.month,dp.p.day,dp.p.hour, $
                                   dp.p.minut,dp.p.sec)
 


ecbland:
@agrocott Good catch! Could you provide some sample IDL code so that I can reproduce the bug?

agrocott:
> @agrocott Good catch! Could you provide some sample IDL code so that I can reproduce the bug?

Hi Emma,

The crux of it is this, when using rbpos, e.g.:

pos=rbpos(0,beam=0,lagfr=prm.lagfr,smsep=prm.smsep,rxrise=prm.rxrise,station=prm.stid,year=prm.time.yr,yrsec=yrsec)

where yrsec=TimeYMDHMSToYrsec(prm.time.yr,prm.time.mo,prm.time.dy,prm.time.hr,prm.time.mt,prm.time.sc)

then if yrsec = 0, rbpos takes yrsec to be undefined (according to keyword_set), instead of it being set to zero.

This causes rbpos to default to using an optional data structure, that I do not have defined. Perhaps I ought to have that defined, but equally, the code ought to function as intended without it.

egthomas:
@agrocott does the same issue occur for other keywords where `0` is a valid input (eg, beam number, receiver rise time, etc)?

ecbland:
@agrocott I was not able to reproduce this bug based on your description, but I do agree that `n_elements()` should be used instead of `keyword_set()`. 

This is what I did (without defining any data structures beforehand):

```
IDL> pos=rbpos(4,beam=0,lagfr=1200,smsep=300,rxrise=100,station=90,year=2018,yrsec=0)
IDL> print,pos
      77.5803      116.437      6472.66
      77.5309      116.966      6472.66

      77.9597      117.635      6472.66
      77.8979      118.254      6472.66
```
Have I misunderstood something here?



> Related: Should we be correcting "bad practice" in the code, or only actual bugs?

Yes, we should fix bad practice in the code :smile:

agrocott:
@ecbland

> Have I misunderstood something here?

More likely I have! I am running the RST IDL code from within IDL code of my own, and quite possibly I am doing something unexpected...

Here's what seems to be happening for me. In rbpos line 124:

`if (keyword_set(yrs)) then yrsec=yrs else $`
`  yrsec=TimeYMDHMSToYrsec(dp.p.year,dp.p.month,dp.p.day,dp.p.hour, $`
`  dp.p.minut,dp.p.sec)`

with yrs=0, then keyword_set(yrs) is false and so the "else" statement is executed. With the structure dp not populated, TimeYMDHMSToYrsec returns -1. This causes line 129 to fail: 

`IDL>  s=TimeYrsecToYMDHMS(yr,mo,dy,hr,mt,sc,yrsec)`

gives sc = -1 (as an aside I notice this line is repeated 2 lines down, so I will tidy that up).

This then causes line 133 to fail:

`site=RadarYMDHMSGetSite(rid,yr,mo,dy,hr,mt,sc)`

returns site = 0

with site being passed to the function RadarPos later on, and so on.

Is "dp" (or dp=dptr, dp=fit_data) definitely not defined and populated in your case? If it was then the above "else" statement would work as intended. "dp" is empty in my case. I am not sure where it is supposed to come from. When I read in a fitacf file (using FitRead in fit.pro) the "prm" and "fit" structures are populated, but nothing matching the structure of "dp". It's possible I'm missing an anticipated step due to using some of my own code...

Just to muddy the water even further, if I execute your example command (using station 90) it fails for me again, but for a different reason (I'm not having a lot of luck here!). I'll start a new comment about that...

egthomas:
I think this may be related to whether `global.pro` is included in the user's `startup.pro` - that legacy routine calls `fitdef` which populates the time values in the `dp` structure with zeroes.

ecbland:
> @agrocott Is "dp" (or dp=dptr, dp=fit_data) definitely not defined and populated in your case? If it was then the above "else" statement would work as intended. "dp" is empty in my case. I am not sure where it is supposed to come from. When I read in a fitacf file (using FitRead in fit.pro) the "prm" and "fit" structures are populated, but nothing matching the structure of "dp". It's possible I'm missing an anticipated step due to using some of my own code...

> @egthomas I think this may be related to whether global.pro is included in the user's startup.pro - that legacy routine calls fitdef which populates the time values in the dp structure with zeroes.

Yep, I don't understand how to define the `dp` structure after reading a fitacf file with `fitread(fp,prm,fit)`. I've just checked my startup file and it does compile `global.pro`, so that means the `dp` structure was already populated when I called `rbpos()`. That's probably why I can't reproduce the bug. 

If I remove `global.pro` from my startup file then the code does not work for me. Perhaps it would work if I knew how to create the `dp` structure...do you have an example?

```
% Execution halted at: RBPOS             114 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
%                      RBPOS             115 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
%                      RBPOS             114 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
% Variable is undefined: FIT_DATA (FITDATA_COM)
```

agrocott:
> > @agrocott Is "dp" (or dp=dptr, dp=fit_data) definitely not defined and populated in your case? If it was then the above "else" statement would work as intended. "dp" is empty in my case. I am not sure where it is supposed to come from. When I read in a fitacf file (using FitRead in fit.pro) the "prm" and "fit" structures are populated, but nothing matching the structure of "dp". It's possible I'm missing an anticipated step due to using some of my own code...
> 
> > @egthomas I think this may be related to whether global.pro is included in the user's startup.pro - that legacy routine calls fitdef which populates the time values in the dp structure with zeroes.
> 
> Yep, I don't understand how to define the `dp` structure after reading a fitacf file with `fitread(fp,prm,fit)`. I've just checked my startup file and it does compile `global.pro`, so that means the `dp` structure was already populated when I called `rbpos()`. That's probably why I can't reproduce the bug.
> 
> If I remove `global.pro` from my startup file then the code does not work for me. Perhaps it would work if I knew how to create the `dp` structure...do you have an example?
> 
> ```
> % Execution halted at: RBPOS             114 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
> %                      RBPOS             115 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
> %                      RBPOS             114 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
> % Variable is undefined: FIT_DATA (FITDATA_COM)
> ```

I'm getting a bit (more) confused now. If you remove global.pro then "fit_data" does not get created, and so it makes sense that rbpos fails when it tries to assign "dp = fit_data" on line 114, as above.

But with global.pro compiled, and fit_data initiated (but left empty), then "dp = fit_data" assigns an empty structure to "dp" (which is what it does for me) - it should then fall over later, when it tries to use the empty structure to create the "yrsec" variable. But It doesn't fail for you, which is what I don't understand, if you are not populating the "fit_data" structure...?

What happens if you read in a fitacf file, and then type "help, fit_data" at the command line? Do you see an empty structure or a populated one?

ecbland:
> @agrocott What happens if you read in a fitacf file, and then type "help, fit_data" at the command line? Do you see an empty structure or a populated one?

I get an empty structure:

```
IDL> fp=fitopen('20150803.0601.00.mcm.a.fitacf3')
IDL> s=fitread(fp,prm,fit)
IDL> help,/st,fit_data
** Structure FIT, 50 tags, length=20696, data length=20694:
   REC_TIME        LONG                 0
   P               STRUCT    -> PARMS Array[1]
   PULSE_PATTERN   INT       Array[16]
   LAG_TABLE       INT       Array[2, 48]
   COMBF           BYTE      Array[80]
   PAD1            LONG                 0
   NOISE_LEV       DOUBLE           0.0000000
   NOISE_LAG0      DOUBLE           0.0000000
   NOISE_VEL       DOUBLE           0.0000000
   PWR_LAG0        DOUBLE    Array[75]
   SLIST           INT       Array[75]
   NSEL            INT              0
   QFLAG           LONG      Array[75]

```


With `fit_data` defined but empty, I tried to use it in `rbpos()`:

```
IDL> pos=rbpos(10,data=fit_data)
% Expression must be a structure in this context: SITE.
% Execution halted at: RBPOS             165 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
%                      RBPOS             165 /rst/idl/lib/superdarn/legacy.1.6/rbposlib.pro
%                      $MAIN$              1 /rst/idl/lib/superdarn/legacy.1.6/global.pro

```

If I also supply the `year` keyword, the code does not crash, but of course the output is wrong because the critical information required to calculate the position is all set to zero. This includes the station ID, (so it defaults to `gbr`), and the output will be the same no matter which value of `range` is supplied:

```
IDL> print,rbpos(10,data=fit_data,year=2015)
      61.4621      22.5867      6364.42
      61.4621      22.5868      6364.42

      61.4621      22.5867      6364.42
      61.4621      22.5868      6364.42

IDL> print,rbpos(50,data=fit_data,year=2015)
      61.4621      22.5867      6364.42
      61.4621      22.5868      6364.42

      61.4621      22.5867      6364.42
      61.4621      22.5868      6364.42
```

At this point, I'd say that the `year` keyword is the critical one, since that's all that's needed to read the a hardware file (which will be `gbr` by default). Provided that `fit_data` exists, this doesn't crash:

```
IDL> print,rbpos(50,year=2015)
      61.4621      22.5867      6364.42
      61.4621      22.5868      6364.42

      61.4621      22.5867      6364.42
      61.4621      22.5868      6364.42
```

-----

So it seems there's a lot of potential for things to go wrong in this routine. It might be "legacy code" for a reason. 

To make progress from here, complete code examples that reproduce the bug would be appreciated (e.g. like the code above).

agrocott:
@ecbland

I wonder of we are using different code? Perhaps I have managed to get an older version or something... For a start, for me to read in a fitacf file, I must specify the "read" keyword:

`fp=fitopen('20010101.0000.00.sas.rst5.fitacf',/read)`

which is consistent with the version of fitopen I see in fit.pro:

```
function FitOpen,fname,read=read,write=write,update=update
  if KEYWORD_SET(read) then $
    openr,unit,fname,/GET_LUN,/SWAP_IF_BIG_ENDIAN
  if KEYWORD_SET(write) then $
     openw,unit,fname,/GET_LUN,/SWAP_IF_BIG_ENDIAN
  if KEYWORD_SET(update) then $
     openu,unit,fname,/GET_LUN,/SWAP_IF_BIG_ENDIAN
  return, unit
end

```
When I call `rbpos` with an empty `fit_data` passed through I get the following error:

```
IDL > pos=rbpos(10,data=fit_data)
IDL: Illegal variable attribute: MAXBEAM.
IDL: Execution halted at: RADARPOS          975 /users/grocott/Documents/GitHub/rst/idl/lib/superdarn/
     main.1.25/radar.pro
IDL:                      RBPOS             162 /users/grocott/Documents/GitHub/rst/idl/lib/superdarn/
     legacy.1.6/rbposlib.pro
IDL:                      $MAIN$
```

which is the same error that I get if I don't pass `fit_data`, because `rbpos` assigns fit_data anyway if it is not passed through:

`if (keyword_set(dptr)) then dp = dptr else dp = fit_data`

This results in `site` being equal to zero. Although why I don't get the `Expression must be a structure in this context: SITE.` error you get, I am not sure. Instead I get the above error when `RadarPos` tries to access `site.maxbeam`.

So, I am wondering how `rbpos` works for you without specifying fit_data, or any of the other parameters such as `lagfr`, `smsep` and `rxrise`?  If fit_data is empty, then where does it get those parameter values from?

Can you point me to the version of the code you are using so I can check whether I am somehow using the wrong version?

Thanks.

ecbland:
> I wonder of we are using different code?...For a start, for me to read in a fitacf file, I must specify the "read" keyword

So I figured out why we are getting different errors. By default my IDL setup uses the DLMs, and you are using the native IDL code. If I disable the DLMs then I get the same error you got above (`Illegal variable attribute: MAXBEAM.`). I'll have another look at this tomorrow.

egthomas:
I think the solution here is to add some kind of check for an error value (of 0) returned by `RadarYMDHMSGetSite`, eg

```
  if site eq 0 then begin
    print, 'Site not found'
    stop
  endif
```

otherwise a site of zero (rather than a structure) is passed to `RadarPos` which causes the `Illegal variable attribute: MAXBEAM.` error.

agrocott:
> > I wonder of we are using different code?...For a start, for me to read in a fitacf file, I must specify the "read" keyword
> 
> So I figured out why we are getting different errors. By default my IDL setup uses the DLMs, and you are using the native IDL code. If I disable the DLMs then I get the same error you got above (`Illegal variable attribute: MAXBEAM.`). I'll have another look at this tomorrow.

Ah... (numpty alert) I had forgotten about the DLMs.... I think I used to use them but they do not seem to be enabled (or present?) in my current installation.

So, this leads me to two questions:

1. Should I even be reporting bugs in the legacy code? Does it being legacy mean it might go out of date and will not be fixed?
2. What might I be doing wrong that means I'm not using the DLMs?  I don't seem to have the path `$RSTPATH/dml` (which is what `idl.bash` suggests I should have. Are there setup steps I might be missing?

Sorry for the confusion...

egthomas:
@agrocott to get the DLMs to compile properly, you need to make sure the `$IDL_IPATH` environment variable in `idl.bash` points to the correct directory where the `idl_export.h` file can be found, eg

https://github.com/SuperDARN/rst/blob/main/.profile/idl.bash#L6

Once you've done that and refreshed your session, you will need to execute `make.code` again and the script should untar the `idl` directory in `codebase/superdarn/src.lib/tk`, compile the necessary libraries, and then compile the DLMs.

pasha-ponomarenko:
@agrocott, did you look at https://github.com/SuperDARN/rst/blob/main/codebase/superdarn/src.idl/lib/main.1.25/radar.pro?
There is an IDL function RadarPos which seems to be a current re-incarnation of rbpos.

pasha-ponomarenko:
By the way, while talking about geolocation, does anyone know if there is some algorithm description besides the brief comments in the codes' headers? I am particularly interested in justification for the [elevation/height adjustment part.](https://github.com/SuperDARN/rst/blob/041093e4cdfd021ea5b5c93a7e0323fd8c867dab/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L636)

egthomas:
> @agrocott, did you look at https://github.com/SuperDARN/rst/blob/main/codebase/superdarn/src.idl/lib/main.1.25/radar.pro?
There is an IDL function RadarPos which seems to be a current re-incarnation of rbpos.

@pasha-ponomarenko `rbpos` is really just a wrapper to `RadarPos` so the user can get the centers or corners of all the beams and range gates for a radar at once.  When I was at VT someone (probably Lasse) re-wrote it to drop the dependencies on legacy data structures and common blocks.

> By the way, while talking about geolocation, does anyone know if there is some algorithm description besides the brief comments in the codes' headers? I am particularly interested in justification for the [elevation/height adjustment part.](https://github.com/SuperDARN/rst/blob/041093e4cdfd021ea5b5c93a7e0323fd8c867dab/codebase/superdarn/src.idl/lib/main.1.25/radar.pro#L636)

The `RPosGeo_v2` function I added is a C version of the IDL `RadarPos` function but re-written to drop that iterative oblate spheroid business and include better support for elevation angles, arbitrary numbers of hops, front vs rear FOV, and the different virtual height models.

pasha-ponomarenko:
Thanks, @egthomas, I will have a look at v2!

agrocott:
> @agrocott to get the DLMs to compile properly, you need to make sure the `$IDL_IPATH` environment variable in `idl.bash` points to the correct directory where the `idl_export.h` file can be found, eg
> 
> https://github.com/SuperDARN/rst/blob/main/.profile/idl.bash#L6
> 
> Once you've done that and refreshed your session, you will need to execute `make.code` again and the script should untar the `idl` directory in `codebase/superdarn/src.lib/tk`, compile the necessary libraries, and then compile the DLMs.

Thanks @egthomas that worked (eventually, at least. Some of the header files (such as `grdidl.h`) were missing from my working branch and I had to reinstate them from release/rst-5.0 - not sure how that happened but quite probably something I did!) Anyway, low and behold, I now get

```
IDL > pos=rbpos(40,beam=0,lagfr=1200,smsep=300,rxrise=100,station=90,year=2018)
IDL: Loaded DLM: RPOSDLM.
IDL: Loaded DLM: AACGMDLM.
```
So, unlike the native IDL version, this works without needing to define `yrsec` (as @ecbland noted above). Out of curiosity, then, what is `rposdlm` using for `yrsec` if it is not provided? Does it default to zero? If so, should I make the legacy idl code do the same so it functions in the same way (assuming we are going to update the legacy code)?

ecbland:
When I merged #546, this one automatically "merged" as well since both PRs included the same actual commit (https://github.com/SuperDARN/rst/commit/426f2c3ae8b97660a2bb9aee3db80824e041fa23)

@agrocott Do you think we should change all the `keyword_set()` instances to `n_elements()` in a new PR?

agrocott:
> When I merged #546, this one automatically "merged" as well since both PRs included the same actual commit ([426f2c3](https://github.com/SuperDARN/rst/commit/426f2c3ae8b97660a2bb9aee3db80824e041fa23))

Sorry, didn't realise it was the same commit, thought it was a new commit with some of the same changes in it (if you commit changes to a file, then make more changes and commit again, then create a PR from the new commit, does it automatically include previous commits of the same file?)
 
> @agrocott Do you think we should change all the `keyword_set()` instances to `n_elements()` in a new PR?

Yes, I suppose so since it is technically wrong. Although this doesn't solve the problem that if fit_data is not populated (or the required parameters are not passed) then the code either crashes or (worse) returns the wrong result. Do we want to add something to catch this?

We can either add a catch for `site=0` as @egthomas suggests, or prevent that from occurring by defaulting to `yrsec=0`, which is what `rposdlm` appears to do (unless it gets the correct yrsec from somewhere else, but I don't think it can since `rbpos` works even without loading data) so we'd have consistency in that case.

Should we also prevent the code from executing and returning an incorrect (or at least unexpected) result if the other parameters are not set?

PR #546, opened 2022-11-28
Fix/rbposlib (and RadarEpochGetSite)

agrocott:
1. Deleted duplicate call to TimeYrsecToYMDHMS in rbpos.
2. Fixed rbpos returning wrong position coords (due to error in tval loop in RadarEpochGetSite).

Before this PR is merged, we might want to understand why issue (2) seems to be user dependent... @ecbland does not seem to encounter this issue. I wonder if it's something to do with me using native IDL routines - maybe @ecbland is not?

The issue with the IDL version of RadarEpochGetSite, at least, seems to be that, for certain radars, rbpos is erroneously returning mag lat = -16.118233, mag lon = 72.000923 (which are basically the geomagnetic coordinates of geographic lat = 0, lon = 0). 

I have traced the issue to where the hardware file is read in and scanned for records within the current time:

`  for s=0,radar.snum do begin`
`    if  (radar.site[s].tval gt tval) then break`
`  endfor`
`return, radar.site[s-1]`

In, e.g. hdw.dat.lyr, there is a single entry, and so radar.snum=1. I think the problem is that the code is looping from zero to radar.snum, rather than radar.snum-1. Since radar.site[1].tval is unpopulated, it is zero, and so passes the check. Upon exiting the loop, "s" is incremented again, so then s=2. The final line returns the site info for s-1, i.e. radar.site[1], which is empty. So, the returned site lat and lon will be zero, which is what I am finding. Changing the loop to:

`for s=0,radar.snum-1 do begin `

seems to fix the problem.

### DIFF
--- a/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro
+++ b/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro
@@ -121,7 +121,7 @@ function rbpos,range,height=height,beam=beam,lagfr=first_lag,smsep=smsp, $
   if (keyword_set(station)) then st_id = station else st_id = dp.p.st_id
 
   if (keyword_set(yr)) then year=yr else year=dp.p.year
-  if (keyword_set(yrs)) then yrsec=yrs else $
+  if (n_elements(yrs)) then yrsec=yrs else $
       yrsec=TimeYMDHMSToYrsec(dp.p.year,dp.p.month,dp.p.day,dp.p.hour, $
                                   dp.p.minut,dp.p.sec)
 
--- a/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro
+++ b/codebase/superdarn/src.idl/lib/legacy.1.6/rbposlib.pro
@@ -126,8 +126,6 @@ function rbpos,range,height=height,beam=beam,lagfr=first_lag,smsep=smsp, $
                                   dp.p.minut,dp.p.sec)
 
   yr=year
-  s=TimeYrsecToYMDHMS(yr,mo,dy,hr,mt,sc,yrsec)
-
   rid=RadarGetRadar(network,st_id)
   s=TimeYrsecToYMDHMS(yr,mo,dy,hr,mt,sc,yrsec)
   site=RadarYMDHMSGetSite(rid,yr,mo,dy,hr,mt,sc)
--- a/codebase/superdarn/src.idl/lib/main.1.25/radar.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/radar.pro
@@ -430,7 +430,7 @@ function RadarEpochGetSite,radar,tval
   if (radar.st_time ne -1) && (tval lt radar.st_time) then return,0
   if (radar.ed_time ne -1) && (tval gt radar.ed_time) then return,0 
 
-  for s=0,radar.snum do begin
+  for s=0,radar.snum-1 do begin
     if  (radar.site[s].tval gt tval) then break
   endfor
 


egthomas:
On further examination I'm not sure there's a problem with `RadarEpochGetSite` and radars with only a single line in their hardware files, eg

```
IDL> rid = RadarGetRadar(network,90)
IDL> tval = TimeYMDHMSToEpoch(2016,10,20,0,0,0)  
IDL> site = RadarEpochGetSite(rid,tval)
IDL> help, site
** Structure SITE, 17 tags, length=152, data length=144:
   STATUS          LONG                 1
   TVAL            DOUBLE       1.4768352e+09
   GEOLAT          DOUBLE           78.153000
   GEOLON          DOUBLE           16.074000
   ALT             DOUBLE           471.80000
   BORESITE        DOUBLE           23.700000
   BMOFF           DOUBLE           0.0000000
   BMSEP           DOUBLE           3.2400000
   VDIR            DOUBLE           1.0000000
   PHIDIFF         DOUBLE           1.0000000
   TDIFF           DOUBLE    Array[2]
   INTERFER        DOUBLE    Array[3]
   RECRISE         DOUBLE           100.00000
   ATTEN           DOUBLE           10.000000
   MAXATTEN        LONG                 7
   MAXRANGE        LONG               110
   MAXBEAM         LONG                16
```

When I made my previous comment, I had been using an older (~4.6) version of RST.  But now that I have freshly installed the 5.0 release branch, I am seeing the expected behavior (both with and without using the DLMs).

@agrocott, when you type `make_fit --version` in the terminal what version number is printed?

egthomas:
@agrocott my mistake!  I thought I had disabled the DLMs, but for some reason they were still finding the DLMS loaded by a previous version of RST (even though I had cleared out all of relevant the environment paths).  I agree that the proposed change

```
for s=0,radar.snum-1 do begin
    if (radar.site[s].tval gt tval) then break
endfor
```

matches the logic of the C and DLM code (which does function as intended), eg

```
for (s=0;(s<ptr->snum) && (ptr->site[s].tval<=tval);s++);
```

egthomas (approved):

agrocott:
> @agrocott my mistake! I thought I had disabled the DLMs, but for some reason my IDL session was still finding the DLMs loaded by a previous version of RST (even though I had exited and cleared out all of the relevant environment paths). I agree that the proposed change
> 
> ```
> for s=0,radar.snum-1 do begin
>     if (radar.site[s].tval gt tval) then break
> endfor
> ```
> 
> matches the logic of the C and DLM code (which does function as intended), eg
> 
> ```
> for (s=0;(s<ptr->snum) && (ptr->site[s].tval<=tval);s++);
> ```

Thanks @egthomas - I'm afraid I will have to leave the merging to someone else - I am far to new to this and even hovering my mouse over the merge button makes me nervous...

ecbland:
> @agrocott  I'm afraid I will have to leave the merging to someone else - I am far to new to this and even hovering my mouse over the merge button makes me nervous...

Our group has a policy that we don't merge our own pull requests (or a PR from someone at the same institution). But in all other cases: _feel the fear, do it anyway_ :rofl:

> @agrocott   One further question though, if you don't mind: I realise I have modified rbpos.pro in both this pull request and also in the other one we're discussing - was that a mistake? This one includes the change made in the other, but not vice-versa. If this one is merged first, then the other one, would that one 'undo' the change that is present here but not in the other?

Git is really good at handling this type of thing, so while it's good to be aware and avoid if possible, it's not a big deal. Git will tell us when there is a "merge conflict", and then we are prompted to fix it manually.

I'm going to merge this one now, and then we can revisit #544 (we should probably fix all of the `keyword_set()` instances where zero is a valid input)

PR #549, opened 2022-12-05
Real-time output file name fix

egthomas:
This pull request slightly modifies the behavior of `rtcfit`, `rtgrid`, and `rtsnd` by allowing the user to set the radar name used for the real-time output files via an optional `-name` command line option.

By default, the three-letter station name is taken from the radar parameter block of the real-time data stream.  However, sometimes that value can become corrupted / invalid, causing the `stid` to be set to zero and the files to be labeled as belonging to the `tst` radar.


ecbland:
This compiles ok. Merging now

egthomas:
Thanks @ecbland!

PR #551, opened 2022-12-06
Modifying field_plot to add new -minbeam and -maxbeam options for FOV outlines

egthomas:
This pull request adds two new options to `field_plot` allowing the user to manually set the start / end beam numbers with `field_plot` for radars where the data in a given scan may not match the maximum beam number listed in the hardware file (eg MSI-style radars which have up to 24 beams but may typically use fewer).  If neither option is set then `field_plot` will behave as normal (ie drawing the outline starting at beam 0 up to the maximum number of beams in the hardware file).

For example, here is what normalscan data from CVW looks like by default:

image

and this is using the new `-minbeam 4 -maxbeam 24` options:

image

Likewise for CVE, first the default:

image

and then using `-maxbeam 20`:

image

(sorry @ecbland, I promise this is really the last one!)

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
@@ -126,6 +126,12 @@
 </option>
 <option><on>-ffan</on><od>plot the filled radar field of view.</od>
 </option>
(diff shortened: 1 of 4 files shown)


ecbland:
This works as described. 
Without the min/max beam options, the field of view outline is cut off:
```
field_plot -x -v -coast -fcoast -fan 20180218.0000.01.fhw.fitacf3
```
image


And with the new options, I get a nice fov outline:
```
field_plot -x -v -coast -fcoast -minbeam 1 -maxbeam 22 -fan 20180218.0000.01.fhw.fitacf3
```
image

PR #552, opened 2022-12-09
Fix custom color options in time_plot

egthomas:
Ok - extreme apologies to @ecbland, but this does seem like one last fundamental bug that's worth fixing in `time_plot` prior to the 5.0 release.

Basically, the command line options to control the background, ground scatter, grid, and text colors in `time_plot` are all broken and won't work on the `develop` / `release` branch.  This pull request replaces the incorrect `sscanf` calls with the `PlotColorStringRGB` function which converts the input hex value to something usable by the rest of the plotting code, consistent with all of the other plotting binaries.

(Note this bug appears to date all the way back to pre-4.0 versions of the RST)


egthomas:
Just adding onto the last fix, there is a new `-pbgcol` option which allows the user to independently set the background color of each panel from the overall plot background, eg

`time_plot -png -v -p -e -b 13 -emax 40 -ekey green-pink.key -pbgcol dddddd`

example

ecbland:
> Ok - extreme apologies to @ecbland,

Hahaha, at this point it's just making me laugh :laughing: 

This PR works as described, and I agree it was worth fixing for the 5.0 release. Very nice to be able to separately control the panel background colour :100:

PR #553, opened 2022-12-12
Updating fov_plot documentation to include description of the alpha component

egthomas:
One minor XML documentation update, this time to `fov_plot` to correctly describe the input format for the hex `aarrggbb` values (previously the documentation was missing any mention of the alpha component and would mislead users).  No functionality has changed.

Sorry @ecbland - this really is the last one, I promise!


ecbland:
@egthomas So it's just `fov_plot` that's affected here? There are several other plotting routines with the same descriptions for the colours (e.g. `time_plot`) but I don't think specifying the alpha channel has any effect for those plots.

egthomas:
@ecbland good catch - the documentation for `-rcol` and `-icol` in `iqplot` was also missing the alpha component description.  The documentation for the other plotting routines with alpha channel support is already up-to-date (ie grid, map, field, and vec plot).

One _could_ add some more functionality to `iqplot` by allowing user control over the background, foreground, grid, and text colors, but maybe that is best left for the next release.

PR #554, opened 2022-12-14
Updating iqplot to allow user control over colors and fonts

egthomas:
Ok, this one I'll blame on @ecbland - this pull request (which can be left until after the 5.0 release) adds user control over colors and fonts for `iqplot`, similar to `time_plot` and the other plotting binaries.  See the revised command line help / XML documentation for a description of each of the new options.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/iqplot.1.6/doc/iqplot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/iqplot.1.6/doc/iqplot.doc.xml
@@ -44,6 +44,26 @@
 </option>
 <option><on>-icol <ar>aarrggbb</ar></on><od>set the color of the imaginary (Q) component to <ar>aarrggbb</ar>, specified as the hexadecimal value for the 32-bit alpha,red,green and blue component color [default is red].</od>
 </option>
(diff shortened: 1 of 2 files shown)


ecbland:
@egthomas I tested this by making some plots with some random/funky colour combinations, e.g., 

```
iqplot -txtcol AA553311 -bgcol FFFFAA66 -gcol FF337799 -dgcol FF113355 -pbgcol AA111111 -fgcol FF55FFBB -fontsize 20 -fontname Times-Roman -tfontsize 25 -tfontname Times-Roman -hgt 800 -wdt 1000 20180218.1401.00.fhe.iqdat
```
image


For some colour combinations, there seems to be a display issue where a different colour appears behind the text (admittedly combinations that nobody would ever use due to low contrast). I wouldn't worry about this too much, but I'm curious if you get the same behaviour on your system.

```
iqplot -txtcol FF006688 -bgcol 88FFFFFF -fontsize 20 -wdt 1000 20180218.1401.00.fhe.iqdat
```

image

egthomas:
@ecbland thanks for testing! Yes, I did notice that issue with the text background color as well.  My hunch is it might have to do with using 32-bit ARGB colors vs 24-bit RGB colors for the text (like the other plotting binaries). I briefly experimented with making that change but there seemed to be some other unintended side effects, so I left it as is.

ecbland:
@egthomas Ok, then let's leave it as it is. Merging now!

PR #558, opened 2023-02-16
Adding Iceland East and West hdw files and updating radar.dat

egthomas:
This pull request adds hardware files for the new Iceland East and West radars, and updates the `radar.dat` file accordingly.

### DIFF
--- a/tables/superdarn/hdw/hdw.dat.ice
+++ b/tables/superdarn/hdw/hdw.dat.ice
@@ -1,0 +1,12 @@
+#  Hardware parameters for Iceland East radar.
+#
+# **********************************************************************
+# ==Notes==
+#
+#
+# **********************************************************************
+#
+#                            GEOGRAPHIC             BORE-   OFF  BEAM V  |--------- interferometer ---------| |--- RX ---| RNG MX
+#      YYYYMMDD HH:MM:SS    LAT      LON      ALT   SIGHT   SET  SEP  SG PS  TDIFF  TDIFF    X      Y     Z    RXRIS  ATTN     BM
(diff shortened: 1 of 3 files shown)


aburrell (approved):
Checked out the locations on google maps and the dots are in the middle of the arrays (visible by satellite).

egthomas:
Merging with approval from the radar PI (@sshepherd) and after running multi-frequency tests to validate `tdiff`.

sshepherd:
Hi Kevin,
I’m not sure if there is a question in your email. If there is then let me
know. The Iceland radars have been collecting data for close to a month. I
shared the hardware files after we had all of the antennas up and the
cables laid. These files have tdiff values. Hardware files can always be
changed and should be, particularly when there are changes to a radar. The
most common change is something that affects tdiff; electronics, new
cables, etc. Those values should be updated as frequently as they changed
and as soon as possible so that the data is useful. We are trying to get
these values out so that the data we have been collecting is of use.

Simon

ksterne:
Hello Simon,

There wasn't any implied question in my post to this closed pull request, my apologies if it seemed like there was one.  I understand you receive this by e-mail but just to be clear it is in the open (publicly available) on the github repo.  My apologies for missing the hdw.dat files being shared; I don't think I received those by e-mail, but maybe they were shared another way.  I agree with your notes about why and when hdw.dat files should be changed and updated.  To clarify the last sentence on my previous post, I meant that if a mistake had been made from what was being added to this repo and what you had shared (basically if there was a copy and paste error), then we could correct that in the future.  I hope this clears up what I was commenting on.

Issue #532, opened 2022-10-24
Removing the RFC documentation


ecbland:
# DEPRECATION

We had planned to remove the RFC documentation from the RST after migrating the content to readthedocs in #377.

https://github.com/SuperDARN/rst/blob/c48c5c3e2ae258905c320ece81d5973387af66d4/build/doc/sk/rfc.inx.local.sk#L24-L32

It would be nice to do this for the RST5.0 release, but please chime in if there's a good reason to keep it a bit longer. I had a quick look at the code and it seems fairly simple to remove. (see also #47 when the RFC documentation was added)

While looking at the code I also noticed a script for building tutorial documentation, `make.doc.tutorial`, but I'm not sure where the source code is. Do these tutorials exist?

egthomas:
@ecbland a snapshot of the old tutorials can be found here:

https://web.archive.org/web/20050908143338/http://superdarn.jhuapl.edu/devel/tutorial/index.html

I used that information to populate the tutorials on the wiki page several years ago (e.g., https://github.com/SuperDARN/rst/wiki/Working-With-Data).

I would not be in favor of removing the RFC documentation from the RST, but admit I am probably in the minority on that front.

pasha-ponomarenko:
@ecbland, I don't have any strong opinion on this, but if this information is either redundant or outdated, then it makes sense to remove it from RST and archive somewhere on this repository.

mtwalach:
Personally, I do use the RFC docs occasionally and I do like having them though I appreciate they are perhaps not very accessible and I don't have a good reason either (other than, I occasionally use them but it's probably just Evan and I).

ksterne:
Interesting to hear that there's still usage of these since I'm not sure how up-to-date they are!  Is it helpful or possible to use our readthedocs site (https://radar-software-toolkit-rst.readthedocs.io/en/latest/)?  I think I'm mostly asking @egthomas and @mtwalach so far.  When we discussed this a bit ago, we figured moving all of the documentation to a central location would be beneficial.

ksterne:
Status update: Still need to check on how much of the older RFC content has been captured in readthedocs.  Ideally, this would be consolidated into one documentation set.

PR #559, opened 2023-03-14
Fixing broken HTML documentation links

egthomas:
This pull request fixes the few remaining broken links in the HTML documentation - no code functionality is modified.


egthomas:
> Does the rplot documentation not exist anymore?

No, `rplot` was from the pre-3.x era of RST and from what I can tell it evolved into other libraries such as `grplot`, which has no documentation (and probably never will).

ksterne:
Confirming that this is still good with compilation, including the documentation building.  I'm reminded of the conversation in #532 to see if we can move this information to a central location.

PR #560, opened 2023-03-14
Adding new -databeam option to field_plot

egthomas:
This pull request adds a new `-databeam` option to `field_plot`, which optionally allows the user to draw the FOV edges according to the beams present in the scan.

By default, `field_plot` will only draw right / left FOV edges if a beam present in the scan is either 0 or matches the maximum beam in the hdw file (see #551 for more details).  So for radars that do not always use all available beams, this new option is helpful for drawing an FOV extent which reflects the actual data coverage.


pasha-ponomarenko:
@egthomas, could you please provide testing instructions for this pull request.

egthomas:
@pasha-ponomarenko sure, no problem.  Here is an example when Saskatoon was operating a 3-beam ulfscan mode and plotted using the default `-fan` option with `field_plot`:

```
field_plot -png -def -v -fan 20180106.1800.02.sas.fitacf > default.png
```

def

and here is what it looks like using the new `-databeam` option:

```
field_plot -png -def -v -databeam 20180106.1800.02.sas.fitacf > databeam.png
```

databeam

Notice how the left/right edges of the outermost beams in the scan are now correctly drawn, whereas with the default `-fan` option those edges will only be drawn in the beams are either 0 or 15 (ie according to the hardware file).  This is particularly useful with the US MSI-style radars with 22-24 beams (eg Fort Hays, Christmas Valley, etc), and this new `-databeam` option has been useful for automatically generating real-time `field_plot` images when there is no prior knowledge of what modes may be operating.

This new option is admittedly not perfect, as seen from the gap along the top of the SAS FOV (due to the skipped beam in this mode), but I believe it is still a worthwhile feature.

pasha-ponomarenko (approved):

Issue #565, opened 2023-05-02
`trim_iq` doubles length of data array


ecbland:
When creating an `iqdat` file using `trim_iq`, the `data` array has double the number of elements as the input file (and the output file approximately doubles in size). 

I expect that this behaviour comes from the [`iq` library](https://github.com/SuperDARN/rst/tree/main/codebase/superdarn/src.lib/tk/iq.1.7) rather than `trim_iq` itself. Does anyone know if this behaviour is intended? 

```
// ORIGINAL IQDAT FILE
$ dmapdump 20180218.1401.00.bks.iqdat | grep '"data' | head
	short	"data" [54472]
	short	"data" [54472]
	short	"data" [54472]
	short	"data" [54472]
	short	"data" [54472]
	short	"data" [54472]
	short	"data" [54472]
	short	"data" [56948]
	short	"data" [54472]
	short	"data" [54472]

// OUTPUT FROM TRIM_IQ (no actual trimming)
$ trim_iq 20180218.1401.00.bks.iqdat | dmapdump | grep '"data' | head
	short	"data" [108944]
	short	"data" [108944]
	short	"data" [108944]
	short	"data" [108944]
	short	"data" [108944]
	short	"data" [108944]
	short	"data" [108944]
	short	"data" [113896]
	short	"data" [108944]
	short	"data" [108944]
```

egthomas:
@ecbland I think this has to do with a change in how IQ samples were treated in the QNX4 vs Linux/QNX6 ROS.  The short answer can be found in the `IQEncode` function:

https://github.com/SuperDARN/rst/blob/734542037d86b24475d8da571ed668659a858e27/codebase/superdarn/src.lib/tk/iq.1.7/src/iqwrite.c#L64

From what I can tell, the QNX4 code set an `nchannel` parameter as either one or two depending on whether XCFs were present.  This `nchannel` was used to both populate the IQ data array, and then set the value of an intermediate `rxchn` and then `iq->chnnum` which is stored in the `iqdat` files themselves:

https://github.com/SuperDARN/ros-archive/blob/a87c608b5e96d33fc1267fa3a31d170aa5d4aa28/qnx/code/src.lib/radarqnx4/integrategc214.1.29/src/integrategc214.c#L230-L241

[...]

PR #566, opened 2023-05-03
Fixing bug in trim_iq

egthomas:
This pull request modifies `IQEncode` to remove the extra multiplication by 2 causing a bug in `trim_iq` output data size (#565), and modifying `make_raw`, `make_sim`, and `sim_real` to correctly handle the IQ `chnnum` parameter.  This bug appears to have originated from the development of the Linux/QNX6 ROS, which is no longer part of the RST.

### DIFF
--- a/codebase/superdarn/src.bin/tk/tool/make_raw.1.8/make_raw.c
+++ b/codebase/superdarn/src.bin/tk/tool/make_raw.1.8/make_raw.c
@@ -326,9 +326,9 @@ int main (int argc,char *argv[]) {
     badrng=ACFBadLagZero(&tprm,prm->mplgs,lag);
 
     if (chnnum > 0) {
-      rngoff = 2*chnnum;
+      rngoff = chnnum;
     } else {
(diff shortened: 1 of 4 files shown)


ecbland:
Thanks for looking into this @egthomas! I will have time to test this early next week.

egthomas:
@ecbland thanks for testing - I think you've done a very thorough job!  Note that `make_raw` also has a `-chnnum` option for overriding the `chnnum` value in the `iqdat` files (likely because of this very issue).

pasha-ponomarenko:
@ecbland, good job! 👍
I also discovered that for proper conversion one needs to specify the number of channels in `make_raw` by using `-chnnum` switch and was going to look into it later, but you beat me to that! 🙂
I will install this branch on my Linux box and check how it works, but I believe your testing is more than sufficient. 

@egthomas , thanks a lot for fixing this problem! 👌

pasha-ponomarenko:
@ecbland , I tested `iq_trim`, and it works. Please go ahead with merging.

PR #273, opened 2020-02-18
Misc bug fixes to fitacfserver and map_addimf

egthomas:
This pull request fixes two bugs (in `fitacfserver` and `map_addimf`) and renames the `rst.version` file to make it hidden and de-clutter the root directory.

To test the [fitacfserver](https://superdarn.github.io/rst/superdarn/src.bin/tk/tcpip/fitacfserver/index.html) changes, try serving a `fitacf`-format file on a system-assigned local port, eg:

`fitacfserver -r yyyymmdd.hhmm.ss.rad.fitacf`

The system-assigned port number will be stored in the file `port.id`.  You can then use [fitacfclient](https://superdarn.github.io/rst/superdarn/src.bin/tk/tcpip/fitacfclient/index.html) or the [Java real-time client](https://github.com/SuperDARN/java_realtime) to display the fitted line-of-sight data.  On the develop branch however, the system-assigned port number is not stored in `port.id` (I see `0` when testing locally).

The second bug fix is to [map_addimf](https://superdarn.github.io/rst/superdarn/src.bin/tk/tool/map_addimf/index.html) when using CDF files from the Ace or Wind spacecraft as the solar wind input, and is related to allocating memory for the new Vx and Kp fields.  This issue and solution was originally identified by Bill Bristow.

PR #569, opened 2023-05-24
Updates to fitacfclientgui

egthomas:
This pull request implements major updates to `fitacfclientgui` ~~when using the `-color` option~~, including:

- an on-screen menu explaining keyboard controls
- use of a data buffer allowing all beams to update when plotting options change
- a new ground scatter flag option (which can also be toggled with the `g` key)
- up/down arrow control of the color scale maximum

As an example using the Iceland East real-time data stream,

iceland_realtime

For testing, if you don't have access to a real-time data stream you can simulate one using `fitacfserver` (eg https://github.com/SuperDARN/rst/pull/273).

### DIFF
--- a/codebase/superdarn/src.bin/tk/tcpip/fitacfclientgui.1.0/doc/fitacfclientgui.doc.xml
+++ b/codebase/superdarn/src.bin/tk/tcpip/fitacfclientgui.1.0/doc/fitacfclientgui.doc.xml
@@ -15,6 +15,8 @@
 </option>
 <option><on>-color</on><od>display the data stream in color according to the selected parameter (default is power).</od>
 </option>
+<option><on>-gs</on><od>color ground scatter white (velocity only).</od>
+</option>
 <option><on>-p</on><od>plot power.</od>
 </option>
 <option><on>-v</on><od>plot velocity.</od>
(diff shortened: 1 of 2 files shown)


ksterne:
Thanks for taking a look at this @egthomas.  I think that makes sense to set the default as the `-color` option since the functionality is greatly improved.  I'm not sure of a case where this shouldn't be the default option.  It looks like it would be easy to modify the default if it became necessary in the future.  I'll merge this in now.

ksterne:
One random question related to this program but not to the updates, in testing this I would enter the command for this GUI on a real radar stream without anything happening.  I would have to run it multiple times in order for the display to come up.  It was almost like it had to get the timing right to grab the first packet.  Does that seem to sense and/or is that normal?  It's possible I'm grabbing the data the wrong place causing this issue.

egthomas:
@ksterne I've also noticed that when trying to connect to the VT radars' real-time streams (and probably some other groups' radars too).  Not sure if it's a latency thing or something, but when I call `fitacfclientgui` at the radar site it connects on the first try; similarly if I launch a local radar simulator or fitacf server.

ksterne:
Thanks for the quick response!  That's good to hear that it's not just me.  I'd think its something with trying to connect to another server.  Maybe one day we'll get bugged enough by that to see what's going on or print out some message of what's going on.

PR #575, opened 2023-06-02
Added CI compilation test

aburrell:
Added a continuous integration (CI) framework, which currently just tests whether or not RST compiles based on our current instructions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,43 @@
+# This workflow will install RST and eventually run tests
+
+name: Compilation Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        cc: ["gcc-default"]
+        cdf: ["39_0"]
+
+    name: ${{ matrix.cc }} on ${{ matrix.os }} with CDF ${{ matrix.cdf }}
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Install Ubuntu dependencies
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        sudo apt-get update
+        cat requirements.linux | xargs sudo apt-get install
+        tar -xzvf cdf${{ matrix.cdf }}-dist-cdf.tar.gz
+        cd cdf${{ matrix.cdf }}-dist
+        make OS=linux ENV=gnu all
+        make test
+        sudo make INSTALLDIR=/usr/local/cdf install
+
+    - name: Bash configuration
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        echo "export RSTPATH=$HOME/rst" >> $HOME/.bashrc
+        echo ". $RSTPATH/.profile.bash" >> $HOME/.bashrc
+        source ~/.bashrc
+
+    - name: Build and make code
+      run: |
+        cd $RSTPATH/build/script
+        make.build
+        make.code
--- a/requirements.ubuntu
+++ b/requirements.ubuntu
@@ -1,0 +1,8 @@
+build-essential
+libhdf5-serial-dev
+libncurses-dev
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    name: ${{ matrix.cc }} on ${{ matrix.os }} with CDF ${{ matrix.cdf }}
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Install Ubuntu dependencies
+      if: startsWith(matrix.os, 'ubuntu')
+      run: |
+        sudo apt-get update
+        cat requirements.ubuntu | xargs sudo apt-get install
```
egthomas, line 25: Is it possible to rename `requirements.ubuntu` as `.requirements.ubuntu` (ie make it a hidden file), just to help reduce the clutter in the top-level `rst` directory?

.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+        cd cdf${{ matrix.cdf }}-dist
+        make OS=linux ENV=gnu all
+        make test
+        sudo make INSTALLDIR=/usr/local/cdf install
+
+    - name: Build and make code
+      run: |
+        export RSTPATH=$(pwd)
+        source .profile.bash
+        cd build/script
```
egthomas, line 39: I'm not sure if this line is still necessary - I haven't needed it on recent installations.

.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+        make.build
+        make.code
```
egthomas, line 41: Could we also include `make.doc`, either as part of `Build and make code` or as a post-run action?

.github/workflows/main.yml
aburrell: Ok, I think you were referring to the directory change, so I made that change and it's working 😄

ksterne:
I think the basic test here is that this branch still compiles?  And I guess a double check that the github workflow has been setup correctly?  It looks like all of that has been done!  Awesome catch that github has this feature now.  We'd lost that with the restrictions travis CI put in place from my vague memory.

.github/workflows/main.yml
```diff
@@ -0,0 +1,46 @@
+# This workflow will install RST and eventually run tests
+
```
ksterne, line 2: Are you able to add your authorship to this file @aburrell?  I see @mts299 and I did this for the mkdocs.yml file on the top directory.

ksterne:
Hello @aburrell, similarly on this pull request, do you think you can take a look these two things?  If it's helpful, I can try to help answer the question @egthomas brought up.

aburrell:
Looking for an easy win today so came back here 😀 

@egthomas - Honestly, I think having a requirements file that's easy for people installing to find and use would be good to have in the top level directory, but it doesn't matter where it lives.  For now I've put it where you requested, but made it visible.  Maybe in the future we can have a Requirements directory or something that helps with the installation.

@ksterne - Updated the zenodo json and the authors file.

.github/workflows/main.yml
aburrell: Is there another place I should put authorship?  Do you want it on this file in particular?

egthomas:
@aburrell thanks for making those changes!  It looks like the compilation test is currently failing because it can't find the requirements file:

```
cat: .github/workflow/requirements.ubuntu: No such file or directory
```

https://github.com/SuperDARN/rst/actions/runs/6408566485/job/17397848606#step:3:49

aburrell:
@egthomas the requirements file needs to be somewhere in the main tree for it to be accessible when building.  I moved it to be a hidden file in the build directory for now.

aburrell:
RTD failure documented in #601

ecbland:
@aburrell I've added a new configuration file for readthedocs. The build is working now.

aburrell:
@ksterne pinging you for a review.  I don't care about adding authorship for the yaml, given that it's copypasta from other projects and not original creative property.

PR #578, opened 2023-06-23
Adding new -xshift and -yshift options to field_plot

egthomas:
This pull request modifieds the `field_plot` binary and `MapStereographic` function (in `tfunc.c`) to support user-controlled horizontal and vertical shifting of the plotting region.  This is particularly useful for radars where the default options, or some combination of `-latmin` / `-lat` / `-lon`, aren't quite sufficient.

For example, this is the default for Iceland East (ICE) in magnetic coordinates:

default

If we want to zoom in a little better, one could set `-latmin 60 -lat 85 -lon 65`:

no_shift

Finally, with the new `-xshift` option one can actually center the radar FOV, eg `-latmin 60 -lat 85 -lon 65 -xshift -0.35`:

shift

These new `-xshift` and `-yshift` options can be used with any `field_plot` figure in the (default) stereographic projection; orthographic projections will be unchanged.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
+++ b/codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
@@ -92,9 +92,9 @@
 </option>
 <option><on>-flip</on><od>flip the direction of the X-axis.</od>
 </option>
-<option><on>-xshift</on><od>shift the plot horizontally by a percentage of the plotting region (+/- 1.0)</od>
+<option><on>-xshift</on><od>shift the plot horizontally by a proportion of the plotting region (+/- 1.0)</od>
 </option>
-<option><on>-yshift</on><od>shift the plot vertically by a percentage of the plotting region (+/- 1.0)</od>
+<option><on>-yshift</on><od>shift the plot vertically by a proportion of the plotting region (+/- 1.0)</od>
 </option>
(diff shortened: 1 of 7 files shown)


codebase/superdarn/src.bin/tk/plot/field_plot.1.12/doc/field_plot.doc.xml
```diff
@@ -92,6 +92,10 @@
 </option>
 <option><on>-flip</on><od>flip the direction of the X-axis.</od>
 </option>
+<option><on>-xshift</on><od>shift the plot horizontally by a percentage of the plotting region (+/- 1.0)</od>
+</option>
+<option><on>-yshift</on><od>shift the plot vertically by a percentage of the plotting region (+/- 1.0)</od>
+</option>
```
ecbland, line 98: Would this be better wording since the shift values are in the range `-1 < shift < +1`?
```suggestion
<option><on>-xshift</on><od>shift the plot horizontally by a proportion of the plotting region (+/- 1.0)</od>
</option>
<option><on>-yshift</on><od>shift the plot vertically by a proportion of the plotting region (+/- 1.0)</od>
</option>
```

ecbland:
This works well. Just one suggestion about the documentation above.

```
field_plot -png -def -v -xshift -0.2 -yshift -0.2 20141219.0802.00.cly.fitacf3 > plot.png
```

plot

PR #589, opened 2023-08-25
Fix xmldoc bins

egthomas:
This pull request adds option error handling and `--version` support to the `scdoc` and `xmldoc` binaries, and also modifies `make.build` so that `xmldoc` is compiled twice.  This allows for the help message (printed with `xmldoc --help`) to be correctly displayed (since `xmldoc` is needed to parse its own XML documentation).  No functionality has changed.


ecbland:
This compiles ok, merging now

PR #591, opened 2023-09-05
Removing phidiff comments from elevation_v2 algorithm

egthomas:
This pull request removes unnecessary references to `phidiff` from the `elevation_v2` algorithm to reduce possible confusion; `phidiff` is handled elsewhere in the fitting algorithms when calculating `phi0`.


ecbland:
This looks good, merging now

PR #592, opened 2023-09-05
Modifying FitDecode of fit library to check against both phi0 and elv

egthomas:
This pull request modifies `FitDecode` of the `fit` library to check against both the `phi0` and `elv` fields when reading fitted XCF data to determine whether to populate the XCF arrays. This is necessary because older radars do not transmit all of the fitted parameters in real-time (ie `phi0`), meaning that the elevation angles were not being populated when received by modern routines (eg `fitacfclientgui`).


ecbland:
@egthomas This branch compiles ok and the change looks sensible. Do you have a fitacf file that includes the `elv` parameter but not `phi0` that I can use to test? No worries if not.

egthomas:
@ecbland thanks for testing. No unfortunately I don't have an example FITACF file, but I could share an example real-time link from an older radar via email if that helps.

ecbland:
I think this is fine to merge as-is.

PR #593, opened 2023-09-11
Modifying rtcfit to fix documentation and remove unused -f option

egthomas:
This pull request fixes the html documentation for `rtcfit` (ie changing `make_cfit` to `rtcfit` in the examples) and removes an unused `-f` option, thus making the output of `--help` and `--option` consistent. No functionality has changed.


ecbland:
This one looks good as well...merging now

PR #596, opened 2023-09-26
Cleaning up profile scripts

egthomas:
This pull request removes the `.tcsh` profile scripts since they have gone unchanged for years and are redundant given the equivalent set of `.bash` profile scripts; users can still refer to previous versions of the RST if they need them. This pull request also removes an unnecessary line in `.profile.bash` which modifies the user's `PATH` environment variable - the RST should still compile and work as usual after removing this line.


ksterne:
I can confirm that this branch still compiles.  It would be easy enough to bring these environment variables back in if needed.

PR #597, opened 2023-09-26
Updating fitacfclientgui to add support for toggling between gate and slant range

egthomas:
This pull request adds support to `fitacfclientgui` for toggling between the range gate number and slant range (km) along the top of the display by pressing the `r` key or with a new `-r` command line option.  There has also been a small change to only print the header origin information if it is actually present in the real-time data stream.


egthomas:
Here is a screenshot showing real-time Iceland East data with range gates running along the top of the display at 10 gate increments (default):

rt_gate_crop

and here is the same display a few seconds later after pressing the `r` key to display slant range instead:

rt_range_crop

pasha-ponomarenko (approved):

PR #610, opened 2024-04-07
Modifying rtsnd to add support for bmoff parameter from hdw file

egthomas:
As indicated by the title, this pull request adds support for the `bmoff` parameter when calculating `bmazm` in the `rtsnd` binary (similar to #517 and #521).


ksterne:
I see this is compiling on Ubuntu 22 VM.  Looks simple enough to add in.

Issue #452, opened 2021-08-18
Linguist overrides in .gitattributes


egthomas:
Github uses something called [linguist](https://github.com/github/linguist) to detect the language / file type of repositories for the purposes of language statistics shown on the main page, syntax highlighting, and organizing search results.  In this repository for the RST, a significant number of files are misclassified, eg

- IDL `.pro` files as Prolog (e.g., [aacgm_v2.pro](https://github.com/SuperDARN/rst/blob/master/codebase/analysis/src.idl/lib/aacgm.2.0/aacgm_v2.pro))
- C `.h` files as C++
- `hdw.dat` files as all kinds of different extensions (e.g, [hdw.dat.sas](https://github.com/SuperDARN/rst/blob/master/tables/superdarn/hdw/hdw.dat.sas))
- AACGM `.asc` coefficient files as AGS script

The `.gitattributes` file I've tried using to [override](https://github.com/github/linguist/blob/master/docs/overrides.md) this behavior looks something like this:

```
*.pro linguist-language=IDL
*.h linguist-language=C

hdw.dat.* -linguist-detectable
*.asc -linguist-detectable
```

but I'm not sure if this is correct, and it is not clear to me under what circumstances the language statistics / syntax highlighting will actually be updated (eg does the file need to be modified by a new commit, or does a new version of linguist need to be pushed out to Github, etc).

Is this something worth pursuing?  I personally find it helpful to have the correct syntax highlighting available when browsing code on Github, and it can also be difficult to use the search feature when the file you're looking for appears under the wrong language.

ecbland:
@egthomas Completely agree that the misclassified files are annoying! 

I just installed `github-linguist` locally and put your example `.gitattributes` file in the root directory of RST:
```
$ github-linguist 
80.52%  4807712    C
13.80%  823872     IDL
3.39%   202673     Makefile
1.22%   72673      Shell
0.98%   58274      Prolog
0.05%   2974       CSS
0.02%   1270       JavaScript
0.01%   627        Scheme
0.01%   621        SAS
0.01%   478        Scilab
```

So it doesn't seem to detect the `.gitattributes` file. I then renamed it to `.git/info/attributes`:
```
$ github-linguist 
80.52%  4807712    C
14.79%  883257     IDL
3.39%   202673     Makefile
1.22%   72673      Shell
0.05%   2974       CSS
0.02%   1270       JavaScript
```

[...]

PR #611, opened 2024-04-12
Adding .gitattributes file

egthomas:
This pull request addresses #452 by overriding the linguist file detection with a new `.gitattributes` file, which has been found to work well in a similar RST-based repository.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+*.pro linguist-language=IDL
+*.h linguist-language=C
+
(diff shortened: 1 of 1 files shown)


ksterne:
Finally got everything installed for `github-linuguist` and tested this out on an Ubuntu 22 VM.  Looks like I didn't have the same issue brought up in #452....but that was over 3 years ago, so maybe something was updated since then.  Think we can add/change this around if this becomes a problem in the future.

PR #612, opened 2024-04-23
Bugfix: setting `xqflg` to zero in `fit_speck_removal`

ecbland:
This PR fixes a bug identified by @pasha-ponomarenko in `fit_speck_removal`. For radars with `xcf=1`, the XCF quality flag also needs to be set to zero. This bug was not obvious when reading the output files using standard RST C code, but it does become apparent when reading the data with IDL.


pasha-ponomarenko:
I have deployed the branch and tested it against HOK and HKW data for 20_22UT on 05 November 2018, which contain a lot of 'salt-and-pepper' noise caused by their mutual cross-interference. With the updated script both datasets are processed properly, i.e., the specks are effectively filtered out (see examples below).

HOK unfiltered 
image

HOK filtered
image

HKW unfiltered
image

HKW filtered
image

pasha-ponomarenko:
I guess I can merge it now, although there is a warning "All checks failed", which refers to Read the Docs. Is it important?

pasha-ponomarenko (approved):

Issue #570, opened 2023-05-26
noise.lag0 and noise.vel in fitacf-format files


egthomas:
I have been trying to better educate myself on the different noise parameters in the `rawacf` and `fitacf` format files, and was wondering about the `noise.lag0` and `noise.vel` fields in `fitacf` files.  As far as I can tell, these fields were populated by older versions of FITACF via the `noise_acf` function, which was removed from FITACF following discussions at the 2006 SD workshop, eg:

```
# $Log: version.info,v $
# Revision 2.0  2007/02/02 21:42:06  code
# Changes agreed to at the 2006 SuperDARN meeting.
```

and

```
 $Log: do_fit.c,v $
 Revision 1.16  2007/02/02 21:38:42  code
  Removing everything belonging to NOISE_ACF calculation and extraction
 (lines 153-160 and 198-251 in Revision 1.15)
```

(and likely influenced by the work of [Ponomarenko and Waters [2006]](https://angeo.copernicus.org/articles/24/115/2006/))

The "modern" fitting algorithms in the RST (eg, `fitacf2`, `fitacf3`, `lmfit2`) all seem to set `noise.lag0` and `noise.vel` to zero (the `noise.sky` value, however, is of course still used).

So, here is a philosophical question - we have added several new fields to `fitacf`, `map`, etc files over the years, but would we ever **_remove_** a field that is no longer relevant?  Is there value in reading/storing the `noise.lag0` and `noise.vel` values from `fitacf`-format files (or more likely `fit`-format files) processed with pre-2007 versions of the software?

Note there are several fields from the old parameter block which were dropped with the conversion from `dat`/`fit` to `rawacf`/`fitacf` (e.g., `nbaud` and the various `usr_*`-defined parameters).

pasha-ponomarenko:
@egthomas , from years ago my recollection is that there was some serious technical problem with removing the existing fields, but I never knew the exact details. I think then we were talking about removing fields related to XFC power fitting.

carleyjmartin:
jjknvjlndfgl

ecbland:
@pasha-ponomarenko I removed all the unused XCF fields for `fitacf3` files in #472, so there shouldn't be any technical issue here. I'm all in favour of @egthomas's suggestion to remove the unused fields.

egthomas:
[...]

PR #614, opened 2024-05-06
Modifying FitReadRadarScan to store sky noise in RadarBeam structure rather than search noise

egthomas:
This pull request addresses one issue recently raised in #570, where the fitted sky noise value is now used instead of the `search.noise` parameter to populate the `noise` fields in grid and map files.  You can test this by comparing noise values in grid files for different radars (some of which calculate search noise differently or not at all).


egthomas:
I've also just updated the online documentation to more clearly state that the noise values recorded in grid and map files originate from the sky noise parameter.

pasha-ponomarenko (approved):

Issue #609, opened 2024-04-04
Installing RST on a Mac...once more!


mtwalach:
# BUG

I am trying to install RST on a new MacBook and seeing an error I have never seen before. When running `make.code`. The installation aborts upon the following error: 
`locate.c:190:31: error: incompatible function pointer types passing 'int (struct dirent *)' to parameter of type 'int (*)(const struct dirent *)' [-Wincompatible-function-pointer-types]
    status=scandir(dir,&dlist,dsel,alphasort);
                              ^~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/dirent.h:150:11: note: passing argument to parameter here
    int (*)(const struct dirent *), int (*)(const struct dirent **, const struct dirent **)) __DARWIN_INODE64(scandir);
          ^
10 warnings and 1 error generated.
make: *** [locate.o] Error 1
Compilation Aborted. `

I am guessing it's something to do with the C libraries or the compiler but it is not something I have ever encountered before, so I don't know how to fix it. Any ideas? 

## Priority

- [ ] Minor - typo, incorrect naming, fails in a specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation
- [ ] Urgent - failure in usage, large misrepresentation of data

## RST version

*Does the bug occur in an official RST release?*

- [x] Yes. RST version: 5.0
- [ ] No. Branch containing the bug (e.g. develop): ______

## Example of the bug
The error above occurs when compiling the following: 

`Compiling Binary:/Users/walach/Documents/rst/codebase/analysis/src.bin/cdf/istp_plot.1.11

make clean
rm -f *.o
rm -f version.h
rm -f errstr.h
rm -f hlpstr.h
rm -f istp_plot
make
make.version /Users/walach/Documents/rst/codebase/analysis/src.bin/cdf/istp_plot.1.11
make.help
./doc/istp_plot.doc.xml `

See log file here: 
[superdarn-rst.build.20240403.1717.txt](https://github.com/SuperDARN/rst/files/14873022/superdarn-rst.build.20240403.1717.txt)

### Attempts

*What have you tried already to fix it?*
 I followed @billetd's instructions (https://github.com/SuperDARN/rst/issues/556#issuecomment-1652250845) for M chipped Macs and am still getting errors.

*What have you tried to work around it?*
N/A

### Data Access

[...]

PR #616, opened 2024-07-08
Adding new srng field to grid and map files

egthomas:
This pull request adds a new `srng` field to grid and map files that contains the (average) slant range of the velocity vector recorded by a radar in a given grid cell.  All C and IDL read/write routines have been updated to handle files both with and without this new field.

Note this is a preliminary step towards more flexible filtering of line-of-sight data based on the measured slant range, which can possibly be contaminated by E-region echoes at near ranges or incorrectly mapped at very far ranges.  Ultimately, it should be faster to apply a post-processing range filter on this new `srng` field than using the current `-minsrng` and `-maxsrng` options with `make_grid` (since those require slant range calculations for all ranges of every input scan).

To test, try creating grid and map files with your favorite options and see if anything breaks.  The contents of the `srng` field can be viewed with either `dmapdump -d` or your favorite IDL plotting software (at least until pydarnio is updated).

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/time_plot.1.8/time_plot.c
@@ -1428,18 +1428,18 @@ int main(int argc,char *argv[]) {
         }
 
         if ((geoflg) || (magflg)) {
-          double rho,blat,tlat,lon;
+          double rho,blat,tlat,lon,srng;
           double tmp_swap;
           if (magflg) RPosMag(0,tplot.bmnum,rng-1,site,tplot.frang,
                               tplot.rsep,tplot.rxrise,300,&rho,
-                              &blat,&lon,chisham,old_aacgm);
+                              &blat,&lon,&srng,chisham,old_aacgm);
           else RPosGeo(0,tplot.bmnum,rng-1,site,tplot.frang,tplot.rsep,
-                       tplot.rxrise,300,&rho,&blat,&lon,chisham);
+                       tplot.rxrise,300,&rho,&blat,&lon,&srng,chisham);
 
           if (magflg) RPosMag(0,tplot.bmnum,rng,site,tplot.frang,tplot.rsep,
-                              tplot.rxrise,300,&rho,&tlat,&lon,chisham,old_aacgm);
+                              tplot.rxrise,300,&rho,&tlat,&lon,&srng,chisham,old_aacgm);
           else RPosGeo(0,tplot.bmnum,rng,site,tplot.frang,tplot.rsep,
-                       tplot.rxrise,300,&rho,&tlat,&lon,chisham);
+                       tplot.rxrise,300,&rho,&tlat,&lon,&srng,chisham);
 
           if (tlat<blat) {
             tmp_swap=blat;
--- a/codebase/superdarn/src.dlm/rposdlm.1.9/rposdlm.c
+++ b/codebase/superdarn/src.dlm/rposdlm.1.9/rposdlm.c
@@ -885,7 +885,7 @@ static IDL_VPTR IDLRadarConvert(int type,int argc,IDL_VPTR *argv,char *argk) {
   struct RadarIDLSite *isite;
   int frang=180,rsep=45,rxrise=0;
   double height=300;
-  double rho,lat,lng;
+  double rho,lat,lng,srng;
 
   IDL_VPTR outargv[11];
   static IDL_LONG chisham;
@@ -1140,7 +1140,7 @@ static IDL_VPTR IDLRadarConvert(int type,int argc,IDL_VPTR *argv,char *argk) {
     if (type !=0) RPosGeoGS(center,bcrd,rcrd,&site,frang,rsep,rxrise,height,
                             &rho,&lat,&lng);
     else RPosGeo(center,bcrd,rcrd,&site,frang,rsep,rxrise,height,
-                 &rho,&lat,&lng,chisham);
+                 &rho,&lat,&lng,&srng,chisham);
 
      rptr[n]=rho;
      latptr[n]=lat;
@@ -1170,7 +1170,7 @@ static IDL_VPTR IDLRadarConvert(int type,int argc,IDL_VPTR *argv,char *argk) {
     if (type !=0) RPosGeoGS(center,bcrd,rcrd,&site,frang,rsep,rxrise,height,
                             &rho,&lat,&lng);
     else RPosGeo(center,bcrd,rcrd,&site,frang,rsep,rxrise,height,
-                 &rho,&lat,&lng,chisham);
+                 &rho,&lat,&lng,&srng,chisham);
 
     IDL_StoreScalar(argv[8],IDL_TYP_DOUBLE,(IDL_ALLTYPES *) &rho);
     IDL_StoreScalar(argv[9],IDL_TYP_DOUBLE,(IDL_ALLTYPES *) &lat);
--- a/codebase/superdarn/src.idl/lib/main.1.25/cnvmap.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/cnvmap.pro
@@ -217,7 +217,7 @@ function CnvMapRead,unit,prm,stvec,gvec,mvec,coef,bvec
            'program.id','noise.mean','noise.sd','gsct', $
            'v.min','v.max','p.min','p.max','w.min','w.max','ve.min', $
            've.max', $
-           'vector.mlat','vector.mlon','vector.kvect', $
+           'vector.mlat','vector.mlon','vector.kvect','vector.srng', $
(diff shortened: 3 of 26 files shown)


egthomas:
@pasha-ponomarenko thanks for the question.  This new field only stores the slant range (or group path) to the line-of-sight measurement, which can vary depending on radar operating parameters (eg, `frang` and `rsep`) and is independent of any geolocation method.  So here a grid file is produced as normal using either the standard or Chisham virtual height models to calculate latitude/longitude locations and assign them to a grid cell; this just adds one extra field allowing you to possibly exclude data afterwards based on the original slant range.

pasha-ponomarenko:
OK, so it just memorises the slant range for each contributing radar which is not stored otherwise, right? So that it enables the user to post-filter the data without re-running the whole gridding/mapping/fitting chain.

egthomas:
@pasha-ponomarenko yes, exactly!  Originally no slant range information was retained once a grid vector was created (note for the TS18 model I modified the code to store the slant range in the grid files' power field, which is not a portable or long-term solution).

Here is an example plot using the new `-rng` option with `grid_plot` (note it can take the average, min, or max in a cell depending on the command line options if data from more than one radar is present):

mtwalach:
Hi, I really welcome these changes and it looks great but I am afraid I can't test right now. 
I don't have a working copy of RST-dev at the moment and due to all the install issues I have been having  (https://github.com/SuperDARN/rst/issues/609). I don't really want to mess with the little bits of RST that I have got right now because it took me two days to get working what I have got. Until I have a set-up that allows me to switch branches and reinstall RST quickly, I can't do any testing I am afraid.

egthomas:
@mtwalach no problem! Even if you are unable to test, any feedback or ideas based on your own extensive work with map files will be greatly appreciated as this pull request is inevitably discussed.

egthomas:
I've just modified the `grid_filter` tool (which previously only supersampled a grid file in time, which seems useless?) to allow removal of grid vectors from a file based on user-defined slant range criteria.  Note that by default, nothing will be removed (default limits are 0 km and 10,000 km).

With the `-vb` option, some diagnostics are printed about the number of grid vectors in each record before and after filtering, eg

```
grid_filter -minrng 800 -maxrng 2000 -vb 20240101.north.grdmap > out.grdmap
2024-01-01 00:00:00  530 > 349 pts (65.85%)
2024-01-01 00:02:00  527 > 340 pts (64.52%)
2024-01-01 00:04:00  549 > 357 pts (65.03%)
2024-01-01 00:06:00  553 > 379 pts (68.54%)
```

The idea is to eventually support a `-cpid` option with this routine as well that can apply some further exclusion based on radar operating mode (ie CPID), but that requires further discussion by the DPTF and others.

egthomas:
I've just made one further change to `grid_filter`:  with the new `-cpid` option, one can now also exclude velocity vectors from a grid file whose CPIDs do not fall within the ranges listed in a new `cpid.dat` file, eg something like

```
0000 1000
3300 3310
3500 3503
```

Obviously these values will need further iteration, so feel free to modify for your own testing purposes.

egthomas:
Note that support for the new `srng` field in grid and map files has also been added to `pydarnio` in https://github.com/SuperDARN/pyDARNio/pull/75 and https://github.com/SuperDARN/pyDARNio/pull/76 (pending the outcome of this pull request).

mtwalach:
Hi @egthomas, @ksterne & @pasha-ponomarenko, 

I tried testing this as we discussed and here is how far I got...
1) I installed the feature branch on Ubuntu 24.04 without issues: ✅ (yay!)
2) I made a grid file using make_grid - and it's not empty: ✅(yay!)
3) I used dmapdump to check if the contents are sensible - yes, there is data and rng is in there: ✅(yay!)
4) I tried to use the grid_filter tool using 
`grid_filter -minrng 800 -maxrng 2000 -vb test.grid > out.grdmap`  - and that also has data: ✅(yay!)
5) I can't plot the outputs to check.. for whatever reason, the version(s) of pydarn I have struggle to read in the data - I think it's not the latest, though because for some other reason I can't install the latest and I haven't got to the bottom of that yet either. When I use `grid_plot` on the ubuntu machine, it just generates empty files (zero bytes). 

I have tried: 
`grid_plot -ps -chisham ./out.grdmap > ./out.ps`
`grid_plot -ps -stdout -chisham ./out.grdmap > ./out.ps`
`grid_plot -png -chisham ./out.grdmap > ./out.png`
`grid_plot -png -stdout -chisham ./out.grdmap > ./out.png`

I have also tried it without the -chisham flag but that also doesn't work. I have never used grid_plot before so I don't know if I am using it correctly and there is no -vb flag or error, so I have no clue how to trouble shoot this one. 

I attach some example files I made in case it's helpful for someone else to plot them and see if it's sensible. 
(because GitHub won't allow me to upload .grid files, I had to manually rename them 🤨 I put .txt at the end but don't let that fool you..)
[20120101.0000.12.sas.grid.txt](https://github.com/user-attachments/files/19373991/20120101.0000.12.sas.grid.txt)
[out.grdmap.txt](https://github.com/user-attachments/files/19373996/out.grdmap.txt)
 
Sorry it's taken me a while to get round to this and the result is so unsatisfactory.

Thanks,
Maria

pasha-ponomarenko:
OK, some basic testing:
(1) Presence of the `srng` field
```
pbp672@arts-phys-4826:/data/data/temp/map$ dmapdump 20180625.map > test.txt
pbp672@arts-phys-4826:/data/data/temp/out$ grep -i 'srng' test.txt
	float	"vector.srng" [141]
	float	"vector.srng" [141]
	float	"vector.srng" [126]	

```
(2) Plotting a map with slant range (unfiltered):
`pbp672@arts-phys-4826:/data/data/temp/map$ grid_plot -png -mag -latmin 40 -sf 2.00 -fcoast -coast -st 18:31 -et 18:33 -rmax 4000 -vecp -xkeyp -time -raw -max -rng -xkey rainbow.key 20180625.18.map `

unfiltered

(3) Plotting the same map with filtered data:

```
pbp672@arts-phys-4826:/data/data/temp/map$ grid_filter -minrng 800 -maxrng 2000 -vb 20180625.18.map > out.map
2018-06-25 18:01:00  264 > 188 pts (71.21%)
2018-06-25 18:03:00  264 > 188 pts (71.21%)
...

bp672@arts-phys-4826:/data/data/temp/map$ grid_plot -png -mag -latmin 40 -sf 2.00 -fcoast -coast -st 18:31 -et 18:33 -rmax 2000 -vecp -xkeyp -time -raw -max -rng -xkey rainbow.key out.map 

```
filtered

pasha-ponomarenko:
Still need to see how it affects the CPCP.

pasha-ponomarenko:
Oh, Maria beat me to it! I have not refreshed the issue page for an hour, and here we are!
:-D

mtwalach:
> Still need to see how it affects the CPCP.

I think it will most definitely reduce it in your case: since it has taken out some of the higher flow vectors, it is most likely going to reduce the CPCP overall. From the work we did in 2022 (with https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2021JA029559 ), I can tell you that this isn't always the case though. Sometimes the CPCP goes up too. 

> So, can we merge it, or some other tests are needed?

Personally, I think we should do what we have done with everything else given the situation. If I don't hear from @egthomas or @ksterne by tomorrow, I'll merge so we can move this along.

egthomas:
> I have also tried it without the -chisham flag but that also doesn't work. I have never used grid_plot before so I don't know if I am using it correctly and there is no -vb flag or error, so I have no clue how to trouble shoot this one.

@mtwalach maybe you saw this from Pasha's example, but unfortunately `grid_plot` requires lots of options to create a sensible output figure.

Regarding the impact on the CPCP, that is a bit more complicated.  Attached are two figures I created many years ago trying to see how well the different latitudinal tiers of radars agreed with one another in terms of their gridded velocities.  The top panel of this first plot shows a dusk-dawn cut of the sunward velocity for a given statistical pattern (imagine the cross-track drift measured by a low-altitude spacecraft like DMSP) when including velocities from all ranges, with mid-latitude, high-latitude, and polar radar velocities in red, blue, and green, respectively; the average (or median? I can't recall) velocity is in thick black.  The bottom panel shows the total number of grid vectors vs latitude in each data set for reference.

standard_new_4_4

You might notice that in some places one latitudinal tier of radars might under- or over-estimate the sunward velocity relative to another, and this seems to mostly occur at the edges of the data coverage.  If you apply a slant range filter (eg, this second attached figure) you can see the agreement in velocity between different latitudinal radar tiers is improved in the (now smaller) regions of overlap:

standard_halfhop_4_4

I'm not sure a repeat analysis is needed for this pull request, but hopefully that helps explain the original motivation.

pasha-ponomarenko:
@egthomas , yep, the 1.5-hop data are not as reliable as those for the 0.5-hop. 
I will leave the honour of merging this pull request to Maria. :-)

PR #618, opened 2024-07-19
Adding tdiff.dat files for multi-frequency calibration results

egthomas:
This pull request adds `tdiff.dat` files for the 12 radars listed in Table 1 of [Thomas et al. [2024]](https://doi.org/10.1029/2024RS007957), who used multi-frequency sounding mode data for interferometer calibration.  I've also updated the documentation for `make_fit` to reflect this new method that can be specified with `-tdiff_method` (ie 5).  To test, try something like

```
make_fit -fitacf[2 or 3] yyyymmdd.hrmt.sc.rad.rawacf > default_tdiff.fitacf
make_fit -fitacf[2 or 3] -tdiff_method 5 yyyymmdd.hrmt.sc.rad.rawacf > multi_freq_tdiff.fitacf
```

and compare the `tdiff` values in the parameter block / fitted elevation angles.  Note these new tdiff values are entirely optional, and `make_fit` will still default to using the tdiff values in the `hdw.dat` files.


egthomas:
@ksterne note that `tdiff.dat` files have start and end date/time values for each line; the date you've chosen is outside the valid range for the `tdiff.dat.gbr` file I added (20231108-20231206).  More generally, because GBR has not often collected XCFs for the past several years it probably isn't a good candidate for testing this pull request.

ksterne:
Thanks for the additional info.  I hadn't looked as closely as I should have on the start/end times.  I should have added originally that it could have been user error.  Using a date from within kap's range (20240408 this is working as expected.

Issue #619, opened 2024-09-26
Mac OS Sonoma Installation Issues


aburrell:
# BUG

Unable to compile code.  Specifically, I cannot compile the executables that depend on the `complex.h` header.

Using gcc complier:

```
COLLECT_GCC=/opt/local/bin/gcc-mp-14
COLLECT_LTO_WRAPPER=/opt/local/libexec/gcc/aarch64-apple-darwin23/14.2.0/lto-wrapper
Target: aarch64-apple-darwin23
Configured with: /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_lang_gcc14/gcc14/work/gcc-14.2.0/configure --prefix=/opt/local --build=aarch64-apple-darwin23 --enable-languages=c,c++,objc,obj-c++,lto,fortran,jit --libdir=/opt/local/lib/gcc14 --infodir=/opt/local/share/info --mandir=/opt/local/share/man --datarootdir=/opt/local/share/gcc-14 --with-local-prefix=/opt/local --with-system-zlib --disable-nls --program-suffix=-mp-14 --with-gxx-include-dir=/opt/local/include/gcc14/c++/ --with-gmp=/opt/local --with-mpfr=/opt/local --with-mpc=/opt/local --with-isl=/opt/local --with-zstd=/opt/local --enable-checking=release --disable-multilib --enable-lto --enable-libstdcxx-time --with-build-config=bootstrap-debug --with-as=/opt/local/bin/as --with-ld=/opt/local/bin/ld-classic --with-ar=/opt/local/bin/ar --with-bugurl=https://trac.macports.org/newticket --enable-host-shared --with-darwin-extra-rpath=/opt/local/lib/libgcc --with-libiconv-prefix=/opt/local --disable-tls --with-gxx-libcxx-include-dir=/opt/local/libexec/gcc14/libc++/include/c++/v1 --with-pkgversion='MacPorts gcc14 14.2.0_1+stdlib_flag' --with-sysroot=/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 14.2.0 (MacPorts gcc14 14.2.0_1+stdlib_flag) 
```

## Priority

- [ ] Minor - typo, incorrect naming, fails in a specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation
- [x] Urgent - failure in usage, large misrepresentation of data

## RST version

[...]

Issue #475, opened 2022-02-28
make_fov: CFLAG required for compilation


aburrell:
# BUG

When compiling my code on:

LSB Version:	:base-4.0-amd64:base-4.0-noarch:core-4.0-amd64:core-4.0-noarch:graphics-4.0-amd64:graphics-4.0-noarch:printing-4.0-amd64:printing-4.0-noarch
Distributor ID:	CentOS
Description:	CentOS release 6.10 (Final)
Release:	6.10
Codename:	Final

with:

cc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23)
Copyright (C) 2010 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

I get the following error:

```
================================================================================
Compiling Library:/home/aburrell/Git/rst/codebase/analysis/src.lib/cdf/rcdf.1.5
================================================================================
make clean
rm -f *.o
make
make.version /home/aburrell/Git/rst/codebase/analysis/src.lib/cdf/rcdf.1.5/src/..
make.hdr /home/aburrell/Git/rst/codebase/analysis/src.lib/cdf/rcdf.1.5/src/../include /home/aburrell/Git/rst/include/analysis 
/home/aburrell/Git/rst/codebase/analysis/src.lib/cdf/rcdf.1.5/src/../include/rcdf.h
`/home/aburrell/Git/rst/codebase/analysis/src.lib/cdf/rcdf.1.5/src/../include/rcdf.h' -> `/home/aburrell/Git/rst/include/analysis/rcdf.h'
cc -fPIC -Wall -pedantic -O3 -D_GNU_SOURCE -D_LINUX -I/home/aburrell/Git/rst/include/analysis -I/home/aburrell/bin/cdf38_0-dist/include -c rcdf.c    
In file included from rcdf.c:36:
/home/aburrell/bin/cdf38_0-dist/include/cdf.h:2000: warning: ISO C90 does not support ‘long long’
/home/aburrell/bin/cdf38_0-dist/include/cdf.h:2003: warning: ISO C90 does not support ‘long long’
/home/aburrell/bin/cdf38_0-dist/include/cdf.h:2006: warning: ISO C90 does not support ‘long long’
/home/aburrell/bin/cdf38_0-dist/include/cdf.h:2009: warning: ISO C90 does not support ‘long long’
[...]

PR #620, opened 2024-09-29
FIX: Bug when `CMPLX` macro undefined.

RemingtonRohel:
Targeting #619 

**NOTE**: This branch has no conflicts with the main branch, so the target branch for this PR could be changed if this change would be better suited as a hotfix.

## Changes
* Replaced complex number instantiation using `CMPLX(real, imag)` with `real + imag * I` as per https://stackoverflow.com/a/9860772
* Specified double precision for complex numbers in `rylm.c`
* Simplified complex number instantiation/multiplication in some places. Previously, some calculations were extracting the real and imaginary parts separately and using them in a `CMPLX()` call to create a new complex number, rather than simply using the already-complex number.
* Answer at https://stackoverflow.com/a/59597456 explains that `CMPLX` is missing on MacOS when compiling with `gcc` because the responsibility of defining macros like `CMPLX` is not up to `gcc`, so it truly is not defined anywhere. This is not a linker issue nor an issue with the `complex.h` header file missing/not found/wrong.

### DIFF
--- a/codebase/analysis/src.bin/cdf/istp_plot.1.11/locate.c
+++ b/codebase/analysis/src.bin/cdf/istp_plot.1.11/locate.c
@@ -67,12 +67,7 @@ int infname(char *body,char *str) {
   return 1;
 }
 
-
-#ifdef  _DARWIN
-int dsel(struct dirent *dp) {
-#else 
 int dsel(const struct dirent *dp) {
-#endif
   if (dp->d_name[0]=='.') return 0;
   return 1;
 }
--- a/codebase/analysis/src.bin/cdf/istp_text.1.24/locate.c
+++ b/codebase/analysis/src.bin/cdf/istp_text.1.24/locate.c
@@ -67,12 +67,7 @@ int infname(char *body,char *str) {
   return 1;
 }
 
-
-#ifdef  _DARWIN
-int dsel(struct dirent *dp) {
-#else 
 int dsel(const struct dirent *dp) {
-#endif
   if (dp->d_name[0]=='.') return 0;
   return 1;
 }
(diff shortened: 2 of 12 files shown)


RemingtonRohel:
I believe this also would address the problems of #475

aburrell:
I just tried compiling this branch on my computer.  This does resolve the COMPLEX problem I encountered previously.  However, I ran into the flag issue reported in #609.  I tried the recent suggestion of removing the -D_DARWIN flag and that helped me get past `istp_plot.1.11`.  However removing the -D_DARWIN flag from the makebin and makelib files causes `map.1.18` (an potentially others after this) to fail.

RemingtonRohel:
I added another commit to remove the effect of the `-D_DARWIN` flag in `istp_plot.1.11` and other instances where it was used in the same way. @aburrell does it compile now for you? If not, could you include the output when compiling?

aburrell (approved):
This solves my compilation issues on OS X Sonoma.

aburrell:
@RemingtonRohel happy to merge or to allow another person to test.

RemingtonRohel:
I'm not a maintainer, so I can't merge this myself. If it's fine with the normal RST workflow, by all means go ahead!

pasha-ponomarenko:
Merged.

egthomas:
Just to clarify, were the output of FITACF 2.5 or the `map_addmodel` / `solve_model` routines tested to make sure there was no change in behavior?

RemingtonRohel:
I did not test before it was merged, I've got no good excuse. I just verified that fitacf2.5 yields identical results, but `map_addmodel -old_aacgm` was giving different results for `model.kvect` and `model.vel.median` fields. It's a -1 error in in `cnvmodel.c`, I've made a change and verified that I get identical results. I've opened a new PR #623 with the change.

PR #628, opened 2024-12-04
Updating compilation test to use new NASA CDF library version

egthomas:
This pull request modifies the `main.yml` file used by the GitHub compilation test to retrieve the latest version of the NASA CDF library (39.1).  Currently, the compilation test fails for both Ubuntu and MacOS because it can't find version 39.0.

Note this has solved the issue for Ubuntu, but the test is still failing for MacOS (at a later stage when compiling CDF):

```
install_name_tool -change libcdf.dylib @executable_path/../lib/libcdf.dylib ./bin/cdfedit
error: /Applications/Xcode_15.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: can't open file: ./bin/cdfedit (No such file or directory)
make[1]: *** [dylibbundler.file] Error 1
make: *** [copy.tools] Error 2
Error: Process completed with exit code 2.
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "macos-latest"]
         cc: ["gcc-default"]
-        cdf: ["39_0"]
+        cdf: ["39_1"]
 
     name: ${{ matrix.cc }} on ${{ matrix.os }} with CDF ${{ matrix.cdf }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


aburrell (commented):
If the MacOS test isn't working with the new CDF, we should set it up to run with a version that does work.  Otherwise we need to solve the problem.  Which way do you want to go?

egthomas:
@aburrell that's a great point. I just poked around the NASA CDF directories and saw the previous version is still available here:

https://spdf.gsfc.nasa.gov/pub/software/cdf/dist/cdf39_0/linux/cdf39_0-dist-cdf.tar.gz

Maybe it would be easiest to stop pointing to the "latest" subdirectory and instead use a fixed version that won't change on us?

aburrell:
That works!

egthomas:
@aburrell looks like that worked - hurray!

aburrell (approved):

PR #638, opened 2025-05-29
Updating Siziwang names

ksterne:
I got an e-mail from JiaoJiao Zhang apologizing about not capturing this issue prior to the 5.1 release.  But it seems like the Siziwang radars had the wrong name based on an initial error from the planning stages of the radars.  Fairly simple pull request, so not sure what else to say here.


aburrell:
I checked the spelling against the publication: https://agupubs.onlinelibrary.wiley.com/doi/full/10.1029/2024SW004131.

aburrell (approved):

tables/superdarn/radar.dat
```diff
@@ -62,8 +62,8 @@
 50   1 20190625 25000101 "Jiamusi East"            "National Space Science Center, CAS"  "hdw.dat.jme" "jme"
 51   1 20230626 25000101 "Longjing East"           "National Space Science Center, CAS"  "hdw.dat.lje" "lje"
 52   1 20230626 25000101 "Longjing West"           "National Space Science Center, CAS"  "hdw.dat.ljw" "ljw"
-53   1 20221101 25000101 "Siziwanqi East"          "National Space Science Center, CAS"  "hdw.dat.sze" "sze"
-54   1 20221101 25000101 "Siziwanqi West"          "National Space Science Center, CAS"  "hdw.dat.szw" "szw"
+53   1 20221101 25000101 "Siziwang East"          "National Space Science Center, CAS"  "hdw.dat.sze" "sze"
+54   1 20221101 25000101 "Siziwang West"          "National Space Science Center, CAS"  "hdw.dat.szw" "szw"
```
egthomas, line 66: Ideally the column alignment should be retained after shortening the radar names

ksterne: Good catch, I know there was something off when I did it before, not sure how I didn't see it.  I've got this and another set of corrections I missed before coming up.

PR #640, opened 2025-06-10
Update locate.c

berng:
To be compiled under Ubuntu 24.


aburrell:
Hi @berng thanks for finding and fixing a bug in our code.

I have changed the branch you asked to merge into because we merge our changes into develop and not directly into main, to avoid unexpected breaking changes for our user base.  In the future it would also be helpful if your description said the change you had made (e.g., "fixed a bug in a print statement").

PR #651, opened 2025-11-14
Updated Mac install guide

billetd:
I recently successfully installed RST on an M4 Pro Mac mini with surprisingly minimal problems. This update to the Mac install guide chronicles exactly what I needed to do to get it working.

This should hopefully help alleviate the problems with Mac installations, e.g. #609 

The biggest changes are related to ensuring the x86 versions of Homebrew and netcdf are installed, along with a fix for the `png.c` compilation error. The latter is a hack, because for some reason it can't find `png.h` on compilation, even when XPATH is defined correctly (and `png.h` is in there). I'm not sure why this is the case, but no matter what you set XPATH too (even where `png.h` is directly in the top level), it doesn't work.

One thing to note is that I only did this with Homebrew, and not Macports. Many of the changes may apply to Macports enjoyers, but for now, I recommend using Homebrew.

### DIFF
--- a/docs/user_guide/mac_install.md
+++ b/docs/user_guide/mac_install.md
@@ -2,16 +2,6 @@
 
 This guide is intended to provide installation instructions on **macOS** for Apple-branded desktops and laptops. It begins with a list of dependencies required with instructions for installing RST. This process was designed for OS **Sierra** and above, but should still work on any machine which can run XCode. If you encounter any problems with installing RST, please create an [issue](https://github.com/superdarn/rst/issues/new) describing your problem and the error message you receive. The community will then help you solve the problem and add it into our troubleshooting section to help other users. 
 
-Table of Contents: 
--------------------
-
-1. [Library Requirements](#library-requirements) 
-    1. [Macports](#macports)
-    2. [Homebrew](#homebrew)
-    3. [CDF](#cdf)
-2. [Installation](#installation) 
-3. [Troubleshooting](#troubleshooting)
-
 ## Library Requirements
 
 > Warning! sudo privileges needed to install the various libraries. 
--- a/docs/user_guide/mac_install.md
+++ b/docs/user_guide/mac_install.md
@@ -113,7 +113,11 @@ Error: curses.h not found
    	(Note: for users without IDL, modifying the `IDL_IPATH` environment variable is
    	not required).
 
-2. Load the RST environment variables. Open and edit your `~/.bashrc` (or `~/.zshrc` if you're on an ARM file, or have switched from bash to zsh) to include:
+3. When compiling RST later, `png.c`, located in `/RSTINSTALLLOCATION/codebase/base/src.lib/graphic/fbuffer.1.19/src/png.c`, doesn't seem to find the `png.h` header file correctly, even though it is located in the XPATH defined in the previous step. As a patch fix while we work out what to do more cleanly, you need to change a line to point directly to `png.h`
+
+	- Go to `png.c` and change line 32 `#png.h>` to `#include </opt/X11/include/png.h>` or `#include </usr/X11/include/png.h>`, for ARM and x86-based Macs, respectively. If your compilation still errors on `png.c`, make sure this is pointed to `png.h` correctly on your machine.
+
+5. Load the RST environment variables. Open and edit your `~/.bashrc` (or `~/.zshrc` if you're on an ARM file, or have switched from bash to zsh) to include:
 
         # bash profile for rst
         export RSTPATH="INSTALL LOCATION"/rst
@@ -125,10 +129,10 @@ Error: curses.h not found
    
        source ~/.bashrc
 
-3. Run `make.build` from the command line.  You may need to change directory to `$RSTPATH/build/script`.
+6. Run `make.build` from the command line.  You may need to change directory to `$RSTPATH/build/script`.
    This runs a helper script that sets up other compiling code.
 
-4. In the same directory run `make.code` to compile all of the code.
+7. In the same directory run `make.code` to compile all of the code.
    This runs a script to find all of the source codes and compile them into binaries.
    A log of this compilation is stored in `$RSTPATH/log`.
 
--- a/docs/user_guide/mac_install.md
+++ b/docs/user_guide/mac_install.md
@@ -1,6 +1,6 @@
 # Mac 
 
-This guide is intended to provide installation instructions on **macOS** for Apple branded desktops and laptops. It begins with a list of dependencies required with instructions for installing RST. This process was designed for OS **Sierra** and above, but should still work on any machine which can run XCode. If you encounter any problems with installing RST, please create an [issue](https://github.com/superdarn/rst/issues/new) describing your problem and the error message you receive. The community will then help you solve the problem and add it into our troubleshooting section to help other users. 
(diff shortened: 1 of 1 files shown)


docs/user_guide/mac_install.md
```diff
@@ -120,33 +102,37 @@ Error: curses.h not found
...
+   - For NETCDF_PATH, point this to where you installed netcdf. This should be `"/usr/local/Cellar/netcdf/X.X.X_X"` if you installed it with Homebrew for both types of Macs, where the X's are the version number.
+     
+   - For CDF_PATH, this will be where the CDF package is installed. It should be something like `"/Applications/cdf/cdf"`. The directory pointed to should contain the CDF `ReadMe.txt`.
    
    - If you have **IDL**, check to see that `IDL_IPATH` in `rst/.profile/idl.bash` is correct.
    	(Note: for users without IDL, modifying the `IDL_IPATH` environment variable is
    	not required).
 
-2. Load the RST environment variables. Open and edit your `~/.bashrc` file to include:
+3. When compiling RST later, `png.c`, located in `/RSTINSTALLLOCATION/codebase/base/src.lib/graphic/fbuffer.1.19/src/png.c`, doesn't seem to find the `png.h` header file correctly, even though it is located in the XPATH defined in the previous step. As a patch fix while we work out what to do more cleanly, you need to change a line to point directly to `png.h`
```
aburrell, line 116:
```suggestion
3. When compiling RST later, `png.c`, located in `$RSTPATH/codebase/base/src.lib/graphic/fbuffer.1.19/src/png.c`, doesn't seem to find the `png.h` header file correctly, even though it is located in the XPATH defined in the previous step. As a patch fix while we work out what to do more cleanly, you need to change a line to point directly to `png.h`
```

docs/user_guide/mac_install.md
```diff
@@ -120,33 +102,37 @@ Error: curses.h not found
...
+
+	- Go to `png.c` and change line 32 `#png.h>` to `#include </opt/X11/include/png.h>` or `#include </usr/X11/include/png.h>`, for ARM and x86-based Macs, respectively. If your compilation still errors on `png.c`, make sure this is pointed to `png.h` correctly on your machine.
```
aburrell, line 118: This is too specific.  My png.h header was not located under X11.  Instead, I recommend directing users to find the location of their png.h header using something like: `find /opt/ -type f -name 'png.h'` or `find /usr/ -type f -name 'png.h'`

docs/user_guide/mac_install.md
aburrell: I recommend putting this whole section in "Troubleshooting" with an example of the error message that pops up when things aren't working correctly.

aburrell (commented):
Ok, this didn't work for me. Not because your hack didn't do what it was supposed to, but because in addition RST wasn't correctly linking to my libraries.  What I needed to do was:

1) Add `-I/opt/local/include` to my `makebin.darwin` file on line 10.  This is where my `png.h` file was located and meant I got past the issue of finding it without needing to change the specific file called out in your tutorial below.

2) Adding `-L/opt/local/lib` to LFLAGS on line 29 of makebin.darwin.

The installed libraries weren't being detected correctly (`-lpng`, `-lcdf`, `-lnetcdf`).  I tested adding `-L/opt/local/lib` (my library location) to the individual makefiles after the offending flags and that worked.  I then tested the more general fix, which also worked.

This leads me to suspect that your installation issue might be solved by only applying point (1).  If so, we might be able to solve this problem by adding instructions to alter the makebin, which is more top level and therefore a more desirable solution in my opinion.

There is also a change in the way that the CDF libraries are installed (at least on Mac). We probably need to revisit the way our environment variables are set to better link to the available libraries and header file locations.  That would be even better than asking users to alter the `makebin` files.

The individual comments below are suggestions for better implementing your fix.  However, if altering makebin.darwin also works for you, I think that's the more appropriate hack.

mtwalach:
Hey @billetd and @aburrell! 👋
Thanks for taking the time to try and solve this. I just wanted to let you know that I am _very_ happy to look at this but I'm a little snowed under at the moment. It's on my to do list and will do it as soon as I can but it might be next week.

billetd:
I agree @aburrell that a more top level solution would be better. I tried out your fix, pointing to the location of png.h in `makebin.darwin`, but that did not work for me. There is also already `-I$(XPATH)/include` on that line, which points directly to my png.h file. I also never had an issue with it finding `netcdf` or `cdf`, when my paths were set correctly. 

These errors are likely very device specific, and may vary depending on how much the installation environment has changed since the user got their Mac. As an example, this install with these errors was on a brand new Mac mini. When I tried to reproduce it on my older Macbook, I got different errors, but that is probably related to the fact that I had a weird homebrew setup.

aburrell:
@billetd ok, based on the fix that worked for me not working for you, I double down on my suggestion to put the fix for this in the "Troubleshooting" section of the install guide.  Ideally, including some details about what systems are more likely to fail as well.

mtwalach:
@aburrell and @billetd I appreciate the support with this but still struggling to install this. Currently failing at `dmaptoncdf.1.11` because it can't find netcdf.h. I tried fiddling with the "makebin.darwin" file as Angeline suggested but no luck. Any other ideas? 

`...
make clean
rm -f *.o
rm -f version.h
rm -f errstr.h
rm -f hlpstr.h
rm -f dmaptoncdf
make
make.version /rst/codebase/general/src.bin/dmap/dmaptoncdf.1.11
make.help
./doc/dmaptoncdf.doc.xml
cc -Wall -pedantic -O3 -D_GNU_SOURCE -D_DARWIN -I/opt/X11/include -I/opt/local/include -D_SVGLIB_ -I/rst/include/base -I/rst/include/general -I/usr/local/include -c dmaptoncdf.c    
dmaptoncdf.c:34:10: fatal error: 'netcdf.h' file not found
   34 | #include "netcdf.h"
      |          ^~~~~~~~~~
1 error generated.
make: *** [dmaptoncdf.o] Error 1
Compilation Aborted.`

my netcdf.h file is here: `/usr/local/include/netcdf.h` so I don't understand why it can't find it..

aburrell:
If you try Dan's hack of changing the line in the file to be `#include "/usr/local/include/netcdf.h"` does it work?  If it does, there might be something we can do in the makefiles.  If not, it's probably a compiler problem.

mtwalach:
@aburrell still the same problem. 👎

aburrell:
@mtwalach what are your system and compiler details?

mtwalach:
`Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: x86_64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin`

It's an Apple M3 with Sequoia 15....is that the info you're after?

billetd:
Assuming you have your `NETCDF_PATH` set correctly (and you've opened a new terminal), you could also try changing [line 34 of dmaptoncdf.c](https://github.com/SuperDARN/rst/blob/434880f20fc170ff2fc1b97b0938938947892b9d/codebase/general/src.bin/dmap/dmaptoncdf.1.11/dmaptoncdf.c#L34C1-L34C20) to point directly to where your installed x86 version of netcdf is. If `make_build` fails at the same point again, then it's a problem with your netcdf install. If it gets past it, then your `NETCDF_PATH` isn't working with RST, for some reason...

If it's the latter, could be worth double checking your .zshrc file is loading in your paths correctly. Or manually going to everything that wants netcdf and pointing it to the right header manually.

aburrell:
Yes, that's a good chunk of it.  If you use GCC instead of clang, it might work.  I'm on intel, but also have Sequoia.

PR #656, opened 2026-03-11
Reducing HMB lower limit to 30 degrees

### LINKED ISSUE #617: HMB minimum latitude?
There seems to be a hard-coded lower latitude limit to the Heppner-Maynard Boundary (HMB) of 40 degrees when determining the location from gridded velocity vectors with `map_addhmb`.  Normally this is not an issue, however during the recent May 2024 storm I've noticed the HMB becomes "saturated" at 40 degrees magnetic latitude for several hours.  Should this limit be pushed farther equatorward to better handle such storms?

Note that even though there is limited observational coverage equatorward of 40 degrees magnetic latitude, the non-circular HMB is compressed by ~10 degrees on the dayside such that higher-latitude observations can still help constrain the solution under extreme conditions (eg, from the mid-latitude radars).

ksterne:
As discussed in #617 and then more recently within the Data Analysis Working Group meetings, theres a need to reduce the minimum latitude of the HMB to 30 degrees from the 40 degrees it is at now.

I'm not as familiar with this code, so possibly double check that I got the edit correct as I'm going off of what was provided in #617.

I don't have great way to test this beyond a simple compilation test at the moment, but I can generate some data and/or plots if it needs it.

### DIFF
--- a/codebase/superdarn/src.bin/tk/plot/hmb_plot.1.6/hmb_plot.c
+++ b/codebase/superdarn/src.bin/tk/plot/hmb_plot.1.6/hmb_plot.c
@@ -216,7 +216,7 @@ int main(int argc,char *argv[]) {
   char *xtitle="Time";
   int xttl_flg=0x03;
  
-  double ymin=40,ymax=80;
+  double ymin=30,ymax=80;
   double ymajor=5,yminor=1;
   int ytick_flg=0x03;
   int ylbl_flg=0x03;
@@ -422,7 +422,7 @@ int main(int argc,char *argv[]) {
 
   if (map->hemisphere==-1) {
     ymin=-80;
-    ymax=-40;
+    ymax=-30;
   }
 
   GrplotPanel(plt,0,fgcol,0x0f,lne);
--- a/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c
+++ b/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c
@@ -54,7 +54,7 @@ struct CnvMapData *map[3];
 struct GridData *grd[3];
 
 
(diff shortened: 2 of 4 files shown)


pasha-ponomarenko:
@ksterne, this is not as simple. :-)
To change the minimum latitude for HMB, one needs to edit following three files:
/data/rst/rst-release-rst-5.1/codebase/superdarn/src.lib/tk/hmb.1.0/src/hmb.c

/data/rst/rst-release-rst-5.1/codebase/superdarn/src.lib/tk/hmb.1.0/include/map_addhmb.h

/data/rst/rst-release-rst-5.1/codebase/superdarn/src.bin/tk/tool/map_addhmb.1.17/map_addhmb.c

In each file, all 36s need to be replaced with 46s and 40s need to be replaced with 30s.

pasha-ponomarenko:
@ksterne, I have done the necessary modifications. The ultimate test would be to generate maps for an active period like 10-11 May 2024 and see if the minimum HMB MLAT goes below 40 deg. Here is an excerpt from my presentation at the last workshop. Pay attention to black dots as they are obtained with the conventional HMB algorithm.

20240510_11storm

egthomas:
I tested this for 10-11 May 2024 when the HMB was observed to saturate at 40 degrees magnetic latitude (#617).  Here is the 48-hour output from `hmb_plot` with the current 40 degree limit:

202405 hmb old

and here is the output for the same interval using the new 30 degree limit:

202405 hmb new

so things look good to me.  Note I made one more change to `hmb_plot` to support this new lower latitude limit.

egthomas (approved):

mtwalach:
Looks good to me. :)

ksterne:
@egthomas, the results look good here. Anything that is holding this PR up?  Can we merge?

PR #662, opened 2026-06-23
Adding new wide (TX) field to the radar parameter block

egthomas:
This pull request modifies the C and IDL code to support a new `widetx` field in the radar parameter block to indicate if data were produced using a widebeam TX pattern.  For example, `dmapdump` will now show something like

```
	short	"mplgexs" = 0
	short	"ifmode" = 0
	short	"widetx" = 1
	short	"nrang" = 75
	short	"frang" = 180
	short	"rsep" = 45
```

To test, you can try reading / fitting files that were produced with and without this new field; all code should be compatible whether or not the field is present in the iqdat, rawacf, fitacf, or snd file.  On the develop branch the new `widetx` field will just be skipped over, but on this branch the `widetx` field should be correctly populated.

### DIFF
--- a/codebase/superdarn/src.idl/lib/main.1.25/rprm.pro
+++ b/codebase/superdarn/src.idl/lib/main.1.25/rprm.pro
@@ -96,6 +96,7 @@ pro RadarMakeRadarPrm,prm
          tfreq: 0, $
          offset: 0, $
          ifmode: 0, $
+         wide: 0, $
          mxpwr: 0L, $
          lvmax: 0L, $
          pulse: intarr(PULSE_SIZE), $
@@ -149,6 +150,7 @@ function RadarEncodeRadarPrm,prm,sclvec,arrvec,new=new
   s=DataMapMakeScalar('mplgs',prm.mplgs,sclvec)
   s=DataMapMakeScalar('mplgexs',prm.mplgexs,sclvec)
   s=DataMapMakeScalar('ifmode',prm.ifmode,sclvec)
+  s=DataMapMakeScalar('wide',prm.wide,sclvec)
   s=DataMapMakeScalar('nrang',prm.nrang,sclvec)
   s=DataMapMakeScalar('frang',prm.frang,sclvec)
   s=DataMapMakeScalar('rsep',prm.rsep,sclvec)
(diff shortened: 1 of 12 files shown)


pasha-ponomarenko:
@egthomas , I think it might be a bit more informative (less confusing for the non-initiated ;-) ) if the name would be "wide_tx" rather than just "wide".

egthomas:
> @egthomas , I think it might be a bit more informative (less confusing for the non-initiated ;-) ) if the name would be "wide_tx" rather than just "wide".

@pasha-ponomarenko I tend to agree, although historically we haven't included underscores in any of the radar parameter block variable names.  I was just trying to get the functionality implemented before sending an email to the relevant darn- mailing lists, and then hopefully we can settle on a final name.

pasha-ponomarenko:
Then "widetx" should be as good as "wide_tx".  ;-)

egthomas:
I've changed the variable name to `widetx` as suggested by @pasha-ponomarenko - please let me know if anyone finds any bugs.

carleyjmartin (approved):
Thanks Evan!

I've installed this branch, made my own rawacf files with the 'widetx' field in set to 0.
dmapdump works for rawacf files with and without widetx field.
make_fit works to process the rawacf file including widetx to make a fitacf that also includes widetx.
dmapdump works with new fitacf file with widetx in

All double checked via rust dmap functions that are also in testing over on the dmap repo. 

opensuse 15.5

pasha-ponomarenko:
@carleyjmartin, could you add a screenshot with dmapdump output here just for the records?

carleyjmartin:
fitacf: 
```
	short	"mplgs" = 22
	short	"mplgexs" = 0
	short	"ifmode" = 0
	short	"widetx" = 0
	short	"nrang" = 75
	short	"frang" = 180
```

rawacf with:
```
	int	"rawacf.revision.major" = 1
	int	"rawacf.revision.minor" = 0
	string	"combf" = "Converted from Borealis file: /borealis_nfs/borealis_data/daily/20230101.0000.29.sas.0.rawacf.hdf5 record 1672538395622 with scaling factor = 1 ; Number of beams in record: 1 ;  ; "
	float	"thr" = 0.000000
	short	"widetx" = 0
``` 

rawacf without:
```
	int	"rawacf.revision.major" = 1
	int	"rawacf.revision.minor" = 0
	string	"combf" = "Converted from Borealis file: /borealis_nfs/borealis_data/daily/20230101.0000.29.sas.0.rawacf.hdf5 record 1672538395622 with scaling factor = 1 ; Number of beams in record: 1 ;  ; "
	float	"thr" = 0.000000
```

egthomas:
Thanks @carleyjmartin!

carleyjmartin:
I just have a quick question on implementation (as I'm assuming I'm going to be the one reprocessing our files and I want to make sure I'm doing it correctly).
 
This field is designed to say that the data in the record is derived from a wide beam transmission program? It doesn't state if the data is the full wide beam data or downsampled to resemble normalscan correct? 

So in the case that I have two files, one is the full wide beam data, and the other is the downsampled version, both of these files will have widetx = 1?

pasha-ponomarenko:
@egthomas , I guess I can merge it, right?

egthomas:
> I just have a quick question on implementation (as I'm assuming I'm going to be the one reprocessing our files and I want to make sure I'm doing it correctly).
> 
> This field is designed to say that the data in the record is derived from a wide beam transmission program? It doesn't state if the data is the full wide beam data or downsampled to resemble normalscan correct?
> 
> So in the case that I have two files, one is the full wide beam data, and the other is the downsampled version, both of these files will have widetx = 1?

@carleyjmartin yes, I think in both cases the `widetx` flag should = 1.  Hopefully the user will notice the full data file is ~16-24x larger and has data from different beam directions with identical time stamps.

> @egthomas , I guess I can merge it, right?

@pasha-ponomarenko well maybe we should allow a few more days for feedback since it is the summer time, and there is also the large CEDAR workshop happening right now in the US.

pasha-ponomarenko:
@egthomas , OK, then I will leave it to simmer for a while...

egthomas:
@pasha-ponomarenko or @carleyjmartin I think this is safe to merge now (we have already begun producing RAWACF files in Iceland and Oregon with this new field).

pasha-ponomarenko:
Done.

